Issue #2, opened 2015-09-14
Blocking call when loading credentials in Session.create_client


mpaolini:
in {{aiobotocore.session.Session}} we make use of {{self.get_credentials}} that is implemented in botocore.session.Session.

This method's implementation depends on credential provider config, and in some cases it can block while doing IO, for instance when [loading from ini file](https://github.com/boto/botocore/blob/develop/botocore/config.py#L130) or when [using a IAM] role(https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L263)

jettify:
Thanks for thre report! On my project we explicitly pass key and secret as keyword argumens, so I missed this issue.

I will take a look. Can we assume that credential loading happens only once? If so I think, spawning thread in order to read small file once is not worth effort.

mpaolini:
Yep only once, if it is OK for you I can cook a pach for this

jettify:
Sure, please do.

mpaolini:
Which `botocore` version are you targeting? looks like the one in `setup.py` is a bit old. Lots of test fail...

mpaolini:
ok I found out . it is `botocore<1.2` (they've changed the way the endpoint is created so it doesn't work on 1.2)

thehesiod:
hmm, I remember there being a comment that the credentials are periodically updated.  I looked into this and it was not a trivial fix

PR #4, opened 2015-09-14
Avoid blocking on create_client call

mpaolini:
Some credentials providers do blocking IO. Wrap all of them
with a thread so we don't block the event loop.

See issue #2

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,9 +36,9 @@ Basic Example
         key = '{}/{}'.format(folder, filename)
 
         session = aiobotocore.get_session(loop=loop)
-        client = session.create_client('s3', region_name='us-west-2',
-                                       aws_secret_access_key=AWS_SECRET_ACCESS_KEY,
-                                       aws_access_key_id=AWS_ACCESS_KEY_ID)
+        client = yield from session.create_client('s3', region_name='us-west-2',
+                                                  aws_secret_access_key=AWS_SECRET_ACCESS_KEY,
+                                                  aws_access_key_id=AWS_ACCESS_KEY_ID)
         # upload object to amazon s3
         data = b'\x01'*1024
         resp = yield from client.put_object(Bucket=bucket,
--- a/aiobotocore/session.py
+++ b/aiobotocore/session.py
@@ -15,6 +15,7 @@ class AioSession(botocore.session.Session):
         self._loop = loop
         self._loader = loader
 
+    @asyncio.coroutine
     def create_client(self, service_name, region_name=None, api_version=None,
                       use_ssl=True, verify=None, endpoint_url=None,
                       aws_access_key_id=None, aws_secret_access_key=None,
@@ -37,7 +38,11 @@ class AioSession(botocore.session.Session):
                 secret_key=aws_secret_access_key,
                 token=aws_session_token)
         else:
-            credentials = self.get_credentials()
+            # XXX Avoid blocking call spawning a thread. Some providers ('explicit', 'env')
+            # don't need this special treatment, but we don't know which one is goint to
+            # be used as yet.
+            credentials = yield from self._loop.run_in_executor(None, self.get_credentials)
+
         endpoint_resolver = self.get_component('endpoint_resolver')
 
         client_creator = AioClientCreator(
--- a/tests/test_s3.py
+++ b/tests/test_s3.py
@@ -1,3 +1,5 @@
+# coding: utf-8
+
 from functools import wraps
 import os
 import time
@@ -8,6 +10,7 @@ import shutil
 import unittest
 from unittest import mock
 import contextlib
+
 from botocore.vendored.requests import adapters
 from botocore.vendored.requests.exceptions import ConnectionError
 from botocore.compat import six
@@ -15,7 +18,9 @@ import botocore.auth
 import botocore.credentials
 import botocore.vendored.requests as requests
 from botocore.client import Config
+
 import aiobotocore
+import aiobotocore.endpoint
 
 
 def run_until_complete(fun):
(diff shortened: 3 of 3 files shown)


tests/test_s3.py
```diff
@@ -168,6 +176,12 @@ def assert_num_uploads_found(self, operation, num_uploads,
             num_uploads, amount_seen))
 
 
+class TestS3Client(BaseS3ClientTest):
+
+    def test_client_endpoint_is_patched(self):
+        self.assertIsInstance(self.client._endpoint, aiobotocore.endpoint.AioEndpoint)
```
mpaolini: this test fails for `botocore>=1.2`

tests/test_s3.py
```diff
@@ -742,7 +757,7 @@ def test_bucket_in_other_region(self):
 
     @run_until_complete
     def test_bucket_in_other_region_using_http(self):
-        client = self.session.create_client(
+        client = self.create_client(
```
jettify: Not sure if this works, should produce `RuntimeError`, since you do `loop.run_until_complete` inside coroutine, ``yield from create_client(....)`` will work or ``yield from self.session.create_client(....)``

tests/test_s3.py
```diff
@@ -1,3 +1,5 @@
+# coding: utf-8
```
jettify: do we need this line?)) This library will not work with python 2 anyway.

tests/test_s3.py
jettify: Looks like internals was changed, I need to take a look

tests/test_s3.py
mpaolini: it works. Tested on python3.4

tests/test_s3.py
mpaolini: basically the `_get_endpoint` hook was removed and now we have to rewrite the whole `create_endpoint` method

tests/test_s3.py
jettify: Strang. Anyway it is fragile and something should be done about this.  Because asyncio does not allow such calls, here simple test:
```python
import asyncio

loop = asyncio.get_event_loop()

@asyncio.coroutine
def go(loop):
    yield from asyncio.sleep(1, loop=loop)
    loop.run_until_complete(asyncio.sleep(1, loop=loop))

loop.run_until_complete(go(loop))
```
will produce error:
```
  File "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/lib/python3.4/asyncio/base_events.py", line 304, in run_until_complete
    self.run_forever()
  File "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/lib/python3.4/asyncio/base_events.py", line 271, in run_forever
    raise RuntimeError('Event loop is running.')
RuntimeError: Event loop is running.
```

mpaolini: ok will fix it

thehesiod:
this fix is wrong because IAMs are refreshable this this theoretically would only handle the initial case.  Looking at the code-paths this is not going to be easy to fix correctly.  You would need to make 
```python
RefreshableCredentials._protected_refresh()
```
a coroutine and everything that call those coroutines and out, refactoring the locks.  This is used by _make_api_call->_send_request...sign...

and also the above codepath, and also people can call stuff like botocore.auth.S3SigV4Auth.add_auth directly.

jettify:
@thehesiod
Maybe we need to advise in README to use only explicitly passed credentials or env variables in order to avoid blocking?

thehesiod:
for IAM credentials it needs to be periodically refreshed as the token expires after a certain amount of time.  I think a potential hack would be to monkeypatch botocore.utils.InstanceMetadataFetcher, and have it's retrieve_iam_role_credentials run in an executor.  However that still wouldn't completely solve the issue as you can see botocore.credentials.RefreshableCredentials running it around  lock (_refresh_lock).

thehesiod:
@mpaolini as an update what are you currently doing for this?  Eventually we need to get this resolved...it's scary having a periodic auth call that can block.

mpaolini:
@thehesiod I don't have time right now sorry

thehesiod:
oh I just meant to ask if you had solved this yourself given concerns above...and if so what strategy you used.

README.rst
```diff
@@ -36,9 +36,9 @@ Basic Example
         key = '{}/{}'.format(folder, filename)
 
         session = aiobotocore.get_session(loop=loop)
-        client = session.create_client('s3', region_name='us-west-2',
-                                       aws_secret_access_key=AWS_SECRET_ACCESS_KEY,
-                                       aws_access_key_id=AWS_ACCESS_KEY_ID)
+        client = yield from session.create_client('s3', region_name='us-west-2',
```
vir-mir: aiobotocore supports version python3.5.3 + let's use `async/await`

aiobotocore/session.py
```diff
@@ -15,6 +15,7 @@ def __init__(self, session_vars=None, event_hooks=None,
         self._loop = loop
         self._loader = loader
 
+    @asyncio.coroutine
```
vir-mir: aiobotocore supports version python3.5.3 + let's use `async/await`

aiobotocore/session.py
```diff
@@ -37,7 +38,11 @@ def create_client(self, service_name, region_name=None, api_version=None,
                 secret_key=aws_secret_access_key,
                 token=aws_session_token)
         else:
-            credentials = self.get_credentials()
+            # XXX Avoid blocking call spawning a thread. Some providers ('explicit', 'env')
+            # don't need this special treatment, but we don't know which one is goint to
+            # be used as yet.
+            credentials = yield from self._loop.run_in_executor(None, self.get_credentials)
```
vir-mir: aiobotocore supports version python3.5.3 + let's use `async/await`

thehesiod:
closing in favor of https://github.com/aio-libs/aiobotocore/pull/659

PR #18, opened 2015-12-30
Temporary fix for streaming body responses - load body to BytesIO

oselivanov:
…ect.

### DIFF
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -1,3 +1,4 @@
+import io
 import asyncio
 import aiohttp
 from aiohttp.client_reqrep import ClientResponse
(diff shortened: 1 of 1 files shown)


oselivanov:
Hm, looks like version in master already fixed. Closing as invalid for now.

oselivanov:
I've ran wrong test, actually it's still broken, so my fix is valid.

oselivanov:
This quickfix breaks API BTW.

    body = await obj['Body'].read()

will become 

   body = obj['Body'].read()

I'll fix it today if I have a minute.

oselivanov:
Done. I've amended commit.

cbrichford:
I believe #19 ( which was just merge ) fixes this issue.

PR #19, opened 2016-01-19
Fixing streaming responses.

cbrichford:
Closing the body stream now closes the underlying aiohttp reponse object.

This is an attempt to improve on the fix proposed in #18.  The fix in this PR does not require streaming the entire response into memory.  The FlowControlStreamReader instance coming from aiohttp is wrapped with a proxy that adds a close method ( which closes the underlying aiohttp request ) and proxies all other attributes to the FlowControlStreamReader instance.

This allows you to write code like this:
```python

resp = yield from s3.get_object(Bucket='mybucket', Key='k')
stream = resp['Body']
try:
    chunk = yield from stream.read(s)
    while len(chunk) > 0:
      ...
      chunk = yield from stream.read(s)
finally:
  stream.close()

```

### DIFF
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -32,6 +32,23 @@ def convert_to_response_dict(http_response, operation_model):
     return response_dict
 
 
+class ClientResponseContentProxy:
+    """Proxy object for content stream of http response"""
+
+    def __init__(self, response):
+        self.__response = response
+        self.__content = self.__response.content
+
+    def __getattr__(self, item):
+        return getattr(self.__content, item)
+
+    def __dir__(self):
+        attrs = dir(self.__content)
+        attrs.append('close')
+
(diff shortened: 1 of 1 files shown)


jettify:
Looks good to me. I will play with your PR for testing purposes.

thehesiod:
btw Chris and I were discussing for adding "await with" support to this solution so clients don't need to manually close the response.  In fact I believe the underlying aiohttp ClientResponse class already supports this.

jettify:
I am sorry for delay. I tested your code and it works as expected :+1:  It took me a while to figure out why there are no disconnection errors. Seems like main reason is that response object is collected by `gc` as result we see disconnection error. If I disable `gc` version without fix works correctly.

@oselivanov @cbrichford thank you for tracking this issue!

oselivanov:
Thank you guys!

PR #21, opened 2016-01-28
S3 keep-alive fix

thehesiod:
This fixes keep-alive for S3 ClientSessions.  Per AWS the keep-alive timeout is 20s whereas the default timeout for aiohttp is 30s.  This results in cached session connections periodically failing.  This change bumps down the default connection timeout to 12s.  I will have a pending change that allows users to parameterize the TCPConnection to enable things like DNS caching.


jettify:
good point!

PR #22, opened 2016-01-28
aiohttp.TCPConnector client parameter support

thehesiod:
This PR allows clients to specify allowable TCPConnector parameters, like keepalive_timeout and use_dns_cache.


aiobotocore/client.py
```diff
@@ -6,16 +6,50 @@
...
 
 
+class AioConfig(botocore.client.Config):
+    def __init__(self, connector_args=None, **kwargs):
+        super().__init__(**kwargs)
+
+        self._validate_connector_args(connector_args)
+        self.connector_args = copy.copy(connector_args)
+
+        if 'keepalive_timeout' not in connector_args:
```
jettify: this line produces error:
```
    if 'keepalive_timeout' not in connector_args:
TypeError: argument of type 'NoneType' is not iterable
```
`connector_args` should be checked for `None`

aiobotocore/client.py
```diff
@@ -6,16 +6,50 @@
...
+                if not isinstance(v, bool):
+                    raise ParamValidationError(report='{} value must be a boolean'.format(k))
+            elif k in ['conn_timeout', 'keepalive_timeout']:
+                if not isinstance(v, float) and not isinstance(v, int):
+                    raise ParamValidationError(report='{} value must be a float/int'.format(k))
+            elif k == 'limit':
+                if not isinstance(v, int):
+                    raise ParamValidationError(report='{} value must be an int'.format(k))
+            elif k == 'ssl_context':
+                import ssl
```
jettify: any reason to have import here and not in the top of the file?

thehesiod: I figured you don't want to import modules you don't need due to memory implications, and since this is a built-in module it shouldn't break scripts.  I can move it to the top if you'd like though.

aiobotocore/client.py
thehesiod: whoops, thanks, fixed!

aiobotocore/client.py
jettify: totally fine with module, just want to make sure we import module once and can find it in expected place.

thehesiod: from what I read all modules are globally cached so I think we're good

jettify:
Also rebase is required since master moved ahead a bit.

thehesiod:
TravisCI is failing because the 80 character width.  That seems pretty rigid and makes the code look really ugly.  Is there a way to bump it up a bit?

jettify:
I am sorry but that is principle, from other side i have plan to move
library to https://github.com/aio-libs (as i did with my other projects )
to get support from more experinced devs and community.  pep8, pyflakes,
good quality and test coverage is must there.

thehesiod:
ok fixed

jettify:
Thanks!

thehesiod:
thank you!

PR #24, opened 2016-01-30
add async for support to pagination

thehesiod:
This adds "async for" support to pagination. Example:
```python
                paginator = self._s3_client.get_paginator('list_objects')
                iter = paginator.paginate(Bucket=self._bucket, Prefix=fullpath)
                async for p in iter:
                    for o in p['Contents']:
                        results.append(o['Key'])
```
fixes #23


jettify:
Thanks for PR, but it is a bit early. I am not sure that current pagination is working at all. It is pain to rewrite pagination iterator from botocore correctly since it is not possible to to use `yield` statement.

I need to find courage and finally fix that huge iterator, adding `__aiter__` would be trivial then.

thehesiod:
I'm using it but haven't tested it for large result sets yet. I'll give it
a try and report back.

jettify:
Cool! Thanks! I need find time to review that  monster too and finally finish it.

thehesiod:
Just tested it on prefix that had 12,644 items and seems to work ok.  It returned the same number of items as from aws s3 ls

jettify:
I am surprised that paginator worked since ported tests are failing.

aiobotocore/paginate.py
```diff
@@ -59,12 +59,14 @@ def _init_pager(self):
         self._starting_truncation = 0
         self._inject_starting_params(self._current_kwargs)
 
-    @asyncio.coroutine
-    def next_page(self):
+    async def __aiter__(self):
+        return self
+
+    async def __anext__(self):
```
jettify: code should be backward compatible, just do something like:
```
@asyncio.cororutine
def __anext__(self):
    ....
```
and replace `await` with `yield from`, also not need for now to remove`next_page` method.

jettify:
Do you think we need to get rid of python versions <3.5 ? I like this idea, better do this earlier then later.

aiobotocore/paginate.py
thehesiod: I did it like this because I saw your comment about wanting to get rid of python <3.5.  I can make it backwards compatible as well.  Up to you :)

jettify: Lets do this backward compatible for now, since my  tests working only for 3.4,  then we can think about moving to 3.5.

thehesiod:
its failing due to the line width being 80.  Is that a hard limit or could you make it a little wider?  To make it 80 wide I'd have to uglify the code a bit to break up lines.

jettify:
Because of `StopAsyncIteration`, I should have mention that, just add:
```
if PY35:
    @asyncio.coroutine
    def __aiter__(self):
        ...
```
here is example from aiopg: https://github.com/aio-libs/aiopg/blob/master/aiopg%2Fcursor.py#L380-L384

thehesiod:
ok I've added that, however it seems the travis CI tries to analyze the code even though its marked as 3.5 only.  ideas?

thehesiod:
hah! figured out the "magic" comment I needed to add to make this work :)

jettify:
Looks good! Thanks

PR #28, opened 2016-02-09
variety of fixes

thehesiod:
This PR fixes pagination with the new botocore, adds the missing before event, and some new features.

- \__aenter__/\__aexit__ forwards to the aiohttp.ClientResponse for "async with" support to context close the response.  This will "release" the response object instead of close to allow HTTP1.1 sockets reuse.
- hook up missing 'before-event-' (prevented add_expect_header from being called)
- add force_close TCPConnector parameter support
- fix AioBaseClient.close as its async + was misspelled


jettify:
Thank you for working on this! I need to play with your PR and write couple test fixtures to make sure everything working smoothly. But for nearest couple of days I have other open source project that need attention.

thehesiod:
thank you for this project!  I've been slowly migrating from aio-s3, we basically ended up re-writing that and marrying it to botocore.  As an FYI the author of aio-s3 is now telling others to migrate to your library.

thehesiod:
btw I've just encountered https://github.com/boto/botocore/pull/303/files in aiobotocore.  It is actually a fairly involved fix, I'll merge it into this PR as its really important to fix as put_object won't work w/o it.

thehesiod:
I think I figured out an issue I was having.  I had two aiobotocore sessions, one doing a ton of GetObjects, the other doing PutObjects, the second one will eventually fail with a errno 61 (connection refused).  I have a feeling this is similar to the issue I reported and was fixed last year: http://bugs.python.org/issue25593.  I'll ping the python folks about it.

thehesiod:
ok nevermind, ends up our router is doing some kind of QoS shennanigans, what time sink ;)

jettify:
that is funny) you never know where true subtle bug and where sink)

thehesiod:
FYI, I just found that our version of aio-s3 (https://github.com/farmersbusinessnetwork/aio-s3) is about 25% faster than aiobotocore.  Running the code through an async aware profiler (https://github.com/what-studio/profiling/tree/eventloop-aware  FYI it's pretty cool!) it seems botocore adds a lot of overhead to each request.  I'm working on a project where I'll be doing millions of file operations so every ounce counts.  I'll keep you apprised if interested.  I'm trying to hit > 2000 ops/s on a dual core AWS instance.  I'm currently at ~1000 with a single core pegged.

jettify:
1) `aios3` should be faster, since it is thin wrapper around s3 REST API and probably better for your case.
2) if you do some sort of download (get_object), you may try to presign url and then use regular `aiohttp.ClientSession`
3) Thans for profiler! I was looking for event loop aware profiler myself. Does it exclude event loop poll time?

aiobotocore/endpoint.py
```diff
@@ -44,9 +48,45 @@ def __getattr__(self, item):
...
-        attrs.append('close')
+        attrs += ['read_all', 'text', 'json' 'close']
         return attrs
 
+    def __del__(self):
+        self.close()
+
+    @asyncio.coroutine
+    def read_all(self):
+        """
```
jettify: can we hold on on adding this changes? I want to get red of  `__getattr__` all together and do right thing without magic at all.

aiobotocore/endpoint.py
```diff
@@ -97,18 +137,19 @@ def __init__(self, host,
...
     @asyncio.coroutine
     def _request(self, method, url, headers, data):
+        expect100 = method in ['PUT', 'POST'] and hasattr(data, 'read')
 
         headers_ = dict(
             (z[0], text_(z[1], encoding='utf-8')) for z in headers.items())
-        resp = yield from self._aio_seesion.request(
-            method, url=url, headers=headers_, data=data)
+        resp = yield from self._aio_session.request(
+            method, url=url, headers=headers_, data=data, expect100=expect100)
```
jettify: could you elaborate on this? s3 could send 100 status code?

jettify:
I am porting integration tests from botocore (see recent commits) and will tests with your PR. Can you remove new api: `text`, `json` and leave only fixes? I have couple of reasons not to add them now:

1) while we have poor code coverage, it is not good idea too add new public APIs
2) I want to get rid with all magic I have introduced with `__getattr__` and implement explicit methods for status code, headers etc. This way we can have only minimum required set of APIs that do not break things internally.

aiobotocore/endpoint.py
thehesiod: yep!  I'll fix this up

aiobotocore/endpoint.py
thehesiod: hmm, it may be that this is not needed after all: https://github.com/boto/boto/issues/840

thehesiod:
sorry for delays, and thanks for the tips! I'm still trying to optimize my scenario (getting recursive paginated list working with a fixed number of workers was not fun, having to work with a limited number of files/sockets not fun either [AWS docker containers]). I'm validating a variety of ways to break the 1000 S3 ops/s barrier :) I've finished creating an async multiprocess S3 Pool class which allows one to run aios3/aiobotocore requests across a pool of processes (didn't help as much as I had hoped).  I'm going to look next into the presigned urls however I've been hesitant as I'm working with over 500,000 prefixes.

re: profiler it does seem to exclude the polling time.  I've been using that and PyVMMonitor (to look at pretty graphs heh).  There really isn't a smoking gun anymore, I'll keep you apprised as to any insights.

aiobotocore/endpoint.py
thehesiod: removed!  I actually realized that read_all wasn't needed as I can now use "async with" with the aenter/aexit methods I added to get at the response object and call on it directly

aiobotocore/endpoint.py
thehesiod: removed!

aiobotocore/client.py
```diff
@@ -222,6 +235,7 @@ def get_paginator(self, operation_name):
                 self._cache['page_config'][actual_operation_name])
             return paginator
 
+    @asyncio.coroutine
     def close(self):
         """Close all http connections"""
-        self._endpoint._aio_seesion.close()
+        yield from self._endpoint._aio_session.close()
```
thehesiod: btw, any reason to not do this automatically from the __del__ with a loop.run_until_complete?   Otherwise you get a warning from aiohttp.

jettify: Not reasons, we just need to add warning that user forget to close sessions

jettify:
Merged this branch manually with rebase in order to keep history more readable. My tests passed agains this branch, with one minor fix, `sessing.close()` is not coroutine :)

thehesiod:
interesting in re to session.close, perhaps our versions of aiohttp differ?  I know they changed recently

jettify:
Good catch! https://github.com/KeepSafe/aiohttp/blob/master/aiohttp%2Fclient.py#L425-L435 actually it could be used in both ways since, method returns `Future`. I think I will do same trick.

thehesiod:
btw congrats on getting into aio-libs!  And sorry for my messy PRs, I’m still getting used to git. I was a long time perforce user before :)

jettify:
Thanks! I have made same trick as aiohttp so `close()` could be used with `yield from` https://github.com/aio-libs/aiobotocore/commit/76e9566c8d2dc09aa82fdb9f35be024912409f39

PR #40, opened 2016-03-18
bugfix for pagination

thehesiod:
looks like we missed this.  Here's a testcase when you have only one page of data:

```python
import aiobotocore

session = aiobotocore.get_session()
ecs_client = session.create_client('ecs')

@asyncio.coroutine
def paginated(obj, method_name, result_list, **kwargs):
    paginator = obj.get_paginator(method_name).paginate(**kwargs)
    result = []

    while True:
        p = yield from paginator.next_page()
        if p is None: break

        result += p[result_list]

    return result

paginated(ecs_client, 'list_clusters', 'clusterArns')
```

### DIFF
--- a/aiobotocore/paginate.py
+++ b/aiobotocore/paginate.py
@@ -63,7 +63,7 @@ class AioPageIterator(PageIterator):
         else:
             self._total_items += num_current_response
             self._next_token = self._get_next_token(parsed)
-            if all(t is None for t in self._next_token):
+            if all(t is None for t in self._next_token.values()):
                 self._is_stop = True
                 return response
             if self._max_items is not None and \


jettify:
`for` loop is not supposed to work in this case, that is why I do in tests:
```
while True:
    page = yield from pag.next_page()
    if not page:
        break
```
we can use `async for` but only in python 3.5

thehesiod:
sorry, accidentally put the botocore version I was testing against, updated the comment for the aiobotocore.  You can see the issue if you very carefully compare the botocore version of the PaginationIterator with the aiobotocore verison.

jettify:
Got it, I executed tests, looks everything ok. Thank you!

PR #41, opened 2016-03-25
fix profile parameter, remove unused arg, and make init future safe

thehesiod:
this makes \__init__ future safe by passing all non-aio args to the parent.  The one case where this could break things is if people were not using "loop" as a kwarg, but I think people would have to have been kinda crazy to do that since they would have had to pass it a bunch of Nones :)

By doing this if allows you to pass the "profile" parameter


jettify:
Thanks!

thehesiod:
my pleasure!

PR #42, opened 2016-03-26
fix timeouts

thehesiod:
This fixes timeouts by setting the connect_timeout specified against the TCPConnector, and doing a wait_for on the request with the request_timeout.  This can be a big issue as I've seen in prod S3 requests that took > 9000s! :)


jettify:
Thanks! Changes working as expected, I only fixed simple issue with loop, `asyncio.wait_for` requires event loop as argument. This is important because `aiobotocore` can also work with explicit loop and do not rely on global one.

thehesiod:
oh cool thanks for the fix!  ya that's important when using ZMQ

PR #44, opened 2016-03-31
Timeout and retry fixes

thehesiod:
Timeouts
* this value is now automatically set so should no longer by set manually

Retries
* previously any retry would halt ALL aio calls on the active thread (this was a bad bug)

Misc
* simplify ClientResponseProxy
* bugfix in aenter (should return self not aenter of ClientResponse
* fix \__del__ to call release/close appropriately
* fix aenter/aexits
* add client aenter/aexit
* fix retries of aiohttp exceptions


thehesiod:
btw, I'm not sure the retrylogic is completely correct.  in botocore.retryhandler it uses an EXCEPTION_MAP which currently has urllib3 exceptions in them: ConnectionError, ClosedPoolError.  I think this whole sections needs to be reviewed to make sure the aiohttp exceptions will be handled correctly.

jettify:
Could you point me where is time.sleep logic? From quick look I do not see it...

jettify:
Sorry found.

thehesiod:
I found it while I was trying to see if I could use the botocore retry logic for aio-s3 :)

jettify:
Changes looks good, I will execute test suite and mere this PR tomorrow.

This types of bugs is very nasty and hard to find, I should have enabled logging for slow callbacks in event loop: https://docs.python.org/3/library/asyncio-dev.html#debug-mode-of-asyncio long time ago.

thehesiod:
cool, what do you think about the issues with the aiohttp exception types?  Seems like it will require some research.

jettify:
I think that logic should not that bad : https://github.com/boto/botocore/blob/90382a41c5a23e27eda1c8fbf23fc1a3b17804ed/botocore/endpoint.py#L175-L204

here `botocore` does not do any exception magic, just tires to create better exceptions in two cases, so aiohttp exception bubbles up.

jettify:
It could be improved but issue is not the end of the world at least :)

thehesiod:
the issue is that it may not retry aiohttp connection exceptions correctly, see https://github.com/boto/botocore/blob/90382a41c5a23e27eda1c8fbf23fc1a3b17804ed/botocore/retryhandler.py#L147

jettify:
Oh, I see, that should be investigated, I create issue https://github.com/aio-libs/aiobotocore/issues/47 in order not to forget.

thehesiod:
btw, added some more clean-up and a couple bugfixes

thehesiod:
just submitted fix for https://github.com/aio-libs/aiobotocore/issues/47

jettify:
Sorry for delay, I manually merged part of commits from this PR to master. Lets do retry logic in separate PR. I will outline some issue here.

aiobotocore/endpoint.py
```diff
@@ -51,43 +69,43 @@ def __dir__(self):
         return attrs
 
     def __del__(self):
-        self.close()
+        # callers may not read the Body resulting in an unclosed ClientResponse
+        if self._loop.is_running():
+            asyncio.ensure_future(self.__response.release(), loop=self._loop)
```
jettify: If body is not fetched, connection and response should be closed with warning like aiohttp does, you can use:
`(self.__response.close()` without `ensure_future`. ensure_future is dangerous here because we can loose exception.

aiobotocore/endpoint.py
```diff
@@ -51,43 +69,43 @@ def __dir__(self):
...
 
     def __init__(self, *args, **kwargs):
-        self._impl = ClientResponse(*args, **kwargs)
-        self._body = None
+        super().__init__(*args, **kwargs)
 
     def __getattr__(self, item):
         if item == 'status_code':
-            return getattr(self._impl, 'status')
+            return self.status
```
jettify: Better just drop `__getattr__`, that was my stupid idea and laziness )) adding required properties and methods would be much nicer and more explicit.

thehesiod: your wish is my command, updated!

aiobotocore/client.py
```diff
@@ -183,6 +186,17 @@ def get_paginator(self, operation_name):
                 self._cache['page_config'][actual_operation_name])
             return paginator
 
+    if PY_35:
+        @asyncio.coroutine
+        def __aenter__(self):
```
jettify: just curious why we need this? Looks like it is not public api

aiobotocore/endpoint.py
jettify: Thanks! new plain properties look much nicer!

aiobotocore/endpoint.py
thehesiod: from what I understand release is better than close as close can close the connector.  And close is a coroutine so you need to run it with an ensure_future.  Perhaps then just remove the \__del__?  that way you get the warning

aiobotocore/endpoint.py
```diff
@@ -36,9 +52,11 @@ def convert_to_response_dict(http_response, operation_model):
 
 
 class ClientResponseContentProxy:
-    """Proxy object for content stream of http response"""
+    """Proxy object for content stream of http response.  This is here in case
+    you want to pass around the "Body" of the response without closing the
+    response itself."""
 
-    def __init__(self, response):
+    def __init__(self, response: ClientResponse):
```
jettify: is this function annotation ? ;) I do not see code that uses it or it is for linter?

aiobotocore/client.py
thehesiod: this is so you can do:
```python
session = aiobotocore.get_session()
async with session.create_client() as client:
    # do work
```

this way the TCPConnector is closed when you're done using the client and you don't need to call client.close()

aiobotocore/endpoint.py
jettify: So socket connection is not useful if all data from buffer has not been fetched, `release()` tries to read all available data in order to save connection and return it to connection pool (session). But since this object is garbage collected anyway, we should drop connection and warn user that he should release/close connection explicitly.

jettify: here could be corner case for instance, user decided not to fetch huge file (1GB), and ensure_future with release will try to read all the data on backgound.

thehesiod: actually ya that's a better example, also I found an issue with the new properties as it seems someone is trying to set them, working on fix

aiobotocore/client.py
jettify: Look good! we need some tests for that in separate file, because this is only python3.5 ((

aiobotocore/endpoint.py
thehesiod: actually I found it would't work as-is, see update change

aiobotocore/endpoint.py
thehesiod: removed \__del__ and added comment about behavior

thehesiod:
closing and opening new PR per comments

aiobotocore/endpoint.py
thehesiod: it's a type hint for pycharm (my favorite editor)

PR #48, opened 2016-04-08
client + endpoint updates

thehesiod:
* adds "async with" support to client
* enables aiohttp retry-able exceptions
* removes ClientResponseContentProxy \__del__ since ClientResponse will close + warn user
* fixes "async with" support for ClientResponseContentProxy (missing yield from)
* fixes endpoint's \_get_response to handler aiohttp exceptions and put logging back
* fixes endpoint's request timeout to be connect + read timeout

this continues https://github.com/aio-libs/aiobotocore/pull/44

### DIFF
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -69,7 +69,8 @@ class ClientResponseContentProxy:
 
         @asyncio.coroutine
         def __aexit__(self, exc_type, exc_val, exc_tb):
-            return (yield from self.__response.__aexit__(exc_type, exc_val, exc_tb))
+            return (yield from self.__response.__aexit__(exc_type,
+                                                         exc_val, exc_tb))
 
     def close(self):
         self.__response.close()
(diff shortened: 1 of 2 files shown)


aiobotocore/endpoint.py
```diff
@@ -72,24 +93,28 @@ class ClientResponseProxy:
     """Proxy object for http response useful for porting from
     botocore underlying http library."""
 
+    # NOTE: unfortunately we cannot inherit from ClientReponse because it also
+    # uses the `content` property
     def __init__(self, *args, **kwargs):
-        self._impl = ClientResponse(*args, **kwargs)
-        self._body = None
+        self._response = ClientResponse(*args, **kwargs)
```
jettify: This even better! Later we can figure out what exact methods and  properties are used and only proxy those.

thehesiod: let me know if you want me to put it back to _impl

jettify: `_resonpce` is more clear I think

jettify:
Tests are green, changed looks good!

jettify:
Probably I will add some code to the tests today, so we can do `async/awiat` tests in same codebase

thehesiod:
cool!  sorry about my unittest laziness :)

PR #57, opened 2016-08-11
Force plain (not gzipped) aws response on aiohttp request

rymir:
When using aiobotocore with dynamodb, requests fail on crc32 checksum computation as soon as the response data reaches ~5KB.

It seems that this is  due to aws sending data back in gzip compressed format, aiohttp automatically uncompressing it and crc32 checksum being computed on the uncompressed data bytes.
While crc32 has to be computed on compressed data in order to pass.

The same requests work as expected when using botocore, this commit forces aws not to use gzip compression.

I created this fix to show the problem and to temporarily patch the issue, but the right way would be configuring aiohttp not to uncompress the response and leave that up to botocore, so we can take advantage of the gzip compression when data >= ~5KB.

### DIFF
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -154,6 +154,18 @@ class AioEndpoint(Endpoint):
 
     @asyncio.coroutine
     def _request(self, method, url, headers, data):
+        # Note: When using aiobotocore with dynamodb, requests fail on crc32
+        # checksum computation as soon as the response data reaches ~5KB.
+        # When aws response is gzip compressed:
+        # 1. aiohttp is automatically uncompressessing the data
+        # (http://aiohttp.readthedocs.io/en/stable/client.html#binary-response-content)
+        # 2. botocore computes crc32 on the uncompressed data bytes and fails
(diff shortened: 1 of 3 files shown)


jettify:
Thanks for report, I will take a look once I got home, next week.

rymir:
@jettify any update on this one?

jettify:
sorry  for delay I will execute my test suite for s3 today, to make sure everything is ok. Also I need help with testing, could you port at least one test from botocore that hits dynamo?

rymir:
I'm struggling finding the tests we could port, which one of https://github.com/boto/botocore/tree/master/tests you think we should port?

jettify:
oh I see, they do not have any. Then very simple test create/delete dummy table will work fine.

rymir:
@jettify ok, good chance to learn some `pytest`!

jettify:
@rymir I executed s3 test suite with your branch, tests are green. Lets also add comment why that header is important, just to avoid confusion later.

rymir:
@jettify added a test and a comment

jettify:
thanks!

PR #75, opened 2016-11-16
Tests cleanup

jettify:


setup.py
```diff
@@ -4,7 +4,7 @@
 from setuptools import setup, find_packages
 
 
-install_requires = ['botocore>=1.4.0, <=1.4.73', 'aiohttp>=0.21.2']
+install_requires = ['botocore>=1.4.0, <=1.4.28', 'aiohttp>=1.1.0']
```
adamrothman: @jettify Does aiobotocore actually require aiohttp>=1.1.0? The service in which I would like to aiobotocore is stuck on aiohttp==0.22.5 until https://github.com/KeepSafe/aiohttp/issues/1180 is resolved. As far as I can tell there aren't any changes here that strictly require bumping the aiohttp version? Would it be possible to leave this alone? Or at least make it >=0.22.5?

jettify:
@adamrothman aiobotocore should work with older version too. If tests passes with `0.22.5` I will change it to `>=0.22.5`. I just wanted to bump aiohttp since v1.0.0 available for a while.

Will do testing and update of PR  a bit later.

adamrothman:
Thanks @jettify that would be awesome. If bumping that dependency isn't required, far more people will be able to use this new version.

jettify:
new version released https://pypi.python.org/pypi/aiobotocore/0.0.6

PR #80, opened 2016-11-29
pass proxy config for aiohttp request

### LINKED ISSUE #79: HTTP Proxy settings not passed to aiohttp request
HTTP proxy defined in *HTTP_PROXY* and *HTTPS_PROXY* environment variables are not passed to *aiohttp request*, which makes the library unusable in private VPCs for other things than S3 with VPC endpoint.

Looks like the proxy configuration is read using utils derived from *botocore* but then it's just ignored in the AioEndpoint constructor.

jjonek:
Fixes #79

Endpoint URL seems to get validated before it's passed to *AioEndpoint*.

### DIFF
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -11,6 +11,8 @@ from botocore.exceptions import EndpointConnectionError, ConnectionClosedError
 from botocore.utils import is_valid_endpoint_url
 from botocore.hooks import first_non_none_response
 
+from urllib.parse import urlparse
+
 PY_35 = sys.version_info >= (3, 5)
 
 # Monkey patching: We need to insert the aiohttp exception equivalents
@@ -145,6 +147,8 @@ class AioEndpoint(Endpoint):
             connector = aiohttp.TCPConnector(loop=self._loop,
                                              conn_timeout=self._conn_timeout,
                                              **connector_args)
+        scheme = urlparse(host).scheme
+        self.proxy = proxies.get(scheme)
 
         self._aio_session = aiohttp.ClientSession(
             connector=connector,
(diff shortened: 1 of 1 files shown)


jettify:
Good point, this feature should give us ability to tests aiobotocore with moto library. I will take a look, also we need to figure out how to test it.

jjonek:
Yeah I had to skip the tests as it looks like you'd need IAM credentials to run them.

Passing the proxy config could be easily unit tested with mocking *aiohttp.ClientSession*. However, as the tests seem to be more like integration tests (I see the point) one could also try passing a proxy config that shall fail fast and expect *aiohttp.errors.ProxyConnectionError* to be raised.

thehesiod:
if you could give a snippet of how I could test this locally I can write a unittest

jjonek:
For simple test you could run any snippet on command line, for example the _examples/simple.py_ with proxy settings in ENV:
```bash
$ HTTPS_PROXY=http://localhost:54321 python simple.py
# should fail with aiohttp.errors.ProxyConnectionError unless you have http proxy listening there

$ HTTPS_PROXY=http://localhost:54321 NO_PROXY=amazonaws.com python simple.py
# should succeed because AWS endpoints are white listed
```

For the tests, looks like the interface doesn't provide any clean ways for defining proxy settings, so the only way I can think of is either update `os.environ` on the fly (which is imho bit ugly) or patching `AioEndpointCreator.get_proxies` (or the super class `botocore.endpoint.EndpointCreator`), which should return something like this:

```bash
$ HTTP_PROXY=http://myproxy:8080 HTTPS_PROXY=http://myproxy:8080 python << EOF
import botocore.endpoint
print(botocore.endpoint.EndpointCreator(None)._get_proxies('https://sts.amazonaws.com'))
EOF
{'http': 'http://myproxy:8080', 'https': 'http://myproxy:8080'}
```

`botocore.endpoint.EndpointCreator._get_proxies` is being called here: https://github.com/jjonek/aiobotocore/blob/491aeb83a57b38aee597709876f282664fa33d6a/aiobotocore/endpoint.py#L296

thehesiod:
ya only environment support really sucks, already tracked here in botocore: https://github.com/boto/botocore/issues/541

thehesiod:
ok, I've got it all working, see https://github.com/aio-libs/aiobotocore/pull/97  let me know if you want me commit that or if you'd like to update your PR and commit this one.  Thanks!

jjonek:
Sorry to keep you waiting. Feel free to merge that, I don't really mind how it gets merged, I'm satisfied if it just makes it to upstream :)

thehesiod:
done!

PR #74, opened 2016-11-16
Newest botocore (1.4.73) support. Fixing #55, #53 and #52.

vas3k:
When I tried to use aiobotocore for sending SMS through SNS I've got an error: 
```
ParamValidationErrorbotocore.validate in serialize_to_request
errorParameter validation failed: Unknown parameter in input: "PhoneNumber", must be one of: TopicArn, TargetArn, Message, Subject, MessageStructure, MessageAttributes
```
That's because botocore dependency version was pinned to >=1.4.0 <=1.4.28 because of issues #55, #53 and #52. But support for a `PhoneNumber` parameter in `sns.publish()` was added later. But I really needed the `aiobotocore` library because I'm using Multipart Upload to S3 from `aiohttp` web server.

So I spent an evening to compare the newest `botocore==1.4.73` source code with current aiobotocore modules and have fixed some incompatibilities.

Now I have perfectly working aibotocore in my project. I'm using `aiohttp==1.1.1`, `botocore==1.4.73` and `Python 3.5.2` (with flawless async/await) to upload multipart files to S3 and SNS features (push notifications, SMS). Everything works great.

I've run included tests and got 18 out of 20 passed. As you see below, 2 failed tests are related to DynamoDB and hardcoded ARN's, so I can't test them. This is related to #65 I think. But all Travis tests are passed successfully :)
```
botocore.exceptions.ClientError: An error occurred (AccessDeniedException) when calling the DeleteTable operation: User: arn:aws:iam::9999999999:user/vas3k is not authorized to perform: dynamodb:DeleteTable on resource: arn:aws:dynamodb:us-east-1:641171361437:table/aiobotocoretest_1479318016

===== 18 passed, 1 skipped, 1 xfailed, 2 error in 67.98 seconds ======
```

I hope my pull request will be helpful and I've not missed any important places in code.

PR #86, opened 2016-12-20
pass timeout to request, so it is propagated into StreamRead and affects .read method

vharitonsky:
Currently StreamReader returned by aiohttp is not affected by read timeout passed into client, so .read method has 300 seconds timeout by default, we should pass the read timeout into request so it is set on StreamReader

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,11 @@
 Changes
 -------
 
+0.1.0 (2017-01-12)
+^^^^^^^^^^^^^^^^^^
+* Pass timeout to aiohttp.request to enforce read_timeout #86 (thanks @vharitonsky)
+  (bumped up to next semantic version due to read_timeout enabling change)
+
 0.0.6 (2016-11-19)
 ^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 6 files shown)


thehesiod:
looking through aiohttp this does seem to be the correct fix for read_timeout.  Note: this ends up going to aiohttp.streams.StreamReader.

thehesiod:
btw, I think with this the asyncio.wait_for can be removed right?  It should now be handled by the TCPConnector's conn_timeout (already set), and now the read_timeout set by this PR.

thehesiod:
btw, per my research mentioned in #88 the removal of the wait_for is indeed correct iff read_timeout is >= conn_timeout (default is 60s each).  Could you please add an assert and comment to ensure this is the case in the future...then I'll merge this given jettify hasn't responded yet.

thehesiod:
FYI I've created a PR for aiohttp to later be able to clearly set the read + connection timeouts: https://github.com/KeepSafe/aiohttp/pull/1523

thehesiod:
I've gone ahead and added the warning to your branch.

thehesiod:
after running unittests I found that aiobotocore was broken for two reasons:
1. https://github.com/aio-libs/aiobotocore/pull/74/files was incomplete
2. latest aiohttp broken aiobotocore because it switched to using YARL and assumes that the incoming url is not encoded.

both have been fixed in this PR

PR #88, opened 2016-12-28
Fix timeout.

hellysmile:
1) used aiohttp.Timeout as replacement for asyncio.wait_for. aiohttp.Timeout( https://github.com/KeepSafe/aiohttp/blob/d222f42fe8a25f325df18ee1e191abcb00437d8f/aiohttp/helpers.py#L35 ) is shortcut to the https://github.com/aio-libs/async-timeout
2) set `None` as default timeout for `aiohttp.ClientSession.request`, cuz there are is hardcoded timeout `5*60`. There is no sense to have 2 different, unrelated timeout controls

### DIFF
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -167,12 +167,15 @@ class AioEndpoint(Endpoint):
         headers['Accept-Encoding'] = 'identity'
         headers_ = dict(
             (z[0], text_(z[1], encoding='utf-8')) for z in headers.items())
+
+        timeout = self._conn_timeout + self._read_timeout
+
         request_coro = self._aio_session.request(method, url=url,
-                                                 headers=headers_, data=data)
-        resp = yield from asyncio.wait_for(
-            request_coro, timeout=self._conn_timeout + self._read_timeout,
-            loop=self._loop)
-        return resp
+                                                 headers=headers_, data=data,
+                                                 timeout=None)
+
+        with aiohttp.Timeout(timeout, loop=self._loop):
+            return (yield from request_coro)
 
     @asyncio.coroutine
     def _send_request(self, request_dict, operation_model):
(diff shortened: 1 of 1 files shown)


thehesiod:
this is in conflict with https://github.com/aio-libs/aiobotocore/pull/86

hellysmile:
It is dropin replacment

thehesiod:
per #86 we should be using the read_timeout as specified with through the botocore config variables, not the default 5 minute timeout.

hellysmile:
I am no sure is read_timeout right behaviour, what to deal with conn_timeout? This pr removes default aiohttp 5 sec timeout as well, and sets own which is sum for conn_timeout+read_timeout

thehesiod:
conn_timeout is handled by the TCPConnector

hellysmile:
It is depends, not sure what is right way. This pr or https://github.com/aio-libs/aiobotocore/pull/86 without wait_for is fine.

hellysmile:
okay, I think we MUST include connection timeout in addition to read timeout, cuz time spend by `TCPConnector` is anyway included in entire request time, for `aiohttp.Timeout` there is no difference what exactly spend, connection timeout or entire. Both of them is tracked together

thehesiod:
so here's the relevant code from `aiohttp.client.ClientSession.request`:

```python
            req = self._request_class(
                method, url, params=params, headers=headers,
                skip_auto_headers=skip_headers, data=data,
                cookies=cookies, encoding=encoding,
                auth=auth, version=version, compress=compress, chunked=chunked,
                expect100=expect100,
                loop=self._loop, response_class=self._response_class,
                proxy=proxy, proxy_auth=proxy_auth, timeout=timeout)

            with Timeout(timeout, loop=self._loop):
                conn = yield from self._connector.connect(req)
```

note that timeout is sent to the request_class, which is used for the stream read.  So as long as your read_timeout is > than your connect timeout (which should be default) it should be fine.

thehesiod:
ok, I've merged the other PR so I'm closing this one.  If you don't agree please re-open

PR #91, opened 2017-01-05
fix endpoint to pass verify argument to aiohttp connector

sunghyunzz:
Currently `AioEndpoint` does not consider `verify` value passed when creating `aiohttp.TCPConnector`. We should pass `verfiy` value as `verify_ssl` to `aiohttp.TCPConnector` as expected.


jettify:
Thanks @sunghyunzz @thehesiod  !

PR #99, opened 2017-01-19
Add mock server fixture

jettify:


jettify:
@thehesiod I created mocking server fixture, more then half of tests passes, probably some asserts could be updated to work with both real and mocked server, and rest of the tests disabled. Anyway it is better then nothing.

Issue #69, opened 2016-10-23
document how to get back a file from s3


allan-simon:
I've just helped somebody answering "how to get a file from s3 and iterate over it line by line"  in stackoverflow there 

http://stackoverflow.com/questions/40203620/aiobotocore-aiohttp-get-s3-file-content-and-stream-it-in-the-response/40204293#40204293

and it's only looking in the source code  and this pull request https://github.com/aio-libs/aiobotocore/pull/19 
that's I've been able to answer it.

I think getting a file from s3 is common enough to see it in the readme. if the piece of code I provided in SO is clear enough, I will propose a PR in the readme

asvetlov:
@jettify ?

thehesiod:
added to homepage

allan-simon:
thanks :)

PR #100, opened 2017-01-20
readme updates

thehesiod:
- switch to await syntax
- add examples for pagination and get_object streams


thehesiod:
this closes https://github.com/aio-libs/aiobotocore/issues/69

PR #111, opened 2017-01-30
bump up version

thehesiod:
changes to AioConfig were non-backwards compatible

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,13 @@ install:
 
 script:
     - make flake
+deploy:
+    provider: pypi
+    user: jettify
+    password:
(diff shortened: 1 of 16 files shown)


jettify:
Could you please squash commits? Also what do you think about releasing new version to pypi?

thehesiod:
yes please! would you like to do that or want me try to do it for the first time? I just realized that the current release uses unlimited TCP connections by default which can be bad.  I'm trying to get to the bottom of an aiohttp resource leak I'm seeing

jettify:
Please do it!

jettify:
https://github.com/aio-libs/aiobotocore/releases and then `draft new release`

PR #116, opened 2017-01-31
Normalize headers and handle redirections by botocore

Fedorof:
S3 may return 307 code with redirection to another server in case of high load or maintenance.

I did not manage to reproduce this issue but it is simulated by creating an object in one region and fetching it from another. In this case S3 returns 301 response. I would appreciate if you advise how to reproduce it with `moto`.

S3 do not provide `Location` or `URI` header, so the redirection should be handled by the application (`botocore` in our case).

An example of the response headers:

    HTTP/1.1 301 Moved Permanently
    x-amz-bucket-region: us-east-1
    x-amz-request-id: CD12xxxxxxxxxxxx
    x-amz-id-2: IVdExxxxxxxxxxxxxxxxxxxxxxx
    Content-Type: application/xml
    Transfer-Encoding: chunked
    Date: Mon, 30 Jan 2017 13:07:10 GMT
    Server: AmazonS3

As you can see, no `Location` or `URI` header is provided. All redirect information is listed in the body:

    <Error>
      <Code>PermanentRedirect</Code>
      <Message>The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.</Message>
      <Bucket>foo</Bucket>
      <Endpoint>s3.amazonaws.com</Endpoint>
      <RequestId>CD12xxxxxxxxxxxx</RequestId>
      <HostId>IVdExxxxxxxxxxxxxxxxxxxxxxx</HostId>
    </Error>

For 307 error the response is similar:

    <Error>
      <Code>TemporaryRedirect</Code>
      <Message>Please re-send this request to the specified temporary endpoint. Continue to use the original request endpoint for future requests.</Message>
      <RequestId>63B2xxxxxxxxxxxx</RequestId>
      <Bucket>foo</Bucket>
      <HostId>S3nuxxxxxxxxxxxxxxxxxxxxxxx</HostId>
      <Endpoint>s3-us-west-2.amazonaws.com</Endpoint>
    </Error>

There are 2 issues with `aiobotocore`:
* `allow_redirects` is set to `True` in `aiohttp` (the default value) so that it fails because can't find redirect location in the headers.
* `multidict` tries to use case-insensitive strings for headers but `botocore` converts them to regular strings. That strings are in mixed case while `botocore` expects headers to be in lower case.

### DIFF
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -193,7 +193,8 @@ class AioEndpoint(Endpoint):
                                                     headers=headers_,
                                                     data=data,
                                                     proxy=proxy,
-                                                    timeout=self._read_timeout)
+                                                    timeout=self._read_timeout,
+                                                    allow_redirects=False)
         return resp
 
     @asyncio.coroutine
@@ -298,6 +299,16 @@ class AioEndpoint(Endpoint):
             operation_model.metadata['protocol'])
         parsed_response = parser.parse(
             response_dict, operation_model.output_shape)
+
+        # Workaround to provide lowercase headers to botocore for redirects
+        # TODO: remove this code block when
+        # https://github.com/boto/botocore/pull/1145 is merged
+        if 300 <= resp.status_code < 400:  # pragma: no cover
+            parsed_response['ResponseMetadata']['HTTPHeaders'] = {
+                k.lower(): v for k, v in
+                parsed_response['ResponseMetadata']['HTTPHeaders'].items()
+            }
+
         return (http_response, parsed_response), None
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -71,7 +71,12 @@ def random_tablename():
 
 
 def random_name():
-    table_name = 'aiobotocoretest_{}'
+    """Return a string with presumably unique contents
+
+    The string contains only symbols allowed for s3 buckets
+    (alpfanumeric, dot and hyphen).
+    """
+    table_name = 'aiobotocoretest-{}'
     t = time.time()
     return table_name.format(int(t))
 
@@ -139,6 +144,11 @@ def region():
 
 
 @pytest.fixture
+def alternative_region():
+    return 'us-west-2'
+
+
+@pytest.fixture
 def signature_version():
     return 's3'
 
@@ -170,6 +180,22 @@ def s3_client(request, session, region, config, s3_server, mocking_test):
     if mocking_test:
         kw = moto_config(s3_server)
     client = create_client('s3', request, session, region, config, **kw)
+    return client
+
+
+@pytest.fixture
+def alternative_s3_client(request, session, alternative_region,
+                          signature_version, s3_server,
+                          mocking_test):
+    kw = {}
+    if mocking_test:
+        kw = moto_config(s3_server)
+
+    config = AioConfig(
+        region_name=alternative_region, signature_version=signature_version,
+        read_timeout=5, connect_timeout=5)
+    client = create_client(
+        's3', request, session, alternative_region, config, **kw)
     return client
 
 
--- a/tests/test_basic_s3.py
+++ b/tests/test_basic_s3.py
@@ -374,3 +374,20 @@ def test_presign_sigv4(s3_client, bucket_name, aio_session, create_object):
     resp = yield from aio_session.get(presigned_url)
     data = yield from resp.read()
     assert data == b'foo'
+
+
+@pytest.mark.parametrize('signature_version', ['s3v4'])
+@pytest.mark.parametrize('mocking_test', [False])
+@pytest.mark.run_loop
+def test_can_follow_signed_url_redirect(alternative_s3_client,
+                                        create_object, bucket_name, loop):
+    yield from create_object('foobarbaz')
+    # Eventual consistency.
+    yield from asyncio.sleep(3, loop=loop)
+
+    # Simulate redirection by provide wrong endpoint intentionally
+    resp = yield from alternative_s3_client.get_object(
+        Bucket=bucket_name, Key='foobarbaz')
+    data = yield from resp['Body'].read()
+    resp['Body'].close()
+    assert data == b'foo'


tests/conftest.py
```diff
@@ -71,7 +71,12 @@ def random_tablename():
 
 
 def random_name():
-    table_name = 'aiobotocoretest_{}'
+    """Return a string with presumably unique contents
+
+    The string contains only symbols allowed for s3 buckets
+    (alpfanumeric, dot and hyphen).
+    """
+    table_name = 'aiobotocoretest-{}'
```
Fedorof: Underscore in a bucket name is prohibited for all regions except `us-east-1`

jettify: good point!

aiobotocore/endpoint.py
```diff
@@ -39,7 +39,9 @@ def text_(s, encoding='utf-8', errors='strict'):
 @asyncio.coroutine
 def convert_to_response_dict(http_response, operation_model):
     response_dict = {
-        'headers': http_response.headers,
+        # botocore converts keys to str, so make sure that they are in
+        # the expected case.
+        'headers': {k.lower(): v for k, v in http_response.headers.items()},
```
jettify: do you know why they do this? headers are case insensitive by standard...

jettify:
@thehesiod please review too

aiobotocore/endpoint.py
jettify: never mind, good point

Fedorof: If I don't miss anything, headers are converted to strings here: https://github.com/boto/botocore/blob/76866243555eead44fb8d0cf709ffed2b823a003/botocore/parsers.py#L218

And then `botocore` fails to find the region here: https://github.com/boto/botocore/blob/bb60fedc64bae2c86934067fea563f52db232a73/botocore/utils.py#L940

jettify:
From my side LGTM

aiobotocore/endpoint.py
thehesiod: can you please add a comment with a link to the code that does the equivalent in botocore?

Fedorof: Do you want me to add a link provided above?

thehesiod:
so this is interesting in terms of testing in that if we switch to moto we need to make sure that all redirect cases work correctly (moto reproduces the redirects)...if not either moto will need to be enhanced or we'll need to continue doing several tests with a real AWS server.  I know redirects also happen when accessing an S3 via virtual/path styles depending on the bucket name you may get a redirect: http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html (IIRC on buckets with periods).  We should ensure this still works correctly if botocore doesn't pre-calculate the endpoint correctly.

aiobotocore/endpoint.py
thehesiod: that but also where they require `Location` to have a specific case

thehesiod: is this it? https://github.com/boto/botocore/blob/76866243555eead44fb8d0cf709ffed2b823a003/botocore/vendored/requests/sessions.py#L119

thehesiod: also now I want to know how in botocore the request got lowercase, need link to that as well

thehesiod: looks like maybe the request module stores them lowercase?  looks like this was introduced here: https://github.com/boto/botocore/pull/962

thehesiod: FYI: from a simple test it doesn't seem dict()'ing a requests response will cause them to go lowercase...so somewhere else this must be happening

Fedorof: `Location` does not go lowercase, S3 simply does not provide this header. 

And yes, the issue is that dict()'ing a request response will not cause the headers to go lowercase, while `botocore` expects them to be lowercased

jettify:
Moto is a cheep way to test aiobotocore on CI, we should execute tests AWS regardless of moto.

`botocore` uses regular dict for header storrage, when aiohttp special dict with case insensitive keys, as result there is no options for them but to transform, so probably not bug but their internal implementation.

thehesiod:
ohh actually I know what the issue is.  aiohttp's multidict ends up CamelCasing the headers...I saw this awhile ago.  We should probably log that bug, let me dig up the details.

thehesiod:
logged https://github.com/KeepSafe/aiohttp/issues/1570, I do not think this is the correct fix as the real issue is that the casing of the headers is not preserved by aiohttp.  I'm not sure botocore will expect all headers to be lower case, the correct fix is for aiohttp to preserve the casing of the headers and still store results in a multidict.

kakovskyi:
@thehesiod that's sad to hear, because `aiobotocore` library users won't have the S3 redirection feature working until the error is fixed in aiohttp and we have the new aiohttp version released.

Could you please consider to apply current solution?

thehesiod:
I think if you limit the change to the "Location" header that would be a good work-around until the aiohttp issue is resolved, what do you think?

jettify:
@thehesiod I think current solution is fine, better to have working redirect, then not.

thehesiod:
I'm worried if anything else breaks based on this as it's wholesale lowercasing all headers.  If all aws headers are guaranteed to be lowercase I'd be fine with this PR added a comment and link. Can anyone find documentation about it?

jettify:
This is internal implementation detail of botocore, we violated all internal interfaces anyway.

thehesiod:
I'm scared, but if jettify is ok with it I'd like a comment added about the real issue I found, and saying this should be removed after the aiohttp issue resolved and this this assumes that all AWS headers are being returned in lowercase...otherwise it would break botocore expectations of the casing of the headers (for example, what if for some other header the AWS REST API returns a header with an upper case letter?  This will now have broken that look-up.)  I doubt anyone in this PR as validated that botocore expects that _all_ headers be lower-case...if that is the case then this PR is fine with a comment too.

thehesiod:
otherwise we're no better than aiohttp and we're causing information loss.  The ultimate consumer should make the decision of case (in this case botocore).  aiohttp/requests/aiobotocore should not be mucking with the casing.

jettify:
It is a bit adventurous, but lets release, new version, and see if users start to complain :)

kakovskyi:
@jettify @thehesiod thank you, guys! we will let you know if observe any issues with the new release

thehesiod:
lol, ok, I have a feeling we'll be referring to this PR in the future =)

thehesiod:
sorry guys, have a long history with stuff like this, worked 15 years @ Adobe, lots of history there and changes like these tended to bite us in the ass later :)

thehesiod:
also another FYI why I'm scared, note http://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html where its showing response headers in camel case

jettify:
@Fedorof @thehesiod what do you think about following idea?

So in aiobotocore here: https://github.com/aio-libs/aiobotocore/blob/master/aiobotocore/endpoint.py#L299
we change` parsed_response['ResponseMetadata']['HTTPHeaders']` to `MultiDict` instead of regular dict. `MultiDict` is regular dict but has case insensative keys,  so `"key" in headers` and `"Key" in headers` returns same value.

jettify:
I think botocore should have done this long time ago

thehesiod:
@jettify can you check to see if it passes this case: https://github.com/boto/botocore/pull/962.

jettify:
Very good catch! So they used case insensitive dict in first place

jettify:
Do we care that `parsed_response['ResponseMetadata']['HTTPHeaders']` will not be json serializable?

thehesiod:
ooo, check ClientResponse. raw_headers!

jettify:
@thehesiod nice! This is our one line fix)

thehesiod:
almost :) I think you can use:

```python
{k.('utf-8', 'surrogateescape': v.('utf-8', 'surrogateescape' for (k, v) in response.raw_headers}
```

per https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/protocol.py#L123

jettify:
@Fedorof could you please update PR with this line?

thehesiod:
also please test it works as expected =)

jettify:
@Fedorof @thehesiod  @kakovsky this was fun :) Thanks everyone!

Fedorof:
Could you please add that change to the PR for me? It's 1:10 AM here and I can't do it till morning, and it would be great to have this change released ASAP

kakovskyi:
@Fedorof take your time, I will try to care about that

jettify:
I am in same timezone with you, will do release in the morning if required.

kakovskyi:
I ran application tests with the latest changes and found a lot of failures like:
```
>       self.assertEqual(document['size'], metadata['ContentLength'])
E       KeyError: 'ContentLength'
```
Probably, due to case-sensitive nature of dict.

jettify:
Tested original fix, it works as expected with aws, 
`raw_headers` do not work.
```
(Pdb) resp._response.raw_headers
(bytearray(b'X-AMZ-REQUEST-ID'), bytearray(b'')), 
(bytearray(b'CONTENT-LENGTH'), bytearray(b'')),
(bytearray(b'SERVER'), bytearray(b''))
```
I will look on this more later, now had to go

Fedorof:
Updated the PR.

Now the headers are converted to lowercase *after* they are processed by the `botocore` parser and only when redirect is required, so for normal situations users should not see the difference.

There is a PR for `botocore` to normalize the headers as well: https://github.com/boto/botocore/pull/1145

kakovskyi:
@Fedorof @jettify good job!

thehesiod:
I'll do some investigation as well, if raw-headers doesn't work somewhere in botocore they must be calling .lower()

jettify:
@thehesiod problem with `raw_headers` that case is not preserved there, see example
```
(bytearray(b'X-AMZ-REQUEST-ID'), bytearray(b'')),
```

thehesiod:
ahh, look at what aiohttp does: https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/protocol.py#L87  what a mess.  I updated my aiohttp bug.

jettify:
@thehesiod I propose to merge  this PR (we change header only in case of 3xx anyway) since it looks like working, I tested with AWS. Then create issue to track headers issue.

thehesiod:
one sec, I'm digging into botocore to see what happens on redirect

jettify:
sure

thehesiod:
I'm really close, I think the original fix is going to be ok as I can see them lower()'ing the header names, as soon as I get when it's done we can update the comment and do the original fix

thehesiod:
headers first come in here: /usr/local/Cellar/python3/3.5.2_3/Frameworks/Python.framework/Versions/3.5/lib/python3.5/http/client.py in HTTPResponse.begin() as mixture of CamelCase from server, and lower-case.  It then gets put in a HTTPHeaderDict (maintains original case) in a botocore.vendored.requests.packages.urllib3.reponse.HTTPResponse in /usr/local/lib/python3.5/site-packages/botocore/vendored/requests/packages/urllib3/connectionpool.py:HTTPConnectionPool.urlopen.  

Thank here: https://github.com/boto/botocore/blob/develop/botocore/vendored/requests/packages/urllib3/response.py#L324 it gets wrapped in a HTTPHeaderDict, which when adding the headers lowers() everything: https://github.com/boto/botocore/blob/develop/botocore/vendored/requests/packages/urllib3/_collections.py#L216

mystery solved!  so I don't think any changes to botocore are required and we should MOST DEFINITELY do the original fix of lower()ing all the keys.

jettify:
@thehesiod thanks for getting to the bottom!

jettify:
i will do merge manually

thehesiod:
can you add a big comment about why we lower?  thanks!

thehesiod:
thanks @Fedorof for getting us to investigate this :)

thehesiod:
Here's the callstack for those interested:

```
  File "/Users/amohr/Library/Preferences/PyCharm2016.3/scratches/scratch_2.py", line 10, in main
    client.get_object(Bucket='insights.fbn.org', Key='prefix.txt')
  File "/usr/local/lib/python3.5/site-packages/botocore/client.py", line 253, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/local/lib/python3.5/site-packages/botocore/client.py", line 530, in _make_api_call
    operation_model, request_dict)
  File "/usr/local/lib/python3.5/site-packages/botocore/endpoint.py", line 141, in make_request
    return self._send_request(request_dict, operation_model)
  File "/usr/local/lib/python3.5/site-packages/botocore/endpoint.py", line 168, in _send_request
    request, operation_model, attempts)
  File "/usr/local/lib/python3.5/site-packages/botocore/endpoint.py", line 204, in _get_response
    proxies=self.proxies, timeout=self.timeout)
  File "/usr/local/lib/python3.5/site-packages/botocore/vendored/requests/sessions.py", line 573, in send
    r = adapter.send(request, **kwargs)
  File "/usr/local/lib/python3.5/site-packages/botocore/vendored/requests/adapters.py", line 370, in send
    timeout=timeout
  File "/usr/local/lib/python3.5/site-packages/botocore/vendored/requests/packages/urllib3/connectionpool.py", line 556, in urlopen
    **response_kw)
  File "/usr/local/lib/python3.5/site-packages/botocore/vendored/requests/packages/urllib3/response.py", line 324, in from_httplib
    headers = HTTPHeaderDict(headers.items())
  File "/usr/local/lib/python3.5/site-packages/botocore/vendored/requests/packages/urllib3/_collections.py", line 147, in __init__
    self.extend(headers)
  File "/usr/local/lib/python3.5/site-packages/botocore/vendored/requests/packages/urllib3/_collections.py", line 251, in extend
    self.add(key, value)
  File "/usr/local/lib/python3.5/site-packages/botocore/vendored/requests/packages/urllib3/_collections.py", line 216, in add
    key_lower = key.lower()
```

Fedorof:
The original change could be a breaking one, though, since users may relay on the fact that `aiobotocore` returns headers as case-insensitive `dict`s.

thehesiod:
are headers exposed to users?  from what I know only the parsed response object is

jettify:
it exposed as request metadata

jettify:
change in https://github.com/aio-libs/aiobotocore/commit/e44d9ba6308ac745907f6b4bcd192e4b02636779 
:-)

jettify:
Will do release soon.

thehesiod:
either way, we'll now match botocore as they lower() it before returning to users, whatever we were doing before was wrong :)

jettify:
New version released https://pypi.python.org/pypi/aiobotocore/0.2.1

Fedorof:
Thank you @thehesiod , @jettify . That was awesome :)

thehesiod:
bad news guys, this broke head_object, if you do head_object it won't return the ETag. I've verified it works if we revert the change.  Tomorrow I'll have to do a deep dive to figure out what's going on.

thehesiod:
can't fix this until https://github.com/KeepSafe/aiohttp/issues/1570 is fixed, have pending PR: https://github.com/KeepSafe/aiohttp/pull/1698.  After that is fixed we can switch to raw_headers and use botocore's HttpHeaderDict

PR #119, opened 2017-02-04
Fix aiohttp warnings

jettify:


thehesiod:
i was just looking at this recently, not sure if it would make things easier: https://github.com/pytest-dev/pytest-asyncio

jettify:
Warning is not pytest related, aiohttp requires loop to be created inside coroutine and `close()` should be awaited too in new versions.

PR #127, opened 2017-02-22
Configure bdist_wheel in travis fixes #58

### LINKED ISSUE #58: release py3 wheels

graingert:


graingert:
@jettify you should configure a different user with maintainer only access rights to aiobotocore, rather than using your own password in your travis config.

jettify:
github/travis provides capability to do deployment, without any need to touch pypi, do not see any reason to do deployment by hand.

graingert:
Sure but Travis only needs maintainership rights to one pypi project. No need to give it your account with ownership access to many projects

Issue #131, opened 2017-03-07
head_object is broken


thehesiod:
it will not return the ETag, the issue is due to https://github.com/aio-libs/aiobotocore/pull/116.  botocore expect the ETag to be in the correct case.  We can't fix this until https://github.com/KeepSafe/aiohttp/issues/1570 is fixed. After that is fixed we can switch to raw_headers and use botocore's CaseInsensitiveDict

thehesiod:
@Fedorof @jettify I knew this was going to bite us =)

jettify:
what a pain, headers supposed to be case insensitive...

PR #132, opened 2017-03-07
fixes header case and brings in line with botocore

thehesiod:
this fixes https://github.com/aio-libs/aiobotocore/issues/131 by using the same case insensitive dict that botocore uses.


thehesiod:
I'm merging ignoring the delta coverage for endpoint.py since I just moved code.  This code is copied from botocore and we never had tests for them before.

PR #141, opened 2017-03-22
upgrade botocore dep

graingert:


graingert:
@jettify any reason this isn't a range?

thehesiod:
see https://github.com/aio-libs/aiobotocore/blob/master/BOTOCORE_README.rst

graingert:
@thehesiod ah major bummer.

graingert:
That basically leaves aiobotocore unusable with pip-compile

jettify:
it is possible to unlock, we just need to test each release to make sure it is working as expected.

thehesiod:
I wouldn't unlock as it's important to let people know what version this has been approved with.  I think we're just stuck with upgrading things when needed or vendoring in botocore and its deps.

python module versioning IMHO is broken when you have an app that pulls in a lot of modules...lots of competing version requirements.  They should instead have allowed for a local store that supports more than one version.

thehesiod:
another reason not to unlock is because if you want to install an old version of aiobotocore you want it to downgrade the botocore to what's required.

graingert:
Hence it would be way better to vendor it

jettify:
I mean we still lock aiobotocore like `"aiobotocore>=1.5.0, <1.5.15"`, but increment version if no significant changes for us and tests are green. this approach requires more work thou

jettify:
Anyway best solution is to figure out what public interface we need, and work with botocore maintainers on implementation, so in future we can be more resistant to library changes.

thehesiod:
ya, I'm just saying if we're going to vendor we should go the whole way and vendor deps of deps so we don't have any surprises :)  Short term we should just upgrade the botocore integration, no biggie. opened: https://github.com/aio-libs/aiobotocore/issues/143

thehesiod:
interestingly I was just noticing that even botocore has two layers of vendored stuff in it, lol: botocore.vendored.requests.packages.urllib3

so that's botocore vendored requests, which itself vendors urllib3, what a mess.

graingert:
Clearly this can only be solved with even deeper nesting

thehesiod:
I want to avoid the whole vendoring thing if we can, it's more work which can be handled by the client.  Production apps always version lock everything, including deps of deps due to versioning hell.

thehesiod:
@graingert FYI latest aiobotocore now supports 1.5.0 - 1.5.28 to help in integration.

PR #152, opened 2017-03-27
Update for aiohttp 2.0+

achimnol:
* All tests are green.
 * ProxyConnectionError is reported as ClientConnectorError.
 * aiohttp now supports separated read/connection timeouts.

### DIFF
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -190,6 +190,7 @@ class AioEndpoint(Endpoint):
                                                     headers=headers_,
                                                     data=data,
                                                     proxy=proxy,
+                                                    timeout=None,
                                                     allow_redirects=False)
         return resp
 
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -1,6 +1,5 @@
 import asyncio
 import sys
-import warnings
 import yarl
 import aiohttp
 import botocore.retryhandler
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -11,7 +11,7 @@ from botocore.exceptions import EndpointConnectionError, ConnectionClosedError
 from botocore.hooks import first_non_none_response
 from botocore.utils import is_valid_endpoint_url
 from botocore.vendored.requests.structures import CaseInsensitiveDict
-from aiohttp import MultiDict
+from multidict import MultiDict
 from urllib.parse import urlparse
 
 PY_35 = sys.version_info >= (3, 5)
@@ -19,10 +19,10 @@ PY_35 = sys.version_info >= (3, 5)
 # Monkey patching: We need to insert the aiohttp exception equivalents
 # The only other way to do this would be to have another config file :(
 _aiohttp_retryable_exceptions = [
-    aiohttp.errors.ClientConnectionError,
-    aiohttp.errors.TimeoutError,
-    aiohttp.errors.DisconnectedError,
-    aiohttp.errors.ClientHttpProcessingError,
+    aiohttp.ClientConnectionError,
+    aiohttp.ServerDisconnectedError,
+    aiohttp.http_exceptions.HttpProcessingError,
+    asyncio.TimeoutError,
 ]
 
 botocore.retryhandler.EXCEPTION_MAP['GENERAL_CONNECTION_ERROR'].extend(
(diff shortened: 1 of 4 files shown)


jettify:
thanks!  I will take a look tonight

aiobotocore/endpoint.py
```diff
@@ -11,18 +10,18 @@
 from botocore.hooks import first_non_none_response
 from botocore.utils import is_valid_endpoint_url
 from botocore.vendored.requests.structures import CaseInsensitiveDict
-from aiohttp import MultiDict
+from multidict import MultiDict
```
jettify: Lets add `multidict` to setupy.py as requirement

tests/test_basic_s3.py
```diff
@@ -30,7 +30,7 @@ def test_can_make_request(s3_client):
 def test_fail_proxy_request(aa_fail_proxy_config, s3_client):
     # based on test_can_make_request
 
-    with pytest.raises(ProxyConnectionError):
+    with pytest.raises(aiohttp.ClientConnectorError):
```
jettify: May be  `ClientHttpProxyError` should be used instead?

aiobotocore/endpoint.py
```diff
@@ -183,14 +182,6 @@ def _request(self, method, url, headers, data):
         headers_ = MultiDict(
             (z[0], text_(z[1], encoding='utf-8')) for z in headers.items())
 
-        # For now the request timeout is: read_timeout and max(conn_timeout)
-        # So we want to ensure that your conn_timeout won't get truncated by
-        # the read_timeout. This should be removed after
-        # (https://github.com/KeepSafe/aiohttp/issues/1524) is resolved
```
jettify: @thehesiod can you please review this part?

jettify:
looks good, few minor things to fix

aiobotocore/endpoint.py
thehesiod: @achimnol nice, thanks!

aiobotocore/endpoint.py
```diff
@@ -199,7 +190,6 @@ def _request(self, method, url, headers, data):
                                                     headers=headers_,
                                                     data=data,
                                                     proxy=proxy,
-                                                    timeout=self._read_timeout,
```
thehesiod: @achimnol  this is interesting, should we be removing it or setting it to `None` given we're setting the read + conn timeouts on the session now? I think the default was 5 minutes. Along those lines I've opened issue: https://github.com/aio-libs/aiohttp/issues/1757.

thehesiod: I think actually we have to set it to `None` or else if the read/conn timeouts are > 5 minutes it will get truncated.

tests/test_basic_s3.py
achimnol: During my tests, the exception thrown to the test case was `ClientConnectorError`. The docstring of `ClientConnectorError` describes its purpose as:
```
Raised in :class:`aiohttp.connector.TCPConnector`
if connection to proxy can not be established.
```

I have also tried `aiohttp.ClientProxyConnectionError` but it was raised as the part of nested exception chain and not caught by the test code.

Exception chain:
* `ConnectionRefusedError` (asyncio/selector_events.py, line 481) ->
* `aiohttp.client_exceptions.ClientConnectorError` (aiohttp/connector.py, line 678) ->
* `aiohttp.client_exceptions.ClientProxyConnectionError` (aiohttp/connector.py, line 692) ->
* `aiohttp.client_exceptions.ClientConnectorError` (aiohttp/connector.py, line 379) ->
* test code

aiobotocore/endpoint.py
achimnol: Agree. I will add it at the next commit.

aiobotocore/endpoint.py
achimnol: Ah, I missed that point. I will set it to None in the next commit.

tests/test_basic_s3.py
jettify: got it

jettify:
@achimnol could you please rebase your branch against master?

achimnol:
I have rebased against master, applying botocore version updated to 1.5.29.

achimnol:
Thanks! 👍

Issue #161, opened 2017-03-31
support waiters


Ludachrispeed:
It seems there's currently no support for creating waiters with aiobotocore clients (for example, see EC2 waiters: https://botocore.readthedocs.io/en/latest/reference/services/ec2.html#waiters)

Since botocore waiters necessarily block for a _long_ time, it would be especially cool if aiobotocore could support them!

thehesiod:
could you give some example code that shows the error?

Ludachrispeed:
@thehesiod Sure thing. I haven't used moto but hopefully these examples show what I'm trying to do:

1. Basic example of using a botocore waiter. No setup required; this just creates one SQS queue, so it's not too expensive (this works): [Gist][1]
2. Same example using standard asyncio (this works): [Gist][2]
3. And finally here is the same example using `aiobotocore`, along with the error that happens when you run this (this doesn't work): [Gist][3]

[1]: https://gist.github.com/Ludachrispeed/78f94885f5bc6b66c475eb199d3ed968
[2]: https://gist.github.com/Ludachrispeed/325c3efa2080696c6baa345bbb29aa4d
[3]: https://gist.github.com/Ludachrispeed/51ea85ff13cf55748188aba57125949e

thehesiod:
perfect thanks so much!  We'll take a look asap

thehesiod:
@Ludachrispeed mind testing out this PR?

thehesiod:
just need to release a new version after confirmed by @Ludachrispeed

Ludachrispeed:
This is working great! Thanks for this feature @thehesiod 

For posterity, here's an example that worked for me:
```
import asyncio
import aiobotocore
import string
import random

loop = asyncio.get_event_loop()
session = aiobotocore.get_session(loop=loop)
ec2_client = session.create_client('ec2', region_name='us-east-1')
s3_client = session.create_client('s3')


async def create_ec2_instance():
    print('starting ec2 run_instance')
    resp = await ec2_client.run_instances(
        ImageId='ami-22ce4934',
        InstanceType='t2.nano',
        MinCount=1,
        MaxCount=1,
    )
    waiter = ec2_client.get_waiter('instance_running')
    print('waiting for all ec2 instances to be running')
    await waiter.wait(InstanceIds=[resp['Instances'][0]['InstanceId']])
    print('ec2 instances are running!')


async def create_s3_bucket():
[...]

PR #180, opened 2017-04-14
fix Waiter support

thehesiod:
fixes https://github.com/aio-libs/aiobotocore/issues/161


thehesiod:
logged https://github.com/aio-libs/aiobotocore/issues/181 to add waiter tests as it seems like it's going to be fairly involved, I'm guessing it will be something like the dynamodb test

PR #223, opened 2017-05-22
integration and stream fix

thehesiod:
This fixes the last botocore integration and also has a provisional fix for github.com/aio-libs/aiohttp/issues/1907.  I still firmly believe aiohttp should revert back to 1.x behavior for streams.  Until then this will fix retries.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -2,6 +2,8 @@ Changes
 -------
 0.3.2 (XXXX-XX-XX)
 ^^^^^^^^^^^^^^^^^^^^
+* Fix botocore integration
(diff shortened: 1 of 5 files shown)


thehesiod:
@jettify please don't bump botocore version w/o doing a source comparison between botocore versions.  Note how you missed the new `auth_type` parameter in client.py.  Passing tests does not mean good integration :)

jettify:
Sorry, I usually look on change log and try to update if only json (with api definition) updated.

thehesiod:
I'm not very familiar with CI, can we have it check changes to specific files?  That would alleviate this issue.

Issue #245, opened 2017-06-05
[Tracking] Config read_timeout not behaving correctly


thehesiod:
We just ran into this due to: https://github.com/aio-libs/aiohttp/issues/1954.  Here until we have a resolution to that issue.  To work around this we'd have to disable read_timeout, and re-impl read() to have a timeout per chunk.

thehesiod:
Unfortunately one of the aiohttp devs doesn't understand both precedence and the literal meaning of read_timeout so we'll have to re-implement this ourselves. I'll work on this today.  Many people are going to need similar functionality (including me in my other projects) so I'm going to see if we can allow users to use the fix for their own usage of aiohttp.   I'm not 100% certain we'll be able to get this to work cleanly as IIRC they create the stream class from cython without backref to the request.  Based on the bizarre design decisions being made in aiohttp: this, the stream `close` issue, the odd multidict camel casing of headers, and various strange behaviors found in the past I'm really almost to the point of creating a separate core impl of aiohttp.

adamrothman:
Man that guy is being such a pain in the ass. Thanks for fighting the good fight @thehesiod!

thehesiod:
@adamrothman  I don't know how many ways I can prove something is wrong, lol

thehesiod:
re-opening until ideally aiohttp supports this natively

thehesiod:
update: asvetlov said he'd impl it for is

thehesiod:
update on this issue, support for timeouts was added to master, just waiting on a release with the feature.

@samuelcolvin as soon as the next aiohttp release happens we'll have a medium sized change to support it and remove a lot of workarounds we added to support proper client timeouts.  Hopefully soon!  You could try however no guarantees it will work.  If you'd like to make sure it will work correctly there are some instructions here of what's required: https://github.com/aio-libs/aiobotocore/blob/master/CONTRIBUTING.rst#how-to-upgrade-botocore

samuelcolvin:
woop. Thank you.

thehesiod:
it has for awhile, we just forgot to close this ;)

PR #248, opened 2017-06-09
resolve read_timeout

thehesiod:
- fixes https://github.com/aio-libs/aiobotocore/issues/245
- cleanup by moving more code to use wrapt
- implement set_socket_timeout


thehesiod:
hey Jettify, let me know what you think.  All this because aiohttp dev didn't want to fix a core issue in their impl :(

jettify:
I think aiohttp will do right thing eventually, we just need to make sure we can remove it easily and be very careful for each aiohttp release.

By the way next aiohttp will have new multi dict that preserves original headers

thehesiod:
We shall see since I can't drive that anymore given his response.  Should I merge this PR or would you like me change something? Thanks!

jettify:
change looks good, it uses a bit internal API but I do not see better way. We can merge, but lets do not close #245 at some point we need revert this PR once aiohttp ready

thehesiod:
@jettify take a look at the last commit :)  Will make our lives easier

Issue #267, opened 2017-06-26
[Tracking] Getting TemporaryRedirect when using aiobotocore but not when using botocore


chuck1:
I have a recently created a new bucket and am trying to read and write to it. If I understand correctly, when a bucket is created, you can get TemporaryRedirect errors for a while. But I have found that I only get that error when using aiobotocore, not when using botocore. Below is my actual test program. The first half, the botocore part, runs with no problems and I get good responses. But the put_object call in the aiobotocore part raises ``An error occurred (TemporaryRedirect) when calling the PutObject operation: Please re-send this request to the specified temporary endpoint. Continue to use the original request endpoint for future requests.`` I am using aiobotocore version 0.4.0 and python 3.5.

I thought that the two should produce the same result. What could be going on here?

```python
import asyncio
import os
import sys

import botocore.session

conf = # object with my access and secret keys and other info

session = botocore.session.get_session()

client = session.create_client(
        's3', 
        region_name=conf.REGION,
        aws_secret_access_key=conf.SECRET_ACCESS_KEY,
        aws_access_key_id=conf.ACCESS_KEY_ID)

resp = client.put_object(
        Bucket=conf.BUCKET,
        Key='test',
        Body=b'hello')

print(resp)
print()

resp = client.get_object(
        Bucket=conf.BUCKET,
        Key='test')

print(resp)
print()


import aiobotocore

loop = asyncio.get_event_loop()

session = aiobotocore.session.get_session()

async def go():

    async with session.create_client(
            's3', 
            region_name=conf.REGION,
            aws_secret_access_key=conf.SECRET_ACCESS_KEY,
            aws_access_key_id=conf.ACCESS_KEY_ID) as client:

        resp = await client.put_object(
                Bucket=conf.BUCKET,
                Key='test',
                Body=b'hello')

        print(resp)
        print()
        
        resp = await client.get_object(
                Bucket=conf.BUCKET,
                Key='test')

        print(resp)
        print()

loop.run_until_complete(go())
```

thehesiod:
interesting, looking into this

thehesiod:
btw for tests like these I suggest using `AWS_PROFILE`: http://boto3.readthedocs.io/en/latest/guide/configuration.html#shared-credentials-file

so I just repro'd with this more complete test:
```python
import asyncio
import botocore.session
import aiobotocore

TEST_BUCKET = 'thehesiod-dummy2'
TEST_KEY = 'blah'
[...]

PR #268, opened 2017-06-27
fix redirects

thehesiod:
- implement workaround to
https://github.com/aio-libs/aiobotocore/issues/267
- add new unittest
- remove sleeps as they're not needed anymore since all regions are read after write consistent


thehesiod:
unfortunately these tests require the real AWS testcases and thus don't run during our normal CI, hence the coverage failures.

thehesiod:
@jettify let me know what you think.  Basically I copied the code from aiohttp until we have native support.

jettify:
Looks good, lests leave issue opened to remove workaround once aiohttp ready.

PR #270, opened 2017-06-27
re-ordered example so it runs

dfee:
Previously, `get_object` was run after `delete_object` raising a Key Error as the resource no longer exists.


jettify:
Thanks @dfee @thehesiod !

PR #273, opened 2017-06-30
add the supported versions of awscli to extras require

graingert:
Currently the biggest pain-point for me using aiobotocore is that I also use awscli.
Whenever I upgrade aiobotocore I have to go find out what version of botocore it supports,
then what version of awscli supports that version of botocore.

Eg currently I have:

awscli == 1.11.107 # for aiobotocore/botocore version compat
aiobotocore >= 0.4.1, < 0.5.0

but I'd like to have:

aiobotocore[awscli] >= 0.4.1, < 0.5.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ def read(f):
     return open(os.path.join(os.path.dirname(__file__), f)).read().strip()
 
 
-extras_require = {}
+extras_require = {'awscli': ['awscli>=1.11.108, <=1.11.115']}
 
 
 def read_version():


jettify:
No big problem to add this change, I worry a bit how we can test that those libraries has compatible botocore version, is there any automatic or semi automatic way?

graingert:
@jettify awscli only support exactly one version of botocore

thehesiod:
presumably we'd have to do this manually with each release

jettify:
But if we remove version of awscli from setup.py, pip should match compatible version, correct? or I missed something?

graingert:
> But if we remove version of awscli from setup.py, pip should match compatible version, correct? or I missed something?

Eh? If you remove it and people depend on aiobotocore[awscli] you'll break their install

thehesiod:
btw @graingert I'm about to bump botocore reqs, see https://github.com/aio-libs/aiobotocore/pull/278  can you update this to that?

jettify:
Tried to install awscli, see your point, I thought pip is a bit smarter.

jettify:
Should be strait forward to maintain since they have one to one dependency

graingert:
@thehesiod done

thehesiod:
thanks @graingert !  did you need a release with this?  I forgot to mention about adding change to CHANGES.txt, I'll add it in master.

graingert:
@thehesiod yeah a release would be nice

thehesiod:
done

PR #278, opened 2017-07-03
add support for latest aiohttp/botocore

thehesiod:


thehesiod:
@jettify can we update the unittests to run against multiple versions of botocore?  Perhaps earliest and latest releases supported.  that would expand the test coverage.

jettify:
Can do that, but should we just set minimal version `1.5.71` and max `1.5.78`? I mean for `<=1.5.70` users may use older `aiobotocore` versions... Feature vise `aiobotocore` is same, just different supported `botocore` versions.

thehesiod:
ah, so we don't need the hack?  ya I can do that

thehesiod:
btw I'm working with asvetlov on aiohttp work-around, next release should have the fix for 3.4

jettify:
I am fine with either your solutions, my point having less code is better and easier to maintain :)

thehesiod:
going to delay this till we get 2.2.3 tomorrow so we can remove the hack

PR #289, opened 2017-07-10
Add gitter badge

jettify:


asvetlov:
Would you add such badge to other aio-libs projects?

jettify:
sure, do we want separate rooms on just main one?

asvetlov:
I think the main room for everything is good until we'll get it flooded by very many messages.

Issue #295, opened 2017-07-17
S3 `get_object` of key that has 'Content-Encoding' metadata set to `gzip` returns wrong data type


thehesiod:
I just ran into this, this is due to https://github.com/aio-libs/aiohttp/issues/1992

thehesiod:
I think this is now resolved

Issue #258, opened 2017-06-19
[Tracking] support compressed responses


thehesiod:
See https://github.com/aio-libs/aiohttp/issues/1992 and comment in endpoint.py

thehesiod:
this is now dependent on https://github.com/boto/botocore/issues/1255

thehesiod:
update: just need to write a unittest for this as it should be supported in 0.5.1

socketpair:
I think this issue should be closed.

PR #298, opened 2017-07-18
add support for compressed responses

thehesiod:
- resolves https://github.com/aio-libs/aiobotocore/issues/295 + https://github.com/aio-libs/aiobotocore/issues/258
- relies on https://github.com/aio-libs/aiohttp/pull/2110

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ from setuptools import setup, find_packages
 
 # aiohttp requirement is pegged as we need to manually ensure that
 # https://github.com/aio-libs/aiobotocore/pull/248 will continue working
-install_requires = ['botocore>=1.5.71, <=1.5.78', 'aiohttp>=2.0.4, <=2.3.0',
+install_requires = ['botocore>=1.5.71, <=1.5.78', 'aiohttp>=2.X.X, <=2.X.X',
                     'multidict>=2.1.4', 'wrapt>=1.10.10', 'packaging>=16.8']
 
 PY_VER = sys.version_info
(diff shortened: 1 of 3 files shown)


jettify:
`I just noticed that if I try a aiobotocore s3 client get_object of a key that a Content-Encoding metadata set to gzip, it's automatically decompressing the body as well, not-good...`

Good catch! Looks like `aiobotocore` hits every edge case of aiohttp client possible.

thehesiod:
crazy eh? :)  Also my service seems to have found a python segfault after updating a defaultdict a few million times <sigh>  that's not going to be fun to track down.

thehesiod:
oh, and in reality the issue with aiohttp is that it doesn't follow http stack norms :(   Basically designed like: I think it should behave like X by default, whereas every other library behaves like Y

jettify:
aio-libs/aiohttp#2110 had been merged, do you have idea how we can test this PR?

setup.py
```diff
@@ -6,7 +6,7 @@
 
 # aiohttp requirement is pegged as we need to manually ensure that
 # https://github.com/aio-libs/aiobotocore/pull/248 will continue working
-install_requires = ['botocore>=1.5.71, <=1.5.78', 'aiohttp>=2.0.4, <=2.3.0',
+install_requires = ['botocore>=1.5.71, <=1.5.78', 'aiohttp>=2.X.X, <=2.X.X',
```
jettify: updates her should fix CI

thehesiod:
it's been merged but there's no release available with it yet

setup.py
thehesiod: ya, broken on purpose for now so we set the right version that's needed

jettify: got it

jettify:
merged manually here https://github.com/aio-libs/aiobotocore/pull/425

PR #336, opened 2017-09-02
SQS Tests, Examples and some docs

terrycain:
The bulk of the PR is some examples and tests for SQS, but I also fixed a missing dependency and invalid rst file.

* Added tests for basic SQS operations along with accompanying test fixtures
* Created examples for basic SQS operations and a simple producer/consumer example

### DIFF
--- a/examples/sqs_queue_consumer.py
+++ b/examples/sqs_queue_consumer.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python3
+"""
+aiobotocore SQS Consumer Example
+"""
+
+# Boto should get credentials from ~/.aws/credentials or the environment
+import asyncio
+
+import aiobotocore
+
+QUEUE_NAME = 'test_queue1'
+
+
+async def go(loop):
+    session = aiobotocore.get_session(loop=loop)
+    async with session.create_client('sqs', region_name='us-west-2') as client:
+        response = await client.get_queue_url(QueueName=QUEUE_NAME)
+        assert response['ResponseMetadata']['HTTPStatusCode'] == 200
+        assert 'QueueUrl' in response
+        queue_url = response['QueueUrl']
+
+        print('Pulling messages off the queue')
+
+        while True:
+            try:
+                response = await client.receive_message(
+                    QueueUrl=queue_url,
+                    WaitTimeSeconds=2,
+                )
+                resp_status = response['ResponseMetadata']['HTTPStatusCode']
+                assert resp_status == 200
+
+                if 'Messages' in response:
+                    for msg in response['Messages']:
+                        print('Got msg "{0}"'.format(msg['Body']))
+                        # Need to remove msg from queue or else it'll reappear
+                        await client.delete_message(
+                            QueueUrl=queue_url,
+                            ReceiptHandle=msg['ReceiptHandle']
+                        )
+                else:
+                    print('No messages in queue')
+            except KeyboardInterrupt:
+                break
+
+        print('Finished')
+
+loop = asyncio.get_event_loop()
+loop.run_until_complete(go(loop))
--- a/examples/sqs_queue_consumer.py
+++ b/examples/sqs_queue_consumer.py
@@ -49,6 +49,7 @@ def go(loop):
     print('Finished')
     yield from client.close()
 
+
 try:
     loop = asyncio.get_event_loop()
     loop.run_until_complete(go(loop))
--- a/examples/sqs_queue_consumer.py
+++ b/examples/sqs_queue_consumer.py
@@ -2,22 +2,31 @@
 """
 aiobotocore SQS Consumer Example
 """
-
-# Boto should get credentials from ~/.aws/credentials or the environment
 import asyncio
+import sys
 
 import aiobotocore
+import botocore.exceptions
 
-QUEUE_NAME = 'test_queue1'
+QUEUE_NAME = 'test_queue12'
 
 
 @asyncio.coroutine
 def go(loop):
+    # Boto should get credentials from ~/.aws/credentials or the environment
     session = aiobotocore.get_session(loop=loop)
     client = session.create_client('sqs', region_name='us-west-2')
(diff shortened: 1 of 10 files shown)


terrycain:
Ahh flake8 on python3.4 is not liking the async/await in the examples, is this an issue? I can change it to the yield form if you want?

thehesiod:
ya for now we still support 3.4, @jettify when do you think we'll drop 3.4?

thehesiod:
btw thanks for the contributing!  Mind updating changes.txt?

thehesiod:
btw another option is creating 3.5+ only tests, we'd need to add some minor infrastructure for that though

requirements-dev.txt
```diff
@@ -9,4 +9,5 @@ aiohttp==2.2.5
 botocore==1.6.4
 multidict==3.1.3
 wrapt==1.10.11
-dill==0.2.7.1
\ No newline at end of file
+dill==0.2.7.1
```
thehesiod: I think we should fix this to follow this model: https://github.com/spulec/moto/blob/master/requirements-dev.txt  where it pulls from requirements.txt .. in this case instead will look more like: https://github.com/spulec/moto/blob/master/requirements-dev.txt

terrycain:
Yeah the tests were fine, the issue was py3.4 flake8 not liking syntax in the examples, that was an easy fix though

terrycain:
Also whats the opinion of a table in the readme showing what aws services are tested and known to be working?

thehesiod:
my opinion the more info the better :)  Last thing that was a surprise was the "waiters".

terrycain:
Yeah I saw a test for that, looks pretty neat

examples/sqs_queue_consumer.py
```diff
@@ -0,0 +1,57 @@
...
+
+QUEUE_NAME = 'test_queue1'
+
+
+@asyncio.coroutine
+def go(loop):
+    session = aiobotocore.get_session(loop=loop)
+    client = session.create_client('sqs', region_name='us-west-2')
+    response = yield from client.get_queue_url(QueueName=QUEUE_NAME)
+    assert response['ResponseMetadata']['HTTPStatusCode'] == 200
```
thehesiod: normally if the call fails an exception is raised, can get_queue_url return a failed response w/o raising?  Like for s3.get_object, a 404 is raised as an exception.

thehesiod: similar for the rest of the checks for 200

examples/sqs_queue_consumer.py
```diff
@@ -0,0 +1,57 @@
...
+            else:
+                print('No messages in queue')
+        except KeyboardInterrupt:
+            break
+
+    print('Finished')
+    yield from client.close()
+
+
+try:
```
thehesiod: since these are examples people may copy, probably worth doing here and below: 

```python
def main():
    try:
        loop = ...

if __name__ == '__main__':`
   main()
```

so they don't run it accidentally on import

terrycain: Good shout

examples/sqs_queue_consumer.py
terrycain: Will check, I just spammed the assertions everywhere :wink:

thehesiod:
cool, so I finished my review, we'll just wait a couple days to see if jettify responds about the requirements.txt, otherwise after other issues addressed can merge this and open an issue to discuss the requirements.txt thing.

terrycain:
Sounds good. :+1:

requirements-dev.txt
jettify: 👍

jettify:
Sorry for delay, I had a long flight, yesterday.

jettify:
Regarding examples, I think we need to convert all of them to async/await style to encourage new syntax. And once aiohttp move to 3.5 ( I think this winter) or python3.7 release, we safe to move too.

terrycain:
I was thinking of adding more examples for dynamo and others, would you rather I just use async/await and let the python3.4 flake to fail (or at least exclude examples from flake on py3.4)

Also whats your opinion on [this](https://github.com/aio-libs/aiobotocore/pull/336/files/e607c760bfac52233ff0495706a3d6ade34c4995#diff-f4c265073ea4274a80eca85440875e56) comment as I could throw that in too.

jettify:
@terrycain sorry for some reason I thought for something else, I am bit not follow what is problem https://github.com/aio-libs/aiobotocore/pull/336/files/e607c760bfac52233ff0495706a3d6ade34c4995#diff-f4c265073ea4274a80eca85440875e56 here?

requirements-dev.txt
jettify: @thehesiod could you please elaborate? I am bit confused what is problem here?

terrycain:
Oh theres not a problem, mainly @thehesiod wanted your opinion on rearranging how all the requirements.txt was handled. For example having the a main requirements.txt with `-e .` which would always use the latest requirements in setup.py and then requirements-dev.txt would install any dev/test dependencies and also have `-r requirements.txt`. 

The issue was (prior to merging this PR) if you just did `pip install -r requirements-dev.txt` as the readme suggested, you would be missing some dependencies in setup.py. The proposed way would mean there is technically less dependency files to update.

If your up for that I'll put those changes in my next PR

jettify:
Got it. Reason why we enumerate packages explicitly with version so pyup bot can track our dependencies and create PRs for new versions. This way it is easy to track breaking changes.  

I think we should add `-e .` to the end for the `requirements-dev.txt` so `aiobotocore` is installed after all pined dependencies.

terrycain:
Ahh I see. Thanks for clearing that up :smile:

jettify:
@terrycain thanks to you, for help with examples and tests!

terrycain:
No problem glad to help out, mainly it's for my own reference :wink:. If you have no objections, I'll happly throw in prs for dynamo and any other services I need to use or want to play with

jettify:
Sure, please do!

PR #339, opened 2017-09-05
better requirements-dev

jettify:


requirements-dev.txt
```diff
@@ -6,8 +6,9 @@ pytest-cov==2.5.1
...
 aiohttp==2.2.5
-botocore==1.6.4
+botocore==1.5.78
 multidict==3.1.3
 wrapt==1.10.11
 dill==0.2.7.1
-packaging==16.8
\ No newline at end of file
+packaging==16.8
+-e .
```
thehesiod: this is wrong, it needs to install `-e .` first, or else when pyup tries to force the latest botocore it will get downgraded and test the wrong version

jettify: got it, will fix

PR #340, opened 2017-09-05
DynamoDB Tests + Examples

terrycain:
Added more DynamoDB tests and added some simple DynamoDB examples

Also added `-e .` to requirements-dev.txt and a table to the readme.


terrycain:
Odd, Python3.5 failed on S3... even though those were not changed and 3.4 and 3.6 worked. And all 3 worked on my [fork](https://travis-ci.org/terrycain/aiobotocore/builds/272241426)

terrycain:
Ok changing literally an S to an N, now made tests fail on python 3.6, again my fork validated fine against travis

terrycain:
Ok, sorta makes sense. I guess if the travis boxes have a bit of load on them, it in theory could slow down the startup of moto. At least the tests work extending the timeout from 0.5 to 1.5

requirements-dev.txt
```diff
@@ -10,4 +10,5 @@ botocore==1.6.4
 multidict==3.1.3
 wrapt==1.10.11
 dill==0.2.7.1
-packaging==16.8
\ No newline at end of file
+packaging==16.8
+-e .
```
thehesiod: I think -e should be first, as I believe the last requirement should work.  In fact the tests should be failing because my patch test will fail with botocore==1.6.4.  I mentioned this in jettify's version as well.

tests/mock_server.py
```diff
@@ -29,7 +29,7 @@ def start_service(service_name, host, port):
             requests.get(url, timeout=0.5, proxies=_proxy_bypass)
             break
         except requests.exceptions.ConnectionError:
-            time.sleep(0.5)
```
thehesiod: instead of bumping this to 1.5s, could you instead bump the range at the top?  That way if it takes < 1.5s to start the service the tests will run faster

thehesiod:
thanks for all this work!

requirements-dev.txt
terrycain: I would agree about the tests but both tests ran with botocore at 1.5.78, and 1cc8450 only Py3.5 failed on S3 tests, and after triggering Travis again d72276d then Py3.6 failed on Dynamo and SQS. Which to me indicated somethings wrong with the starting of moto/the environment/the box, or timing tests, otherwise I would expect the failing tests to be the same, and the same version of python.

Yep you are correct, `-e .` should be first.

tests/mock_server.py
terrycain: Yeah good spot, that's a much better idea.

terrycain:
Updated range from 10 -> 30, this way the timing would be the same as if the delay was 1.5. Also moved `-e .`

thehesiod:
hmm, perhaps we should leave out the requirements changes and have @jettify 's PR do that because this now is missing the removal of `- pip install -e .` from travis.yml.

terrycain:
Yeah I literally just noticed that PR, if you can merge #339 I'll pull in those changes and all should be good.

tests/mock_server.py
jettify: 👍

jettify:
@terrycain thanks a lot!

terrycain:
No problem, for some reason am enjoying writing the async examples, am trying to think which next ones will be the most useful

PR #425, opened 2017-11-07
Update aiohttp

jettify:


setup.py
```diff
@@ -9,7 +9,7 @@
 # If adding requirements make sure to also add to requirements-dev.txt
 install_requires = [
     'botocore>=1.7.28, <=1.7.40',
-    'aiohttp>=2.0.4, <=2.3.0',
+    'aiohttp>=2.3.0, <=2.3.2',
```
asvetlov: Push `<3.0.0` here.
All aiohttp `2.3` bugfixes should be backward compatible.
Honestly I don't expect many of them, maybe one or two with very minor changes.

jettify: done

thehesiod:
@jettify I think we shouldn't have merged this until https://github.com/boto/botocore/issues/1255 is fixed.  I believe this is what is causing https://github.com/aio-libs/aiobotocore/issues/295

aiobotocore/endpoint.py
```diff
@@ -259,23 +219,11 @@ def __init__(self, host,
...
-        # 1. aiohttp is automatically decompressing the data
-        # (http://aiohttp.readthedocs.io/en/stable/client.html#binary-response-content)
-        # 2. botocore computes crc32 on the uncompressed data bytes and fails
-        # cause crc32 has been computed on the compressed data
-        # The following line forces aws not to use gzip compression,
-        # if there is a way to configure aiohttp not to perform decompression,
-        # we can remove the following line and take advantage of
-        # aws gzip compression.
-        # See: https://github.com/aio-libs/aiohttp/issues/1992
-        headers['Accept-Encoding'] = 'identity'
```
thehesiod: @jettify bug:  we need to keep this identity until https://github.com/boto/botocore/issues/1255 is fixed

thehesiod:
@asvetlov thanks for the redirect and compression fixes!

asvetlov:
@thehesiod you are welcome!

aiobotocore/endpoint.py
jettify: Got it, will bring back this lines.

PR #446, opened 2017-11-23
Upgrade moto manually

ticosax:
pyup-bot doens't upgrade it. is it on purpose ?


ticosax:
it was #442 my bad

jettify:
Thanks! They broke a lot of things in 1.0.0 but looks like most of things fixed now.

Issue #451, opened 2017-11-29
Timeout on S3 get_object using asyncio.gather


alliefitter:
I'm attempting to download 300 objects from S3 by building a list of futures and waiting on them to finish with `asyncio.gather`. For some reason, only the ten are finishing and the rest are timing out.

**Code:**
```python
import asyncio

import aiobotocore
from pyboto3 import s3 as s3_client


async def get_logs(s3: s3_client, keys: list = [], token: str = None):
    """

    :param token:
    :type token: str
    :param keys:
    :type keys: list
    :param s3:
    :type s3: s3_client
    :return:
    """
    r = await s3.list_objects_v2(
        Bucket='bucket',
        MaxKeys=1000,
        Prefix='prefix'
    )
    contents = r['Contents'] if 'Contents' in r else []
    is_truncated = r['IsTruncated'] if 'IsTruncated' in r else False
    token = r['NextContinuationToken'] if 'NextContinuationToken' in r else None
    keys = [c['Key'] for c in contents if 'Key' in c] + keys
    print('IsTruncated: {}'.format(is_truncated))
    return keys if not is_truncated else await get_logs(s3, keys, token)


async def get_object(s3: s3_client, key: str):
    """

    :param s3:
    :type s3: s3_client
    :param key:
    :type key: str
    :return:
    """
    print('Downloading {}'.format(key))
    r = await s3.get_object(Bucket='bucket', Key=key)
    print('Finished {}'.format(key))
    return r


async def run(s3: s3_client):
    """

    :param s3:
    :type s3: s3_client
    :return:
    """
    keys = await get_logs(s3)
    r = await asyncio.gather(*[get_object(s3, k) for k in keys])
    print(r)
    return s3.close()


_loop = asyncio.get_event_loop()
_s3 = aiobotocore.get_session(loop=_loop).create_client(
    's3',
    aws_access_key_id='id',
    aws_secret_access_key='key'
)
_loop.run_until_complete(run(_s3))
```
**Output:**
```
IsTruncated: False
Downloading mdm-architecture-reference/server/0f199d90-89fc-4fae-a8ab-f91e39ed3857/013d407166ec4fa56eb1e1f8cbe183b9/000000.gz
Downloading mdm-architecture-reference/server/0f199d90-89fc-4fae-a8ab-f91e39ed3857/077e29b11be80ab57e1a2ecabb7da330/000000.gz
[...]

PR #485, opened 2018-01-09
fix connector timeout

thehesiod:
this ensures the connector timeout only accounts for the time trying to connect to the endpoint  

underlying aiohttp issue: https://github.com/aio-libs/aiohttp/issues/2648
fixes aiobotocore issue: https://github.com/aio-libs/aiobotocore/issues/451


jettify:
Lets not rush patching aiohttp and give @asvetlov time to take a look on aiohttp issue.

jettify:
Depends on timing we can decide move forward with this patch or wait a bit.

jettify:
As for patch, looks good, not sure how to test it.

thehesiod:
based on previous conversations regarding timeouts I doubt they'll fix it, but ya we can wait and see.   ya let me think about a test.

thehesiod:
@jettify ok added test, took awhile to get right as we needed a custom endpoint

tests/mock_server.py
```diff
@@ -1,27 +1,110 @@
...
+        self._shutdown_evt = asyncio.Event()
+
+    def _run(self):
+        self._loop = asyncio.new_event_loop()
+        app = aiohttp.web.Application(loop=self._loop)
+        app.router.add_route('*', '/ok', self.ok)
+        app.router.add_route('*', '/{anything:.*}', self.stream_handler)
+
+        try:
+            aiohttp.web.run_app(app, host='localhost', port=self._port,
```
jettify: you can run aiohttp server in same thread without using threading. See example here https://github.com/aio-libs/aiozipkin/blob/5f914c6eb811db046a58d5d816cd19223ae2435c/demos/queue/runner.py#L7-L25 it is a bit broken I wrote those late night, but you get the idea

tests/mock_server.py
```diff
@@ -56,8 +139,11 @@ def s3_server():
     port = 5000
     url = "http://{host}:{port}".format(host=host, port=port)
     process = start_service('s3', host, port)
-    yield url
-    stop_process(process)
+
+    try:
+        yield url
+    finally:
```
jettify: 👍

tests/mock_server.py
thehesiod: cool, want me migrate?

jettify: yes please.

jettify:
I think timeouts will not be fixed in next major release, so we can move forward with your patch.

tests/mock_server.py
thehesiod: i just realized I shutdown the server via `loop.stop` however we cannot do this in a unittest so I think we need to keep it in a thread.

thehesiod:
getting this run on travis on 3.4 is not fun

thehesiod:
coverage looks good, for some reason didn't update in this PR

PR #499, opened 2018-01-24
remove upper bound on botocore version

supergil82:


thehesiod:
can't do this as it's not guaranteed that it will work

PR #523, opened 2018-02-20
Waiters fix

dalazx:
Apparently waiters had been completely broken.
```python
class NormalizedOperationMethod(object):
    def __init__(self, client_method):
        self._client_method = client_method

    def __call__(self, **kwargs):
        try:
            return self._client_method(**kwargs)
        except ClientError as e:
            return e.response
```

All `ClientError`s here were slipping outside, because ` return self._client_method(**kwargs)` would always return a coroutine object in case of our async client.


dalazx:
need to pass a loop into `wait`.

dalazx:
@jettify please take a look :)

thehesiod:
nice catch!

thehesiod:
@dalazx want to add a thanks for yourself in the changes?  @jettify perhaps we should have a contributors file like aiohttp?

aiobotocore/waiter.py
```diff
@@ -1,14 +1,30 @@
...
+from botocore.exceptions import ClientError
 from botocore.waiter import WaiterModel  # noqa: F401
-from botocore.waiter import Waiter, xform_name, NormalizedOperationMethod, \
-    logger, WaiterError
+from botocore.waiter import Waiter, xform_name, logger, WaiterError
 from botocore.docs.docstring import WaiterDocstring
 from botocore.utils import get_service_module_name
 
 
+class NormalizedOperationMethod(object):
```
thehesiod: don't need to inherit from object in 3+

thehesiod: given that we're now replacing `NormalizedOperationMethod`, we need to add an entry in https://github.com/aio-libs/aiobotocore/blob/master/tests/test_patches.py to ensure that if it changes our test breaks and someone reviews that our replacement is still correct.

thehesiod: btw I'm not sure if completely broken is correct, the coroutine would be returned but IIRC the caller was doing a `yield from` so the method would still execute.  That support was added after someone noticed waiters were broken and I think he said it worked after the change...so either botocore changed since we implemented this, or the real issue is that the exception handler was not being applied correctly.

thehesiod:
btw I'm not sure if completely broken is correct, the coroutine would be returned but IIRC the caller was doing a yield from so the method would still execute. That support was added after someone noticed waiters were broken and I think he said it worked after the change...so either botocore changed since we implemented this, or the real issue is that the exception handler was not being applied correctly.

dalazx:
@thehesiod the waiters were working properly only in cases when responses had status codes < 300, and this limitation completely neglects the purpose. due to the fact that a coroutine object was being returned immediately, the `__call__` method call thus was removed from the stack and `ClientErrors` were simply slipping up the stack.

aiobotocore/waiter.py
dalazx: now the botocore version of `NormalizedOperationMethod` and the aiobotocore one are unrelated. our changes are localized and the only place where our `NormalizedOperationMethod` version gets instantiated is `aiobotocore/waiter.py`.

dalazx: added the digest as requested.

dalazx: removed.

aiobotocore/waiter.py
```diff
@@ -1,14 +1,30 @@
...
 
 
+class NormalizedOperationMethod:
+    def __init__(self, client_method):
+        self._client_method = client_method
+
+    @asyncio.coroutine
+    def __call__(self, **kwargs):
+        try:
+            return (yield from self._client_method(**kwargs))
```
dalazx: actually we could've had a guard to handle non-coroutine cases as well if makes sense.
thoughts @thehesiod @jettify ?

thehesiod: When would there be a non coroutine case? To my understanding this triggers an aiohttp call, thus coroutine

thehesiod:
Cool, i suppose the person who requested the feature didn't really use it or had a very specific use case. I'm going to merge as I did the original waiters fix and this looks good to me. If jettify finds any issues we can address them before releasing the next version. How soon do you need an official release?

thehesiod:
I'll wait on discussion above, going to sleep :)

aiobotocore/waiter.py
dalazx: thought so too. have not read the whole thing to be 100% sure there are only coroutine cases. still can easily be fixed later.

dalazx:
@thehesiod thanks!
regarding the release, there is no hurry. the issue was mitigated by polling `dynamodb` `describe_table` instead. :)

jettify:
@dalazx @thehesiod thanks a lot for handling this! Sorry for delayed responses I have been on business trip for past few weeks.

thehesiod:
@dalazx FYI 0.5.3 released with your fix

dalazx:
@thehesiod awesome! thanks

Issue #519, opened 2018-02-15
Adopt aiohttp 3


asvetlov:
Currently aiobotocore pins aiohttp dependency to 2.x, that's sad.

thehesiod:
:) we wouldn't have to pin if we had the timeout changes we need *g*

asvetlov:
Looks like you are bargaining

thehesiod:
no just fact of life, I'm sure we'll get the update done soon, just need to make sure our monkey patches continue working, would be better if we didn't need to monkey patch.

thehesiod:
note for who updates this, it means we need to drop py<3.5.3 (yay), which means major at least version 0.6

Gr1N:
Maybe you have any estimates? We want to go with `aiohttp>=3.0.0`, but we can't due to `aiobotocore`...(

thehesiod:
If jettify doesn't get to it this week I'll try, feel free to open a PR as well :)   Btw I noticed aiohttp 3 is going through some churn, which is normal for a major release bump, so probably worth waiting a bit anyways.

Gr1N:
Just tried to look into issues and here is a patch (all tests passed):
```
diff --git a/aiobotocore/endpoint.py b/aiobotocore/endpoint.py
index 601b7e6..e1eb768 100644
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -181,11 +181,11 @@ class WrappedTCPConnector(aiohttp.TCPConnector):
         super().__init__(*args, **kwargs)
 
     @asyncio.coroutine
-    def _create_connection(self, req):
+    def _create_connection(self, req, traces=None):
         # connection timeout
         try:
             with CeilTimeout(self.__wrapped_conn_timeout, loop=self._loop):
-                return super()._create_connection(req)
+                return super()._create_connection(req, traces=None)
         except asyncio.TimeoutError as exc:
             raise aiohttp.ServerTimeoutError(
                 'Connection timeout '
diff --git a/requirements-dev.txt b/requirements-dev.txt
index 3a2ba92..96e9f94 100644
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -7,8 +7,8 @@ moto==1.2.0
 pytest-cov==2.5.1
 pytest==3.3.1
 sphinx==1.6.5
-yarl<1.0.0
-aiohttp==2.3.3
+# yarl<1.0.0
+# aiohttp==2.3.3
[...]

PR #536, opened 2018-02-27
Adopt aiohttp>=3.0.0 (#519)

Gr1N:
Dropped support of Python 3.4 and aiohttp < 3.0.0.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ from setuptools import setup, find_packages
 # If adding requirements make sure to also add to requirements-dev.txt
 install_requires = [
     'botocore>=1.8.0, <=1.8.21',
-    'aiohttp>=2.3.0, <3.0.0',
+    'aiohttp>=3.0.0',
     'multidict>=2.1.4',
     'wrapt>=1.10.10',
     'packaging>=16.8',
@@ -17,8 +17,8 @@ install_requires = [
 
 PY_VER = sys.version_info
 
-if not PY_VER >= (3, 4, 1):
-    raise RuntimeError("aiobotocore doesn't support Python earlier than 3.4")
+if not PY_VER >= (3, 5, 3):
+    raise RuntimeError("aiobotocore doesn't support Python earlier than 3.5")
 
 
 def read(f):
@@ -49,7 +49,6 @@ classifiers = [
     'Intended Audience :: Developers',
     'Intended Audience :: System Administrators',
     'Programming Language :: Python :: 3',
-    'Programming Language :: Python :: 3.4',
     'Programming Language :: Python :: 3.5',
     'Programming Language :: Python :: 3.6',
     'Environment :: Web Environment',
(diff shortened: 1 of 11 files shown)


setup.py
```diff
@@ -9,16 +9,16 @@
...
     'multidict>=2.1.4',
     'wrapt>=1.10.10',
     'packaging>=16.8',
 ]
 
 PY_VER = sys.version_info
 
-if not PY_VER >= (3, 4, 1):
-    raise RuntimeError("aiobotocore doesn't support Python earlier than 3.4")
+if not PY_VER >= (3, 5, 0):
```
asvetlov: 3.5.3 actually. The same in all other places

Gr1N: fixed, thanks

tests/test_patches.py
```diff
@@ -28,19 +28,19 @@
...
 
     # for our replacement of read()
-    DataQueue: {'8ad4d5df1d016547daea6389707bc656630582e5'},
+    DataQueue: {'be516f7bcdbf5235218261d8ed1f490d299f611d'},
 
     # for our patch of _wait
-    StreamReader: {'c0a9a31a8c3e550de5985ab642028983f709b37b'},
+    StreamReader: {'d4ffb6ae823ef4bfd810aade8601ba7b01aa08ec'},
 
     # for digging into _protocol ( 2.1.x, 2.2.x )
```
thehesiod: can remove 2.1.x, 2.2.x

thehesiod:
cool looks good!  I validated the changes to aiobotocore and looks like there's no impact to us, yay.  just have one comment to remove a comment :)  Let's see what @jettify says too.

thehesiod:
would be good to bump to the latest botocore too, but we can do that in another PR

tests/test_patches.py
Gr1N: removed

tests/mock_server.py
```diff
@@ -45,8 +46,12 @@ def _run(self):
         app.router.add_route('*', '/{anything:.*}', self.stream_handler)
 
         try:
-            aiohttp.web.run_app(app, host=host, port=self._port,
-                                loop=self._loop, handle_signals=False)
+            # We need to mock `.get_event_loop` function and return
+            # `self._loop` explicitly because from `aiohttp>=3.0.0` we can't
+            # pass `loop` as a kwargs into `run_app`.
+            with mock.patch('asyncio.get_event_loop', return_value=self._loop):
+                aiohttp.web.run_app(app, host=host, port=self._port,
```
jettify: We can use new AppRunner stuff as result we do not need ugly patch
https://docs.aiohttp.org/en/stable/web_advanced.html?highlight=AppRunner#application-runners

jettify:
Changes looks good to me as well, one nitpick. We can merge PR as it is and handle runner stuff in separate PR

Gr1N:
@jettify maybe it's bad, but I don't want to spend time on `AppRunner`. If it possible please merge this PR and migrate to `AppRunner` in separate PR.

jettify:
thanks!

PR #564, opened 2018-04-28
fix non-streaming body retries

thehesiod:
see: https://github.com/aio-libs/aiobotocore/pull/564/files#diff-e941ba728e64131911465db18f803157R268

I've gone ahead and switched us away from `asyncio.coroutine` as well

this has been split to the above fix plus:
- https://github.com/aio-libs/aiobotocore/pull/567
- https://github.com/aio-libs/aiobotocore/pull/566

will merge this after other two so this will collapse to just the fix

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,10 @@
 Changes
 -------
+0.6.1 (2018-04-XX)
+^^^^^^^^^^^^^^^^^^
+* fix timeout retries for non streaming body responses
+* fix content property of response
+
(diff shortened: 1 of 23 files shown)


thehesiod:
hmm, I saw this sns test fail once on my machine, and then pass when I ran it again

thehesiod:
closing in favor of:
- (retries) https://github.com/aio-libs/aiobotocore/pull/568
- (bump to aiohttp 3.1.x) https://github.com/aio-libs/aiobotocore/pull/569

PR #566, opened 2018-05-01
switch tests to py35+

thehesiod:


tests/conftest.py
```diff
@@ -118,7 +62,7 @@ def assert_num_uploads_found(s3_client, bucket_name, operation,
             return
         else:
             # Sleep and try again.
-            yield from asyncio.sleep(2, loop=loop)
+            await asyncio.sleep(2, loop=event_loop)
```
jettify: Not big problem I usually create helper fixture in order not to do bunch of renamings  
```
@pytest.fixture
def loop(event_loop)
    return event_loop
```

thehesiod: ya, but by doing that in the future `pytest-asyncio` users may get temporarily confused and have to figure out what's happening.  However if `pytest-asyncio` users see `event_loop` they'll realize what's going on.

PR #567, opened 2018-05-01
switch code to native coroutines

thehesiod:


thehesiod:
I can't get more code coverage until we land https://github.com/aio-libs/aiobotocore/pull/566 to use `async for`

PR #568, opened 2018-05-01
fix non-streaming body timeout retries

thehesiod:
brings parity with botocore so that timeouts while requesting the body of non-streaming requests will retry


thehesiod:
lets merge https://github.com/aio-libs/aiobotocore/pull/567 first, then update this PR

jettify:
Change looks good, looks like we hit that line in some of our tests?

thehesiod:
ya, any non-streaming response (list-objects, head-object, etc) will hit that line

PR #587, opened 2018-06-01
bump to latest aiohttp

thehesiod:
enables removal of a bunch of monkeypatches for timeout behavior


thehesiod:
FYI @jettify

jettify:
nice!

thehesiod:
@jettify mind checking why PyPI upload is failing?  seems to be something related to the password

PR #608, opened 2018-07-06
Update error message for Python 3.5.3 requirement

### LINKED ISSUE #591: Update error message for Python 3.5.3 requirement
setup.py now requires Python 3.5.3, but systems running earlier 3.5.x versions (in my case 3.5.2 on Ubuntu 16.04) get an error message of `aiobotocore doesn't support Python earlier than 3.5`. A more precise log message would be helpful.

qria:
Fixes #591

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ install_requires = [
 PY_VER = sys.version_info
 
 if not PY_VER >= (3, 5, 3):
-    raise RuntimeError("aiobotocore doesn't support Python earlier than 3.5")
+    raise RuntimeError("aiobotocore doesn't support Python earlier than 3.5.3")
 
 
 def read(f):


thehesiod:
@jettify @asvetlov I think the travis-ci config is wrong, seems like it's going to .org instead of .com, see https://github.com/aio-libs/aiobotocore/settings/installations (I can't click configure), but if you click "Edit" on the button at the bottom of the page it shows the domain to be `notify.travis-ci.org`.  I'm guessing that's why travis isn't getting notified of the branch and triggering a build.  I was reading something about this year projects moving from .org to .com.

jettify:
Integration on org was enabled, I disabled it. But PR still stuck for some reason, not sure how to disable  that required check

PR #637, opened 2018-10-15
Edit code (examples/simple.py)

SeungUkLee:
used `async with` in `README.rst`, but did not use `async with` in the `examples/simple.py` file. so i edited `examples/simple.py` code.

Signed-off-by: SeungukLee <lsy931106@gmail.com>


thehesiod:
ty!

PR #638, opened 2018-10-15
Support botocore 1.12.20

suburbanmtman:


thehesiod:
did you need 1.12.20?

PR #639, opened 2018-10-16
Update to botocore 1.12.49

terrycain:
Botocore hasn't changed much between 1.10.58 to 1.12.24. The main ones are the changes to the event names emitted, and that botocore now use a urllib session class, so some arguments are no longer supported in the endpoint base class.

There's some weird travis bug at the moment (https://github.com/travis-ci/travis-ci/issues/7940), added in `BOTO_CONFIG` env var to solve that until its fixed properly.

PR #648, opened 2018-11-20
PEP8 compliance for examples/simple.py

superduper:
Hey! This PR fixes small PEP8 issues that cause build fail.


jettify:
Looks like fixed as part of #639 Thanks anyway!

Issue #619, opened 2018-08-02
[infra] async'ify Credentials class


thehesiod:
There are at least a couple underlying issues related to this class:

1. IAM credential refresh using requests: https://github.com/aio-libs/aiobotocore/issues/2
2. AssumeRoleProvider uses client call: https://github.com/aio-libs/aiobotocore/issues/553 + https://github.com/aio-libs/aiobotocore/issues/663

First one results in possible hangs of the main thread, while second one is broken because it's not await'ing the request of the client request.

The issue is that these are low-level in classes used by the Credential helper, and ultimately triggered by high level properties like `secret_key`, `access_key` and `token`.  We need to find all the calls that will end up needing to be changed to async, and then figure a strategy of how to battle this.  I'm afraid this is going to touch a LOT of botocore and greatly increase our footprint and amount of maintenance required.

mjmdavis:
Is there a workaround for this at the moment?

mannickutd:
@mjmdavis the work around I used was to use a synchronous botocore call using the refresh method defined in https://www.pydoc.io/pypi/botocore-1.7.0/autoapi/credentials/index.html#credentials.RefreshableCredentials and update credentials in the aiosession for the async clients.

sbrandtb:
@mannickutd Could you elaborate and maybe provide a code snippet?

thehesiod:
prefixing with infra as my experiment yielded a LOT of changes

thehesiod:
btw my PR needs to be re-done to be cleaner by delaying client creation to the `__aenter__` so we don't have an extra await.

thehesiod:
just got permission to expose what our company has been using to resolve this issue internally: https://gist.github.com/thehesiod/05298c1c89c7b5a38da0abc4ccbed7b7

Issue #658, opened 2019-01-04
Is it possible to upload to s3 asynchronously by chunks?


py-radicz:
What I am trying to do: I want to download file from remote source and upload it directly to s3 without storing the bytes on the localhost. (Stream the data from remote machine through my machine to s3). There are tens of files in remote location, thats why I want to use async lib, to stream all files simultaneously to save time.

Is it possible with current version of lib to do something like following?

```python
async with aioboto3.client('s3', loop=loop) as client:
    async for chunk in response.content.iter_chunked(1024):
        if chunk:
            # await put chunk into appropriate file handle on s3
```

Thanks for some pointers.

thehesiod:
I'm interested in your findings, let me know if you need some pointers

thehesiod:
ok, working on this now, will require some support

thehesiod:
omg, this is another painful thing.  So first thing needed was adding:
```python
    def tell(self):
        return self._self_amount_read
```
to `StreamingBody`.

Next issue is that the falling callstack needs to be async:
```
    await s3_client.put_object(Bucket='thehesiod-temp', Key='latitude.csv', Body=get_response["Body"])
  File "aiobotocore/aiobotocore/client.py", line 86, in _make_api_call
    request_signer=self._request_signer, context=request_context)
  File "botocore/hooks.py", line 360, in emit_until_response
    return self._emitter.emit_until_response(aliased_event_name, **kwargs)
  File "botocore/hooks.py", line 243, in emit_until_response
    responses = self._emit(event_name, kwargs, stop_on_response=True)
  File "botocore/hooks.py", line 211, in _emit
    response = handler(**kwargs)
  File "botocore/handlers.py", line 211, in conditionally_calculate_md5
    calculate_md5(params, **kwargs)
  File "botocore/handlers.py", line 189, in calculate_md5
    binary_md5 = _calculate_md5_from_file(body)
```

This is similar in complexity to https://github.com/aio-libs/aiobotocore/issues/619

thehesiod:
ugh, I found that if you do something like:
```python
    boto_session = aiobotocore.session.AioSession()

    async with (await boto_session.create_client('s3')) as s3_client:
        get_response = await s3_client.get_object(Bucket=BUCKET, Key='KEY)

        await s3_client.put_object(Bucket=BUCKET, Key=KEY, Body=get_response["Body"])
```

botocore reads the whole file to calculate an MD5 of the file, so you won't save anything in terms of memory usage.

[...]

PR #659, opened 2019-01-05
async signing support

thehesiod:
This starts the large work of supporting https://github.com/aio-libs/aiobotocore/issues/619

Large changes:
- (api breaking) the result of`create_client` is now a required async context class
- `generate_presigned_url` is now an async call
- `Credentials.[access_key/secret_key/token]` now raise NotImplementedError because they won't call refresh like botocore.  Instead should use get_frozen_credentials async method

In general, this is going to increase our footprint significantly.

TODO
------
- [x] get tests/checks passing
- [x] ensured that all the `events.emit` calls have been properly wrapped (all the way up call chain)
- [x] add patches for new methods
- [x] port credential classes and callers, partial list: IMDSFetcher, ContainerMetadataFetcher, AssumeRoleWithWebIdentityCredentialFetcher, ProcessProvider, AssumeRoleCredentialFetcher
- [x] add tests for coverage (ports from botocore)
- [x] look into `botocore.utils.S3RegionRedirector.redirect_from_error`
- [x] adhoc test using real aws creds


thehesiod:
with the changes as-is we support the simple testcase I list in https://github.com/aio-libs/aiobotocore/issues/658

terrycain:
This'll help with #707 as we essentially want to make botocore.utils.S3RegionRedirector.redirect_from_error async

thehesiod:
I'm going to try working on this this weekend

terrycain:
@thehesiod hows it going?

thehesiod:
@terrycain under a mountain of work with new baby almost here 🤯 :)

terrycain:
Np man, congrats :smile:

stalkerg:
I am very interesting in async generate_presigned_url, how I can help to merge this PR?

thehesiod:
hello @stalkerg , you can take a  look at this PR.  We need to change `create_client` from being an async method to returning an async context class as mentioned in the description.  Also need to merge from master, and ensure all the emit methods that touch these changes are async.  In general need to do a lot of research to ensure this system will work, which means we probably need a lot of more testing.  This PR probably should be the place we do adaptation of the underlying botocore unittests.

thehesiod:
not sure how this happened, but we're out of sync with botocore, I'm working on cleaning this up

thehesiod:
omg, the whole credentials file needs to be ported!  I give up for awhile again, so much work!

terrycain:
So I had a quick look at this just now - https://github.com/terrycain/aiobotocore/tree/streaming
Managed to get generate_presigned_url working... but as all the event emitting is async and the generate_presigned_url in the request signer is async, needed to copy the function and do some awaits, then replace the function that adds generate_presigned_url to the classes in the component tree thing.

Had a look into the credentials class, my god that file is nasty, literally everything needs to be patched as well as some classes from utils.py.

thehesiod:
@terrycain welcome to hell lol

terrycain:
You mind moving your branch to the aiobotocore repo, then I could add to it, if you don't mind that is :D

thehesiod:
@terrycain the branch is here, it's just called `thehesiod/streaming`.  I don't have a fork of this repo

terrycain:
Duh, i even grabbed your changes from it earlier, my brain is not working today. Yeah I'll merge in my fixes to signatures, see where we're at and possibly start on that horrific credentials file

terrycain:
Ok so from my basic test of s3.list_buckets, it seems to be using the asyncified credentials and signatures. Needs many many many tests but we're getting there.

thehesiod:
i say when this and the 3.8 PR land we actually bump to 1.0, what do you think?

terrycain:
And the aiomoto PR, by then when we hit 1.0 we'll have a pretty decent test suite with hopefully a good amount of coverage.

thehesiod:
merged in my py38 changes

aiobotocore/session.py
```diff
@@ -1,25 +1,80 @@
...
-from .client import AioClientCreator
+from .client import AioClientCreator, AioBaseClient
+from .hooks import AioHierarchicalEmitter, AioEventAliaser
 from .parsers import AioResponseParserFactory
+from .signers import add_generate_presigned_url
+from .credentials import create_credential_resolver, AioCredentials
 
 
-class AioSession(botocore.session.Session):
+class ClientCreatorContext:
```
thehesiod: I'm sure this is available from python somewhere

thehesiod:
btw for testing we should try to port the botocore tests as much as possible.  We can probably simplify to just one test per area we're replacing since we're building on top of a tested Botocore.

terrycain:
Yeah, I've had a quick look over the credentials tests mainly the parts where we've changed the requests out for aiohttp. For the tests, they're all unittests, would it be worth grabbing a load and converting to pytest if its simple?

thehesiod:
Botocore uses nose I think, perhaps we could have a separate set of tests that are translated botocore tests. I was wondering if we run the tests as similar as possible where the test runner keeps on running the test while it returns an awaitable perhaps we could almost run the tests unchanged

terrycain:
So I had a look over their credentials tests, for the parts we've patched I think we can get away with stealing one or two tests per class. Not finished on the credentials tests yet but will see where the coverage leaves us

terrycain:
We're changing generate_presign_url to async. We'll need to bump to a major version, but in the meantime would it make sense to add a warning to it? So people are aware when they update

terrycain:
:) coverage is getting there

terrycain:
So the S3RegionRedirector works, there is an edge case where it would not, but I cant see a case where S3 won't return the redirected region header in the redirect response so am not patching the rest of that for now. Signer now respects changes in regions when signing retried requests.

thehesiod:
> We're changing generate_presign_url to async. We'll need to bump to a major version, but in the meantime would it make sense to add a warning to it? So people are aware when they update

Eh. Since it's a major change in don't think we need to. Maybe in front page we can add upcoming changes to expect

thehesiod:
Sorry for not helping recently. Have a major project we're trying to got through at work

terrycain:
Ok so I think the test coverage is better than it was pre-pr :wink: I'm going to test the code against real creds including the assume role variants to see if they actually work for real.

@thehesiod is there anything glaringly obvious were missing apart from spamming patch hashes for stuff we've overridden?

Then we'll merge in any changes from master (which will probably mean updating test patches again, but worthwhile) and then I think we're good

terrycain:
Also now supports the new v2 instance metadata endpoint which requires you doing a PUT request to get a session token of sorts (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html)

terrycain:
Merged in master, no changes required.

Tested with the following env vars
```
AWS_ROLE_ARN=arn:aws:iam::100000000000:role/test
AWS_WEB_IDENTITY_TOKEN_FILE=/tmp/jwt
AWS_ROLE_SESSION_NAME=session1
```
And script:
```python
import aiobotocore
import asyncio
async def main():
    session = aiobotocore.session.AioSession()
    async with session.create_client('s3') as client:
        print(await client.get_object(Bucket='terry', Key='test'))
asyncio.get_event_loop().run_until_complete(main())
```
It complained because my JWT file was not a JWT but it seemed to call assume role with web identity fine, and that's probably the most complicated path through the credentials mess. I believe it'll work with the AWS EKS Kube IAM role integration stuff (https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/).

@thehesiod when you get some time, you mind reviewing this and checking the emitter stuff (think we're all there)

The hell is coming to an end :D

thehesiod:
Hey Terry I'll get to this asap, having a hard time at home with the two kids and work project. Really appreciate all the help!

terrycain:
Npnp when you can :smile:

tests/test_patches.py
```diff
@@ -110,15 +217,56 @@
...
     Waiter.wait: {'5502a89ed740fb5d6238a6f72a3a08efc1a9f43b'},
     create_waiter_with_client: {'c3d12c9a4293105cc8c2ecfc7e69a2152ad564de'},
 }
 
+# REPLACE = backwards incompatible change
+# APPEND = officially supporting more versions of botocore/aiohttp
+
+# If you're changing these, most likely need to update setup.py as well.
+
 _PROTOCOL_PARSER_CONTENT = {'ec2', 'query', 'json', 'rest-json', 'rest-xml'}
```
thehesiod: @terrycain wow =)

terrycain: Yeah... the rabbit hole ran deep on this one. Mainly the credentials class and the utils that it used.

thehesiod:
ok tonight I'll finish reviewing this because i need it for a project :)

aiobotocore/credentials.py
```diff
@@ -0,0 +1,774 @@
...
+            obj._access_key, obj._secret_key,
+            obj._token, obj._expiry_time,
+            obj._refresh_using, obj.method,
+            obj._time_fetcher
+        )
+
+    # Redeclaring the properties so it doesnt call refresh
+    # Have to redeclare setter as we're overriding the getter
+    @property
+    def access_key(self):
```
thehesiod: @terrycain this is a problem no?  Look at `HmacV1PostAuth.add_auth` : `fields['AWSAccessKeyId'] = self.credentials.access_key`

and `HmacV1QueryAuth._inject_signature` : `query_dict['AWSAccessKeyId'] = self.credentials.access_key`. there are a few more places too

thehesiod: this is kinda the lynchpin :)

thehesiod:
@terrycain this is definitely hero status work :)

thehesiod:
@terrycain ok finished, everything looks great except for the access_key etc properties, I made them raise for now because they won't work since they're not calling refresh.  We need to figure out what to do with these and then we can merge!

thehesiod:
btw I re-organized the code so that it would match the order in botocore to make diffs easier

thehesiod:
ok, I think we're good in terms of emitters, now to fix tests again

thehesiod:
ok fixed tests, and set `access_key`+ `secret_key` + `token` properties to raise NotImplementedError, instead callers should use `get_frozen_credentials`. thoughts?

terrycain:
Ok so according to something I read in the botocore source/docs somewhere, everything is supposed to call get_frozen_credentials, lemme have a look

aiobotocore/credentials.py
terrycain: Yeah true

terrycain: Though technically all signing auth classes are passed frozen credentials :D but yeah, the only safe place to read access_key... would be from the ReadOnlyCredential class retrieved from get_frozen_credentials

terrycain:
Yeah I think were good. So tests are failing because of an old pyyaml version having a CVE, pretty sure aws-cli is the culprit as they seem to have moved to using ruamel.yaml. 

Should we just do the usual bump botocore version stuff here and then merge?

terrycain:
Also tests run fine for me so the access_key changes are good

thehesiod:
ya lets bump and merge!  We're hitting this CVE at work too so bump would be awesome, if you don't get to in in a few hours I'll do it and merge. 🎆 🍾

thehesiod:
ok doing it now

terrycain:
Yay :smile:

thehesiod:
@terrycain thanks so much for your help!  Wouldn't have been possible without you, at least for a long time :]

terrycain:
No problem ;-). Just realized we need to override session._set_credentials() as otherwise it breaks when access and secret key are passed explicitly :(. Will pr in morning if you don't get there first, am too tired now :D

thehesiod:
Hmm I left the set, just not the get. Too tired to look now too though :)

PR #678, opened 2019-02-08
document why `__iter__` is not implemented

graingert:
### Description of Change

### Assumptions


thehesiod:
note for future reference, this will look like this: ``
NotImplementedError: <aiobotocore.paginate.AioPageIterator object at 0x107ba45c0> is an AsyncIterable: use `async for` ``

PR #711, opened 2019-07-18
bump botocore req to avoid pyaml CVE

thehesiod:


kestein:
awscli==1.16.101 requires pyyaml 3.3.13, which [has a vuln](https://nvd.nist.gov/vuln/detail/CVE-2017-18342)

PR #715, opened 2019-07-29
add support for pinning aws-xray version

oba11:
### Description of Change
Pin [aws-xray-sdk](https://github.com/aws/aws-xray-sdk-python) to aiobotocore to avoid incompatibility with botocore

* [x] I have added change info to [CHANGES.txt](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.txt)


terrycain:
:+1:

oba11:
thank you @terrycain for the very fast review

thehesiod:
what issue did this solve, and why is it hard locked?

thehesiod:
We can't have changes described as `Pin aws-xray-sdk to aiobotocore to avoid incompatibility with botocore` because it gives no context as to what the issue was, which means someone can come along and remove this change because we have no idea to test if it's needed.  I've updated our pull request template to ensure we get this information in the future. 

I think `aws-xray-sdk` just needed to be a `>=` some base version with a comment in setup.py for why we're locking that module, given it doesn't follow the reasoning for why boto3 + awscli are provided in extras.

If I don't hear back from anyone w/in a week or so I'm going to change it to be  a `>=` check in https://github.com/aio-libs/aiobotocore/pull/736

oba11:
@thehesiod you are right, hard-lock might be too much and its fine to change it from `==` to `>=` as long as the [pinned botocore](https://github.com/aws/aws-xray-sdk-python/blob/master/setup.py) in aws-xray-sdk doesn't change.
Thank you

thehesiod:
where I work we've been switching to pipenv to resolve install dependencies.   I'm kinda torn now with these extras, we need to define exactly what should go here, and how versions of modules in this section should be locked.  Really this is all a problem with how pip doesn't enforce version requirements correctly :(  Even pipenv has bugs where it sometimes won't figure out a solution when one exists.

I'm not this change is needed anymore given the botocore version we currently have.  I'll do some tests to see the current behavior.  I'm afraid of my proposed change given it lists botocore as >= which means if I set it in aiobotocore also as >=, as soon as xray requires a newer botocore version than aiobotocore I think it will break again.  We probably need a table to list what  happens with the various scenarios.

thehesiod:
I've done some tests installing aws-xray-sdk + aiobotocore in a fresh environment and couldn't find a way that breaks requirements.  The extras are meant to help find libraries that are version locked to botocore, pre tools like pipenv.  If aws-xray-sdk followed botocore versioning it would make more sense because it would be hard for the user to figure out what version of aws-xray-sdk to use.  Given the current version of aiobotocore works with the very first version of aws-xray-sdk that requires botocore I see so need for this for now. Since this PR didn't specify how to reproduce the issue, how this fixes the issue, and why it's needed I'm going to revert the change from master for now.  If you can come up with the required info please re-open issue and/or PR so we can debate it again.  Sorry for not catching this earlier, was on paternity leave...but at least we don't have an aiobotore release with this for now, so if it's really needed we can still add it w/o too much confusion.

thehesiod:
@oba11 your issue is probably you were installing aws-xray-sdk with `pip install -U`, that's not going to work as pip is dumb as it will only look at the reqs of what you're installing, and not everything you already have.  You should install both aiobotocore + aws-xray-sdk in an atomic operation, like via a requirements.txt file

PR #724, opened 2019-09-01
Closing aiohttp session should be awaited

bsitruk:
Looks like aiohttp close method is a coroutine.
I guess we can make this change now.

### Description of Change

### Assumptions


thehesiod:
I think this is not needed, the existing method will return a coroutine, which can be awaited

thehesiod:
actually this is fine

thehesiod:
btw I'm adding your missing changes.txt entry in https://github.com/aio-libs/aiobotocore/pull/736

PR #739, opened 2019-10-28
Remove vendored version of requests in botocore and bump version

kevchentw:
### Description of Change
New version of botocore remove vendored version of requests, therefore it should remove all code using `botocore.vendored.requests`.

I copied the `CaseInsensitiveDict` from requests to aiobotocore to use this structures. The alternative way it to put requests into requirements, however aiobotocore didn't use other reqeusts function, so I think it is unnecessary.

To reproduces this bug, update botocore to >= 1.13.0, and make requuests with aiobotobore.
The following error will generated.

```
from botocore.vendored.requests.structures import CaseInsensitiveDict
ModuleNotFoundError: No module named 'botocore.vendored.requests.structures'
```

ref:
- https://aws.amazon.com/blogs/developer/removing-the-vendored-version-of-requests-from-botocore/
- https://github.com/boto/botocore/pull/1829

bump version

1. botocore==1.12.252 -> botocore==1.13.2
2. boto3==1.9.252 -> boto3==1.10.2
3. awscli==1.16.262 -> awscli==1.16.266

  * [V ] Detailed description of issue
  * [V ] Alternative methods considered (if any)

### DIFF
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -11,12 +11,12 @@ from botocore.endpoint import EndpointCreator, Endpoint, DEFAULT_TIMEOUT, \
 from botocore.exceptions import ConnectionClosedError
 from botocore.hooks import first_non_none_response
 from botocore.utils import is_valid_endpoint_url
-from botocore.vendored.requests.structures import CaseInsensitiveDict
 from botocore.history import get_global_history_recorder
 from multidict import MultiDict
 from urllib.parse import urlparse
 
 from aiobotocore.response import StreamingBody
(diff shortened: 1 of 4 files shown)


thehesiod:
cool, thank you!  however we should do what botocore is doing and not copy code from them.  Also, while working on https://github.com/aio-libs/aiobotocore/pull/659 I found that somehow we missed a bunch of changes to make use match botocore, so we should pull those as well when bumping botocore base version.  Thanks for your help!

thehesiod:
closing in favor of https://github.com/aio-libs/aiobotocore/pull/743 as we shouldn't need to copy the case insensitive dict as that PR shows.

PR #747, opened 2019-12-17
Event streams

terrycain:
# Whats broken

There is a rarely used EventStream class used for streaming API requests, currently only used by S3 Select and Kinesis Subscribe to Shard. In the non-async world you would run something like
```python
for event in obj['Payload']:
    print(obj)
```
In an async world you would then expect to run the following:
```python
async for event in obj['Payload']
    print(object)
```
Currently when you try and iterate over obj['Payload'] it does not work as botocore is expecting non-existent methods and attributes from StreamReader.

# How its fixed

Have built an async variant of the EventStream class which will read chunks of the response for which there could be many over a long period of time (i.e a few seconds) and push the chunks into the existing "buffer" class which deals with the splitting of the chunks into useful records, and then returning those useful records, formatted from the async iterator.

The new EventStream class, aptly named AioEventStream is then loaded into sessions component registry using the `response_parser_factory` name which botocore later requests.

# Tests

Currently Moto does not support any API calls that return streaming responses. I have ran the code against a real S3 bucket, copied the bytestring chunks the original botocore read and have simulated that.

* Test to ensure AioEventStream returns the same results as EventStream
* Test to check the required classes have not been changed
* Test to check the supported protocol parsers have not changed

## test_protocol_parsers
This just checks botocore has not added new protocol formats. As we need to override the function that creates the event streams as that function hardcodes the EventStream class. - Overrides in `parsers.py`

## test_patches
We have overridden `ResponseParserFactory`, `ResponseParser` and `EventStream`
### ResponseParserFactory
In `parsers.py`, `ResponseParserFactory` has been subclassed by `AioResponseParserFactory` overriding the `create_parser` method so we can use our own `PROTOCOL_PARSERS` dict.

### ResponseParser
In `parsers.py` each parser is based on a class deep in a class hierarchy which eventually all stem from `ResponseParser`, so instead of recreating the 10+ nested class mess, we just override the `_create_event_stream` to change the event stream class to our own. So currently that is the only method of concern.

### EventStream
In `eventstream.py` there is AioEventStream, its a bit complicated but all of that work is just to replace the iterator part of EventStream so that it is async.

# Affected issues

terrycain/aioboto3#178

### DIFF
--- a/aiobotocore/eventstream.py
+++ b/aiobotocore/eventstream.py
@@ -52,7 +52,3 @@ class AioEventStream(EventStream):
             pass
 
         return None
-
-    async def close(self):
-        """Closes the underlying streaming body. """
-        await self._raw_stream.close()
--- a/aiobotocore/eventstream.py
+++ b/aiobotocore/eventstream.py
@@ -1,54 +1,25 @@
 from botocore.eventstream import EventStream, EventStreamBuffer
+from async_generator import async_generator, yield_
 
 
 class AioEventStream(EventStream):
-    def __init__(self, *args, **kwargs):
-        super(AioEventStream, self).__init__(*args, **kwargs)
-        self._buffer = EventStreamBuffer()
-
-    def _create_raw_event_generator(self):
-        return self._raw_stream.iter_chunks()
+    @async_generator
+    async def _create_raw_event_generator(self):
+        event_stream_buffer = EventStreamBuffer()
+        async for chunk, _ in self._raw_stream.iter_chunks():
+            event_stream_buffer.add_data(chunk)
+            for event in event_stream_buffer:
+                await yield_(event)
 
     def __iter__(self):
         raise NotImplementedError('Use async-for instead')
 
     def __aiter__(self):
-        return self
+        return self.__anext__()
 
+    @async_generator
     async def __anext__(self):
-        # Basically this gets chunks of data from a stream,
-        # pumps them into a buffer class that does some error
-        # checking, splitting and then returns results
-        while True:
-            try:
-                chunk = await self._event_generator.__anext__()
-            except StopAsyncIteration:
-                break
-            else:
-                self._buffer.add_data(chunk[0])
-
-            event = self._get_event_from_buffer()
-            if event:
-                return event
-            # If we have no event, there might be more data needed
-            # from the stream, so loop round and try again
-
-        # The stream has been read completely, but the buffer
-        # might still have events in it.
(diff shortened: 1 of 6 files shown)


terrycain:
@thehesiod or @jettify mind reviewing when you get a chance.

I think the AioEventStream class is a bit more complicated than it should be but if you have a simpler way to achieve the same thing then I'm all ears.

thehesiod:
will do!  Thanks Terry!

aiobotocore/eventstream.py
```diff
@@ -0,0 +1,54 @@
...
+    def __aiter__(self):
+        return self
+
+    async def __anext__(self):
+        # Basically this gets chunks of data from a stream,
+        # pumps them into a buffer class that does some error
+        # checking, splitting and then returns results
+        while True:
+            try:
+                chunk = await self._event_generator.__anext__()
```
thehesiod: I think you can use the `async_generator` module I added to make this more similar to the sync version, see paginate.py for usage

aiobotocore/eventstream.py
```diff
@@ -0,0 +1,54 @@
+from botocore.eventstream import EventStream, EventStreamBuffer
+
+
+class AioEventStream(EventStream):
+    def __init__(self, *args, **kwargs):
+        super(AioEventStream, self).__init__(*args, **kwargs)
+        self._buffer = EventStreamBuffer()
+
+    def _create_raw_event_generator(self):
```
thehesiod: I think you can use the async_generator module I added to make this more similar to the sync version, see paginate.py for usage

thehesiod: I think then you won't need the `__init__` method and maybe simplifies `__anext__`

terrycain: Oh yeah, that's genius. I'll have a look this evening as I reckon your right and it should simplify everything. All of that was purely as I couldn't use `yield` inside an async function :D

terrycain:
@thehesiod Much better now :smile:

thehesiod:
🙇 glad to help :]

tokenmathguy:
This is great. Thanks @terrycain and @thehesiod. Looking forward to the merge!

thehesiod:
sorry, I thought terry was an admin, merged, will try to get something out asap

thehesiod:
@terrycain btw did you add a issue to moto to add this support?  If so could you link back to this PR from it?

thehesiod:
new ver is live!

PR #748, opened 2020-01-04
bump to next release version

thehesiod:
* fix extension of CHANGES file
* switch to new release version
* add version test


thehesiod:
release version for https://github.com/aio-libs/aiobotocore/pull/747

thehesiod:
@jettify now we don't need to worry so much about updating the version number :)   We can add some more restrictive tests later, like at least minor version bump if minor botocore version bump, etc.

Issue #756, opened 2020-02-13
pytest failing s3 tests - fails to create bucket


dazza-codes:
**Describe the bug**

- `master` is expected to pass all unit tests on a clean git clone
- pytest is failing s3 tests - fails to create bucket

**Checklist**
- [x] I have reproduced in environment where `pip check` passes without errors
- [x] I have provided `pip freeze` results
- [x] I have provided sample code or detailed way to reproduce
- this is purely unit tests on aiobotocore, so the following do not apply because `master` is expected to pass all unit tests on a clean git clone:
  - [-] I have tried the same code in botocore to ensure this is an aiobotocore specific issue
  - [-] I have tried similar code in aiohttp to ensure this is is an aiobotocore specific issue
  - [-] I have checked the latest and older versions of aiobotocore/aiohttp/python to see if this is a regression / injection

**pip freeze results**

```
$ pip freeze
-e git+git@github.com:dazza-codes/aiobotocore.git@dd860aa8c0ce43e551c00a48de03d361cf964a8a#egg=aiobotocore
aiohttp==3.3.2
alabaster==0.7.12
async-generator==1.10
async-timeout==3.0.1
atomicwrites==1.3.0
attrs==19.3.0
aws-sam-translator==1.20.1
aws-xray-sdk==2.4.3
awscli==1.16.278
Babel==2.8.0
boto==2.49.0
boto3==1.10.14
botocore==1.13.14
certifi==2019.11.28
cffi==1.14.0
cfn-lint==0.27.5
chardet==3.0.4
Click==7.0
colorama==0.4.1
coverage==4.5.4
cryptography==2.8
dill==0.3.1.1
docker==4.2.0
docutils==0.15.2
ecdsa==0.15
entrypoints==0.3
flake8==3.7.8
Flask==1.1.1
future==0.18.2
idna==2.8
idna-ssl==1.1.0
imagesize==1.2.0
importlib-metadata==1.5.0
importlib-resources==1.0.2
itsdangerous==1.1.0
Jinja2==2.11.1
jmespath==0.9.4
jsondiff==1.1.2
jsonpatch==1.25
jsonpickle==1.2
jsonpointer==2.0
jsonschema==3.2.0
MarkupSafe==1.1.1
mccabe==0.6.1
mock==4.0.1
more-itertools==8.2.0
moto==1.3.14.dev326
multidict==4.5.2
packaging==20.1
pluggy==0.13.1
py==1.8.1
pyasn1==0.4.8
[...]

PR #758, opened 2020-02-13
AioSession defaults to asyncio current loop

dazza-codes:
- apply the default in the class initializer rather
  than the utility function that wraps it.


dazza-codes:
travis tests fail for py3.5 and only for an error in creating a bucket, which is also noted in #756 and possibly (likely) not related to this PR

Issue #755, opened 2020-02-13
support moto client wrappers


dazza-codes:
This bug arises in pytest with moto `1.3.14` and although`requirements-dev.txt` has a dev-version, that fix is for something else, i.e. this is irrelevant:
```
# We need: https://github.com/spulec/moto/pull/2436
moto==1.3.14.dev326
```

See also:
- https://github.com/spulec/moto/issues/2706
- https://github.com/boto/botocore/blob/develop/botocore/endpoint.py#L38
- https://github.com/boto/botocore/blob/develop/botocore/parsers.py#L257
  - it uses a `lowercase_dict` function already

Below is an exception detail, when testing the following pytest fixtures:
```python
from moto import mock_config
from moto import mock_batch

@pytest.fixture(scope="module")
def aws_region():
    return "us-west-2"

@pytest.fixture
@pytest.mark.moto
def aio_aws_session(event_loop):
    with mock_config():
        aws_session = aiobotocore.get_session(loop=event_loop)
        yield aws_session

@pytest.fixture
@pytest.mark.moto
async def aio_aws_batch_client(aio_aws_session, aws_region):
    with mock_config():
        with mock_batch():
            async with aio_aws_session.create_client("batch", region_name=aws_region) as client:
                yield client
```

This raises a simple exception when trying to parse a moto response (below) and the source code for botocore seems to match (there is no `AWSResponse.raw_headers` attr).  Maybe there are API version differences between aiobotocore, botocore and moto (at the time of posting this issue).  In the project, the requirements pull in the aiobotocore deps for boto3/botocore and moto is the latest release:
```
aiobotocore==0.11.1
boto==2.49.0
boto3==1.10.14
botocore==1.13.14
moto==1.3.14
```

```
$ python --version
Python 3.6.7
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.4 LTS"
```

The simple test function is:
```python
@pytest.mark.asyncio
async def test_async_aws_batch_client(aio_aws_batch_client):
    assert isinstance(aio_aws_batch_client, BaseClient)
    job_queues = await aio_aws_batch_client.describe_job_queues()
    # AttributeError: 'AWSResponse' object has no attribute 'raw_headers'
```

The moto job-queues should be an empty list (and it is, see pdb details below).

```
[...]

PR #759, opened 2020-02-14
Convert botocore.awsrequest.AWSResponse to dict

dazza-codes:
- it might help if AWSResponse.to_dict() could
  take care of business, but that would be too easy

Possible fix for #755 but it's a guess really and it might be specific to `moto`.

When this PR code is used against live-s3, it works as expected to list buckets and issue a HEAD request on them, and when a print statement is added, i.e.
- `print(f"http_response is {type(http_response)}")`
- the type reported is:
```
http_response is <class 'aiobotocore._endpoint_helpers.ClientResponseProxy'>
```
With `moto`, the `http_response` is an `botocore.awsrequest.AWSResponse`, so this is something different.  Maybe something is not registered as usual when using `moto` vs. live-s3 endpoints?  Maybe `moto` is registering something with the `creating-client-class` and it mutates the `class_attributes` or something to replace some aiobotocore attributes?  There's a few events it could use to intercept aiobotocore behavior.

The unit tests are passing on this PR with both moto and live-s3 endpoints, but it's not entirely clear whether the response data is already retrieved via aiohttp at the point when the parser is called or whether parsing calls should be able to await a read() of the content.

### DIFF
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -9,6 +9,7 @@ from botocore.endpoint import EndpointCreator, Endpoint, DEFAULT_TIMEOUT, \
 from botocore.exceptions import ConnectionClosedError
 from botocore.hooks import first_non_none_response
 from botocore.utils import is_valid_endpoint_url
+from botocore.utils import lowercase_dict
 from multidict import MultiDict
 from urllib.parse import urlparse
 from urllib3.response import HTTPHeaderDict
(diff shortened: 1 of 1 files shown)


thehesiod:
sorry wasted my whole night trying to fix moto on 3.5, will have to look later

dazza-codes:
This PR should not be merged.  It fixes a problem with moto mock responses by changing how they are parsed.  This is not a good fix for that problem.  A better fix is to avoid the moto mock responses so that the test code can hit a moto server; this is described in #755.

dazza-codes:
Closing this in favor of #766

PR #766, opened 2020-02-19
aiomoto

### LINKED ISSUE #583: pytest fixtures available for external projects
It would be great to have access to pytest fixtures defined in [conftest.py](https://github.com/aio-libs/aiobotocore/blob/master/tests/conftest.py) or [mock_server.py](https://github.com/aio-libs/aiobotocore/blob/master/tests/mock_server.py) in external projects which depend on aiobotocore. 

Similar setup is done for example in [aiohttp's pytest_plugin](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/pytest_plugin.py)

What do you think about it?

### LINKED ISSUE #665: [test infra] switch to pytest-moto
https://github.com/jotes/pytest-moto

based on discussion in #583 

perhaps contributing some of our code to it where it makes sense.

### LINKED ISSUE #753: Feature request - package pytest fixtures for downstream libs
Would it be possible to add a development option to the package so that it can provide pytest fixtures to downstream projects?  Something like `aiobotocore[dev]` might provide pytest fixture methods?  (It might introduce additional namespace requirements on the test suite.)

For example, dask provides test utilities in the package, see
- https://distributed.dask.org/en/latest/develop.html#writing-tests

Maybe this is out-of-scope in this project or it means a spin-off for something like a new "aiomoto" project.

PS, this is not a bug report - the issue creation options did not include a "feature request"

Related issues:  #583 and #665 
- given that #583 was closed in favor of #665, maybe this issue will be closed as a dup, if there is no intention to follow through with providing a pytest fixture plugin

### LINKED ISSUE #755: support moto client wrappers

dazza-codes:
### Description of Change

Fix #583
Fix #665 
Fix #753 
Fix #755 
Fix spulec/moto#2706

This most likely replaces #759 and that PR should not be merged.

### Notes
- will not work in py3.5 and that's intentional
- no idea yet about how to integrate this into the rest of aiobotocore tests, so this PR simply uses nested test directories and namespaced fixtures to try to avoid conflicts, but they might need to be resolved (either in this PR or something following this).
- the `MotoService` does not provide a _clean_ backend across tests, by default, so the server fixtures in this PR add and use a `MotoService.reset` option to clear the backends

### DIFF
--- a/aiobotocore/aiomoto/aws_fixtures.py
+++ b/aiobotocore/aiomoto/aws_fixtures.py
@@ -1,0 +1,92 @@
+"""
+AWS test fixtures
+"""
+import os
+
+import boto3
+import pytest
+
+from moto import mock_batch
+from moto import mock_ec2
+from moto import mock_ecs
+from moto import mock_iam
+from moto import mock_logs
+from moto import mock_s3
+
+from aiobotocore.aiomoto.utils import AWS_REGION
+from aiobotocore.aiomoto.utils import AWS_ACCESS_KEY_ID
+from aiobotocore.aiomoto.utils import AWS_SECRET_ACCESS_KEY
+
+AWS_HOST = "127.0.0.1"
+AWS_PORT = "5000"
+
+
+@pytest.fixture
+def aws_host():
+    return os.getenv("AWS_HOST", AWS_HOST)
+
+
+@pytest.fixture
+def aws_port():
+    return os.getenv("AWS_PORT", AWS_PORT)
+
+
+@pytest.fixture
+def aws_proxy(aws_host, aws_port, monkeypatch):
+    # only required if using a moto stand-alone server or similar local stack
+    monkeypatch.setenv("HTTP_PROXY", f"http://{aws_host}:{aws_port}")
+    monkeypatch.setenv("HTTPS_PROXY", f"http://{aws_host}:{aws_port}")
+
+
+@pytest.fixture
+def aws_credentials(monkeypatch):
+    monkeypatch.setenv("AWS_ACCESS_KEY_ID", AWS_ACCESS_KEY_ID)
+    monkeypatch.setenv("AWS_SECRET_ACCESS_KEY", AWS_SECRET_ACCESS_KEY)
+    monkeypatch.setenv("AWS_SECURITY_TOKEN", "test")
+    monkeypatch.setenv("AWS_SESSION_TOKEN", "test")
+
+
+@pytest.fixture
+def aws_region():
+    return AWS_REGION
+
+
+#
+# AWS Clients
+#
+
+
+@pytest.fixture
+def aws_batch_client(aws_region):
+    with mock_batch():
+        yield boto3.client("batch", region_name=aws_region)
+
+
+@pytest.fixture
+def aws_ec2_client(aws_region):
+    with mock_ec2():
+        yield boto3.client("ec2", region_name=aws_region)
+
+
+@pytest.fixture
+def aws_ecs_client(aws_region):
+    with mock_ecs():
+        yield boto3.client("ecs", region_name=aws_region)
+
+
+@pytest.fixture
+def aws_iam_client(aws_region):
+    with mock_iam():
+        yield boto3.client("iam", region_name=aws_region)
+
+
+@pytest.fixture
+def aws_logs_client(aws_region):
+    with mock_logs():
+        yield boto3.client("logs", region_name=aws_region)
+
+
+@pytest.fixture
+def aws_s3_client(aws_region):
+    with mock_s3():
+        yield boto3.client("s3", region_name=aws_region)
--- a/aiobotocore/aiomoto/aws_fixtures.py
+++ b/aiobotocore/aiomoto/aws_fixtures.py
@@ -22,13 +22,22 @@ AWS_PORT = "5000"
 
 
 @pytest.fixture
+def free_tcp_port():
+    tcp = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    tcp.bind(('', 0))
+    addr, port = tcp.getsockname()
+    tcp.close()
+    return port
+
+
+@pytest.fixture
 def aws_host():
     return os.getenv("AWS_HOST", AWS_HOST)
 
 
 @pytest.fixture
-def aws_port():
-    return os.getenv("AWS_PORT", AWS_PORT)
+def aws_port(free_tcp_port):
+    return free_tcp_port
 
 
 @pytest.fixture
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,7 @@ def read(f):
 extras_require = {
     'awscli': ['awscli==1.18.3'],
     'boto3': ['boto3==1.12.3'],
+    'aiomoto': ['moto==1.3.14'],
 }
 
 
@@ -66,5 +67,6 @@ setup(
     python_requires='>=3.6',
     install_requires=install_requires,
     extras_require=extras_require,
-    include_package_data=True
+    include_package_data=True,
+    entry_points={'pytest11': ['aiomoto = aiobotocore.aiomoto']},
 )
--- a/aiobotocore/aiomoto/aiomoto_fixtures.py
+++ b/aiobotocore/aiomoto/aiomoto_fixtures.py
@@ -1,0 +1,184 @@
+"""
+AWS asyncio test fixtures
+"""
+
+import aiobotocore.client
+import aiobotocore.config
+import pytest
+
+from aiobotocore.aiomoto.aiomoto_services import MotoService
+from aiobotocore.aiomoto.utils import AWS_ACCESS_KEY_ID
+from aiobotocore.aiomoto.utils import AWS_SECRET_ACCESS_KEY
+
+
+#
+# Asyncio AWS Services
+#
+
+
+@pytest.fixture
+async def aio_aws_batch_server():
+    async with MotoService("batch") as svc:
+        svc.reset()
+        yield svc.endpoint_url
+
+
+@pytest.fixture
+async def aio_aws_cloudformation_server():
+    async with MotoService("cloudformation") as svc:
+        svc.reset()
+        yield svc.endpoint_url
+
+
+@pytest.fixture
+async def aio_aws_ec2_server():
+    async with MotoService("ec2") as svc:
+        svc.reset()
+        yield svc.endpoint_url
+
+
+@pytest.fixture
+async def aio_aws_ecs_server():
+    async with MotoService("ecs") as svc:
+        svc.reset()
+        yield svc.endpoint_url
+
+
+@pytest.fixture
+async def aio_aws_iam_server():
+    async with MotoService("iam") as svc:
+        yield svc.endpoint_url
+
+
+@pytest.fixture
+async def aio_aws_dynamodb2_server():
+    async with MotoService("dynamodb2") as svc:
+        svc.reset()
+        yield svc.endpoint_url
+
+
+@pytest.fixture
+async def aio_aws_logs_server():
+    # cloud watch logs
+    async with MotoService("logs") as svc:
+        svc.reset()
+        yield svc.endpoint_url
+
+
+@pytest.fixture
+async def aio_aws_s3_server():
+    async with MotoService("s3") as svc:
+        svc.reset()
+        yield svc.endpoint_url
+
+
+@pytest.fixture
+async def aio_aws_sns_server():
+    async with MotoService("sns") as svc:
+        svc.reset()
+        yield svc.endpoint_url
+
+
+@pytest.fixture
+async def aio_aws_sqs_server():
+    async with MotoService("sqs") as svc:
+        svc.reset()
+        yield svc.endpoint_url
+
+
+#
+# Asyncio AWS Clients
+#
+
+
+@pytest.fixture
+def aio_aws_session(aws_credentials, aws_region, event_loop):
+    # pytest-asyncio provides and manages the `event_loop`
+
+    session = aiobotocore.get_session(loop=event_loop)
+    session.user_agent_name = "aiomoto"
+
+    assert session.get_default_client_config() is None
+    aioconfig = aiobotocore.config.AioConfig(
+        max_pool_connections=1, region_name=aws_region
+    )
+
+    # Note: tried to use proxies for the aiobotocore.endpoint, to replace
+    #      'https://batch.us-west-2.amazonaws.com/v1/describejobqueues', but
+    #      the moto.server does not behave as a proxy server.  Leaving this
+    #      here for the record to avoid trying to do it again sometime later.
+    # proxies = {
+    #     'http': os.getenv("HTTP_PROXY", "http://127.0.0.1:5000/moto-api/"),
+    #     'https': os.getenv("HTTPS_PROXY", "http://127.0.0.1:5000/moto-api/"),
+    # }
+    # assert aioconfig.proxies is None
+    # aioconfig.proxies = proxies
+
+    session.set_default_client_config(aioconfig)
+    assert session.get_default_client_config() == aioconfig
+
+    session.set_credentials(AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY)
+    session.set_debug_logger(logger_name="aiomoto")
+
+    yield session
+
+
+@pytest.fixture
+async def aio_aws_client(aio_aws_session):
+    async def _get_client(service_name):
+        async with MotoService(service_name) as srv:
+            async with aio_aws_session.create_client(
+                service_name, endpoint_url=srv.endpoint_url
+            ) as client:
+                yield client
+
+    return _get_client
+
+
+@pytest.fixture
+async def aio_aws_batch_client(aio_aws_session, aio_aws_batch_server):
+    async with aio_aws_session.create_client(
+        "batch", endpoint_url=aio_aws_batch_server
+    ) as client:
+        yield client
+
+
+@pytest.fixture
+async def aio_aws_ec2_client(aio_aws_session, aio_aws_ec2_server):
+    async with aio_aws_session.create_client(
+        "ec2", endpoint_url=aio_aws_ec2_server
+    ) as client:
+        yield client
+
+
+@pytest.fixture
+async def aio_aws_ecs_client(aio_aws_session, aio_aws_ecs_server):
+    async with aio_aws_session.create_client(
+        "ecs", endpoint_url=aio_aws_ecs_server
+    ) as client:
+        yield client
+
+
+@pytest.fixture
+async def aio_aws_iam_client(aio_aws_session, aio_aws_iam_server):
+    async with aio_aws_session.create_client(
+        "iam", endpoint_url=aio_aws_iam_server
+    ) as client:
+        client.meta.config.region_name = "aws-global"  # not AWS_REGION
+        yield client
+
+
+@pytest.fixture
+async def aio_aws_logs_client(aio_aws_session, aio_aws_logs_server):
+    async with aio_aws_session.create_client(
+        "logs", endpoint_url=aio_aws_logs_server
+    ) as client:
+        yield client
+
+
+@pytest.fixture
+async def aio_aws_s3_client(aio_aws_session, aio_aws_s3_server):
+    async with aio_aws_session.create_client(
+        "s3", endpoint_url=aio_aws_s3_server
+    ) as client:
+        yield client
--- a/aiobotocore/aiomoto/aiomoto_services.py
+++ b/aiobotocore/aiomoto/aiomoto_services.py
@@ -1,0 +1,145 @@
+import asyncio
+import functools
+import logging
+import socket
+import threading
+import time
+import os
+
+# Third Party
+import aiohttp
+import moto.backends
+import moto.server
+import werkzeug.serving
+
+
+HOST = "127.0.0.1"
+
+_PYCHARM_HOSTED = os.environ.get("PYCHARM_HOSTED") == "1"
+CONNECT_TIMEOUT = 90 if _PYCHARM_HOSTED else 10
+
+
+def get_free_tcp_port(release_socket: bool = False):
+    sckt = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sckt.bind(("", 0))
+    addr, port = sckt.getsockname()
+    if release_socket:
+        sckt.close()
+        return port
+
+    return sckt, port
+
+
+class MotoService:
+    """ Will Create MotoService.
+    Service is ref-counted so there will only be one per process. Real Service will
+    be returned by `__aenter__`."""
+
+    _services = dict()  # {name: instance}
+
+    def __init__(self, service_name: str, port: int = None):
+        self._service_name = service_name
+
+        if port:
+            self._socket = None
+            self._port = port
+        else:
+            self._socket, self._port = get_free_tcp_port()
+
+        self._thread = None
+        self._logger = logging.getLogger("MotoService")
+        self._refcount = None
+        self._ip_address = HOST
+        self._server = None
+
+    @property
+    def endpoint_url(self):
+        return "http://{}:{}".format(self._ip_address, self._port)
+
+    def reset(self):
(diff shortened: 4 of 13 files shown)


dazza-codes:
Pushed up a revision to add a `MotoService.reset` option to clear moto.backend.BACKENDS for the service required.  It seems to work, in the context of other non-aio-moto clients that need to create fixtures, like an s3-bucket.

```
$ pytest tests/aws/
================================================================================ test session starts ================================================================================
platform linux -- Python 3.6.7, pytest-5.3.5, py-1.8.1, pluggy-0.13.1
rootdir: /home/dlweber/src/aiobotocore
plugins: cov-2.8.1, asyncio-0.10.0, forked-1.1.3, xdist-1.31.0
collected 18 items                                                                                                                                                                  

tests/aws/test_aws_fixtures.py .......                                                                                                                                        [ 38%]
tests/aws/aio/test_aiomoto_clients.py ........                                                                                                                                [ 83%]
tests/aws/aio/test_aiomoto_service.py ...                                                                                                                                     [100%]

=== 18 passed in 6.96s ===
```

### TODO
- add a fixture that creates an s3-bucket and see if tests are still passing
- the `moto.backends.BACKENDS["s3"]["global"]` model might need closer attention to detail about when and how a model instance is created and retained, but that's probably out of scope here

dazza-codes:
Some tests might be failing from something unrelated to this PR, ie
```
The command "if ! [[ -v EXTRA ]]; then make flake mototest; fi" exited with 2.
```

dazza-codes:
I can't reproduce the travis test failures locally, it seems to be some problem with the virtualenv not getting all the project packages installed or found for the tests.  Unfortunately, it works for me.

thehesiod:
wow awesome, will take a look if someone doesn't get to this before me

setup.py
```diff
@@ -76,4 +76,6 @@ def read_version():
       packages=find_packages(),
       install_requires=install_requires,
       extras_require=extras_require,
-      include_package_data=True)
+      include_package_data=True,
+      # entry_points={'pytest11': ['aiomoto = tests.aws.aio.aiomoto_fixtures.py']},
```
thehesiod: reminder need to resolve this

tests/aws/aio/aiomoto_fixtures.py
```diff
@@ -0,0 +1,184 @@
...
+@pytest.fixture
+async def aio_aws_batch_server():
+    async with MotoService("batch") as svc:
+        svc.reset()
+        yield svc.endpoint_url
+
+
+@pytest.fixture
+async def aio_aws_cloudformation_server():
+    async with MotoService("cloudformation") as svc:
```
thehesiod: I think we can replace this line across tests with a parameterized fixture

tests/aws/aio/aiomoto_fixtures.py
```diff
@@ -0,0 +1,184 @@
...
+    # Note: tried to use proxies for the aiobotocore.endpoint, to replace
+    #      'https://batch.us-west-2.amazonaws.com/v1/describejobqueues', but
+    #      the moto.server does not behave as a proxy server.  Leaving this
+    #      here for the record to avoid trying to do it again sometime later.
+    # proxies = {
+    #     'http': os.getenv("HTTP_PROXY", "http://127.0.0.1:5000/moto-api/"),
+    #     'https': os.getenv("HTTPS_PROXY", "http://127.0.0.1:5000/moto-api/"),
+    # }
+    # assert aioconfig.proxies is None
+    # aioconfig.proxies = proxies
```
thehesiod: at work I did something like this, may help:

```python
def _wrapt_boto_create_client(wrapped, instance, args, kwargs):
    def unwrap_args(service_name, region_name=None, api_version=None,
                    use_ssl=True, verify=None, endpoint_url=None,
                    aws_access_key_id=None, aws_secret_access_key=None,
                    aws_session_token=None, config=None):

        if endpoint_url is None:
            endpoint_url = os.environ.get(f'{service_name}_mock_endpoint_url')

        # https://github.com/spulec/moto/issues/2058
        aws_access_key_id = "foobar_key"
        aws_secret_access_key = "foobar_secret"

        return wrapped(service_name, region_name, api_version, use_ssl, verify,
                       endpoint_url, aws_access_key_id, aws_secret_access_key,
                       aws_session_token, config)

    return unwrap_args(*args, **kwargs)

def patch_aioboto():
    """
    Will patch aiobotocore to set endpoint_url to: {SERVICE_NAME}_endpoint_url if
    available
    """
    if not isinstance(aiobotocore.session.AioSession.create_client, wrapt.ObjectProxy):
        wrapt.wrap_function_wrapper(
            'aiobotocore.session',
            'AioSession.create_client',
            _wrapt_boto_create_client
        )
```

but ya even better would be with proxies.  It would involve us writing a proxy server that would re-direct to the appropriate endpoint based on the host in the request

thehesiod: also had this:

```python
# https://github.com/spulec/moto/issues/2058
for key in {'AWS_ACCESS_KEY_ID', 'AWS_SECRET_ACCESS_KEY'}:
    if key in os.environ:
        del os.environ[key]

def patch_aioboto():
    """
    Will patch aiobotocore to set endpoint_url to: {SERVICE_NAME}_endpoint_url if
    available
    """
    if not isinstance(aiobotocore.session.AioSession.create_client, wrapt.ObjectProxy):
        wrapt.wrap_function_wrapper(
            'aiobotocore.session',
            'AioSession.create_client',
            _wrapt_boto_create_client
        )
```

tests/aws/aio/aiomoto_fixtures.py
thehesiod: see https://stackoverflow.com/questions/18011902/pass-a-parameter-to-a-fixture-function/60148972#60148972  I think

thehesiod: then this whole section can be collapsed

tests/aws/aio/aiomoto_fixtures.py
```diff
@@ -0,0 +1,184 @@
...
+    assert session.get_default_client_config() == aioconfig
+
+    session.set_credentials(AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY)
+    session.set_debug_logger(logger_name="aiomoto")
+
+    yield session
+
+
+@pytest.fixture
+async def aio_aws_client(aio_aws_session):
```
thehesiod: I think this can be a parameterized generator fixture

tests/aws/aio/aiomoto_services.py
```diff
@@ -0,0 +1,145 @@
...
+        self._logger = logging.getLogger("MotoService")
+        self._refcount = None
+        self._ip_address = HOST
+        self._server = None
+
+    @property
+    def endpoint_url(self):
+        return "http://{}:{}".format(self._ip_address, self._port)
+
+    def reset(self):
```
thehesiod: each instance should not affect other instances, if they are affecting other instances then we should run each instance in a separate process...however, the problem with that is that some services interact with one another like SNS + lambdas.  I think each test needs a process, where all moto services for that test are run in that process.  So I think you would want the process to be in pytest nomenclature scoped to the function.

tests/aws/utils.py
```diff
@@ -0,0 +1,26 @@
+from moto.core.models import BotocoreStubber
+
+AWS_REGION = "us-west-2"
+AWS_ACCESS_KEY_ID = "test_AWS_ACCESS_KEY_ID"
+AWS_SECRET_ACCESS_KEY = "test_AWS_SECRET_ACCESS_KEY"
+
+
+def assert_status_code(response, status_code):
```
thehesiod: looks like nothing uses this yet

thehesiod:
btw the idea for the tests, in regard to buckets is that each test gets a unique bucket, see https://github.com/aio-libs/aiobotocore/blob/master/tests/conftest.py#L30

thehesiod:
It's been awhile but I think unfortunately moto shares global state across its services.  Which is a good but also bad thing.  Good because it allows cross service linkages (ex SNS messages triggering lambdas), bad because it means you can't reliably run multiple instances of the same service in the same process.  Really it should have a better way of doing this but c'est la vie. So I think we need to re-structure the moto service fixture to ensure that each function scope gets a service helper process, and then each service for the function scope runs in said helper process. This will require a bit of infra around the MotoService class. This will ensure separation of services across tests and allow for cross service linkages for each test.

thehesiod:
Actually ignore above rec for separate process per test. pytest-xdist already separates each test into a separate process

tests/aws/aio/aiomoto_services.py
thehesiod: Instead this probably should just reset the service named after itself automatically during shutdown

tests/aws/aio/aiomoto_fixtures.py
terrycain: Proxy would work, but you'd need to patch use_ssl=False right, otherwise it'll try and tunnel through the proxy using CONNECT? And at that point if your patching the ssl flag you might as well patch the endpoint url?

terrycain:
This looks great :smile: Might have to loot some of it for aioboto3 once its merged :wink:

setup.py
dazza-codes: added a commit to attempt to fix this using `aiobotocore.aiomoto`, but it might be an interim solution that could be substantially revised if some further thought into packaging might provide two packages from one repo.

tests/aws/aio/aiomoto_fixtures.py
dazza-codes: Explored some ideas along these lines but failed so far, will try to follow up with these tips (or feel free to take it and run with it on a patch on this patch, but this PR is WIP and this branch might pull the rug until it's merged, if merged).

tests/aws/aio/aiomoto_fixtures.py
dazza-codes: Yes, it was trying to tunnel through the proxy with CONNECT.

tests/aws/aio/aiomoto_services.py
dazza-codes: Without understanding all the implications, this just makes it an option without imposing it until a fixture instantiates a service.  There are ways to skip the dispatch app-generator, but AFAICT all those options also use the global BACKENDS.  There might be ways to specify both the service-name and a region to instantiate a new model for that service without adding it to BACKENDS, but didn't get that far.

dazza-codes:
Need to consider options to parametrize the server and client fixtures to address the code-cov problem.

tests/aws/aio/aiomoto_fixtures.py
thehesiod: really appreciate all the help!  Just throwing ideas out there, some wrong :)

tests/aws/aio/aiomoto_fixtures.py
dazza-codes: Tried various things and did not have any success, so I'm running out of time and patience to get this aspect of the review sorted out.  For example, the following does not work and some other ideas about returning a callable did not work:
```python
@pytest.fixture
def aws_mock_client(aws_region, request):
    service_name = request.param

    moto_module = getattr(moto, service_name)
    mock_service = getattr(moto_module, f"mock_{service_name}")
    with mock_service():
        yield boto3.client(service_name, region_name=aws_region)
```
I have to punt on this request, sorry.  I'll have to leave this PR as is so that it might be merged somehow and revised in a subsequent changeset.  The `moto` project uses some nice decorators, but I don't have time to figure out how to modify those to provide similar `aiomock_{service}` decorators (so punting on that also).

dazza-codes:
The PR was rebased on master and seems to pass unit tests OK.  It's in a state that might be "good enough to go", although there might be additional feature requests that subsequent PRs can address.  I have enough code to continue with the "real" work that needs some of this and my focus needs to shift to that.  Hope this helps a bit.

tests/aws/aio/aiomoto_fixtures.py
dazza-codes: See `test_aio_aws_client` and the fixture it uses, which returns a client generator.  The `test_aio_aws_client` could be easily parameterized to run all available clients if it has some generic assertions.

tests/aws/aio/aiomoto_fixtures.py
dazza-codes: See `test_aio_aws_client` and the fixture it uses, which returns a client generator.  The `test_aio_aws_client` could be easily parameterized to run all available clients if it has some generic assertions.

tests/aws/aio/aiomoto_services.py
dazza-codes: See `test_aio_aws_client` and the fixture it uses, which returns a client generator.  The `test_aio_aws_client` could be easily parameterized to run all available clients if it has some generic assertions.

chesstrian:
Hi @dazza-codes, do you know if there are any plans to merge this?, I'm able to reproduce the issue `AttributeError: 'AWSResponse' object has no attribute 'raw_headers'` testing some DynamoDB operations.

Thanks in advance.

dazza-codes:
My focus on this has moved to https://gitlab.com/dazza-codes/aio-aws and I've lost track of where this PR is at in the context of this project.  I doubt it's in the right form exactly to be merged for this project.  I don't know if I have time and/or enough feedback to get it into the right shape for this repo.  I understand there are some other branches with various refactoring that might conflict with this, so it's probably not worth the time right now to resolve anything until there's a new release.  Maybe the project maintainers can just cherrypick patches from this into the main line.  I know it took a lot of time and reading to understand the problem and put together this test suite, so let's see what happens as the project refactoring evolves.  At least this PR is visible as one possible solution to the test problems.

chesstrian:
Thanks @dazza-codes, I forked and merged your PR and it didn't work for DynamoDB, need to change the approach to test.

aiobotocore/aiomoto/aws_fixtures.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+@pytest.fixture
+def aws_host():
+    return os.getenv("AWS_HOST", AWS_HOST)
+
+
+@pytest.fixture
+def aws_port():
+    return os.getenv("AWS_PORT", AWS_PORT)
```
thehesiod, line 31: should use a random port or else you can't run tests in parallel

dazza-codes: Pushed a new commit to try to use a free port, although I'm not sure the patch will be used in moto.

thehesiod:
sorry for taking so long to review this, been super busy

dazza-codes:
I haven't been able to keep up with this and I'm a bit blocked on it since the move to `Pipfile` (I've been using poetry and avoiding anything `pipenv`); I don't recall the details of when/why I switched from pipenv to poetry, but I vaguely recall a bunch of issues that totally wrecked my python setups.  So far, despite a few minor hiccups with poetry, it's been a good move.

thehesiod:
nice, we were on pipenv, then got frustrated due to slowness/inability to resolve certain scenarios so moved to poetry, then found another blocking bug on it so switched back :). I initially tried the new pipenv release but it was completely busted so waiting a bit more to try the latest pipenv again

rajeshn96:
hello, @dazza-codes I too am reproducing this error 'AWSResponse' object has no attribute 'raw_headers' when testing with moto. I tried using the patched code solution suggested in [https://github.com/aio-libs/aiobotocore/issues/755#issuecomment-586117397](url) but it doesn't seem to affect the issue. Maybe I missed it in this thread, but is there a better way to deal with this error?

thehesiod:
btw new version of moto does not require an endpoint per service, all services are serviced by the same endpoint so will not be needed.  I started working on that here: https://github.com/aio-libs/aiobotocore/compare/master...thehesiod/aio-checksum-streaming-moto-bump bunch of tests to fix

PR #745, opened 2019-11-13
Update botocore, boto3 and awscli

msfernandes:
### Description of Change
Update `awscli`, `boto3` and `botocore` to the latest versions

### Checklist for All Submissions
* [x] I have added change info to [CHANGES.txt](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.txt)

### Checklist when updating botocore and/or aiohttp versions

* [x] I have read and followed [CONTRIBUTING.rst](https://github.com/aio-libs/aiobotocore/blob/master/CONTRIBUTING.rst#how-to-upgrade-botocore)
* [ ] I have updated test_patches.py where/if appropriate (also check if no changes necessary)
* [x] I have ensured that the awscli/boto3 versions match the updated botocore version

PR #767, opened 2020-02-20
Dependencies bump

terrycain:
Updated botocore/boto3/awscli dependencies, also merged in the dev deps from pyup.

The changes in args.py and client.py are purely cosmetic so that the code matches whats in botocore

### DIFF
--- a/aiobotocore/args.py
+++ b/aiobotocore/args.py
@@ -35,11 +35,6 @@ class AioClientArgsCreator(botocore.args.ClientArgsCreator):
         signing_region = endpoint_config['signing_region']
         endpoint_region_name = endpoint_config['region_name']
 
-        if signing_region is None and endpoint_region_name is None:
-            signing_region, endpoint_region_name = \
-                self._get_default_s3_region(service_name, endpoint_bridge)
-            config_kwargs['region_name'] = endpoint_region_name
-
         event_emitter = copy.copy(self._event_emitter)
         signer = RequestSigner(
             service_model.service_id, signing_region,
--- a/aiobotocore/args.py
+++ b/aiobotocore/args.py
@@ -34,6 +34,7 @@ class AioClientArgsCreator(botocore.args.ClientArgsCreator):
 
         signing_region = endpoint_config['signing_region']
         endpoint_region_name = endpoint_config['region_name']
+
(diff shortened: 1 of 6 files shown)


terrycain:
Replaces #764 
Replaces #745

aiobotocore/args.py
```diff
@@ -34,6 +34,7 @@ def get_client_args(self, service_model, region_name, is_secure,
 
         signing_region = endpoint_config['signing_region']
         endpoint_region_name = endpoint_config['region_name']
+
         if signing_region is None and endpoint_region_name is None:
```
thehesiod: from what I see diffing aiobotocore against 1.15.3 this `if signing_region` was removed along with `if isinstance(client_config, AioConfig):` and adding `connector_args` to `create_endpoint` call.  I think when you did you diff you did aiobotocore <-> 1.13.14 instead of against 1.15.3.  If you want to allow me to change your branch I can help fix, otherwise mind doing another diff against all the files?

thehesiod: actually looked at the diff and I think this is the only file that needs to be updated

thehesiod:
btw don't forget updating `CHANGES.rst`

thehesiod:
btw see my comment in #745 re ClientCreator patch check: `feel free to bump ClientCreator hash. They added some new functions. In reality we should probably only hash the _create_client_class + _get_client_args methods we're overriding that aren't forwards compatible`

aiobotocore/args.py
terrycain: You sure we want to remove
```
        if isinstance(client_config, AioConfig):
            connector_args = client_config.connector_args
        else:
            connector_args = None
```
?

terrycain: As i believe, if you pass config in via AioSession.create_client, get_default_client_config should return None, config is not altered so could end up not being an AioConfig type

thehesiod: whoops, hah, ya that was our change, my bad

thehesiod: probably later we should comment our non obvious await changes

thehesiod:
thank you @terrycain ! as demonstrated needs two minds =)

terrycain:
Np np

Issue #721, opened 2019-08-13
API documentation is empty


mivade:
Currently, navigating to https://aiobotocore.readthedocs.io/en/latest/api.html shows a page that simply has a title "API" and no content.

thehesiod:
:(  @jettify ideas?

Issue #664, opened 2019-01-21
Getting ready for Python 3.8


dimaqq:
```
  /.../python3.7/site-packages/botocore/vendored/requests/packages/urllib3/_collections.py:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import Mapping, MutableMapping
```

It seems to be a matter of upgrading to new[er] `botocore`.

dimaqq:
Version set when encountered:

```
  - aiobotocore [required: >=0.10.0, installed: 0.10.0]
    - aiohttp [required: >=3.3.1, installed: 3.5.4]
      - async-timeout [required: >=3.0,<4.0, installed: 3.0.1]
      - attrs [required: >=17.3.0, installed: 18.2.0]
      - chardet [required: >=2.0,<4.0, installed: 3.0.4]
      - multidict [required: >=4.0,<5.0, installed: 4.5.2]
      - yarl [required: >=1.0,<2.0, installed: 1.3.0]
        - idna [required: >=2.0, installed: 2.8]
        - multidict [required: >=4.0, installed: 4.5.2]
    - botocore [required: >=1.12.49,<1.12.50, installed: 1.12.49]
      - docutils [required: >=0.10, installed: 0.14]
      - jmespath [required: >=0.7.1,<1.0.0, installed: 0.9.3]
      - python-dateutil [required: >=2.1,<3.0.0, installed: 2.7.5]
        - six [required: >=1.5, installed: 1.12.0]
      - urllib3 [required: >=1.20,<1.25, installed: 1.24.1]
    - wrapt [required: >=1.10.10, installed: 1.11.1]
```

dimaqq:
How about this one?
```
site-packages/aiobotocore/endpoint.py:272: DeprecationWarning: The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.
    await asyncio.sleep(handler_response, loop=self._loop)
```

thehesiod:
re-opened ;)

thehesiod:
so I think we should drop a new minor release where we drop the loop param.  Also make warnings fatal during error testing, which means we probably need a pytest.ini to ignore warnings from external modules.

thehesiod:
another option is we could keep the loop param, throw a warning but not pass it to asyncio.sleep, but this could break behavior without people knowing, so i rather just kill it in one shot.

PR #769, opened 2020-02-21
remove 3.5, better support 3.8, Pipfile

thehesiod:
- remove loop parameter as it's deprecated in 3.8 and never really needed
- disable 3.5 as it's going to be nearly impossible to get it to work
- switch to Pipfile to ensure valid sets of modules
- sync and pinpoint patch checks

fixes https://github.com/aio-libs/aiobotocore/issues/664
fixes https://github.com/aio-libs/aiobotocore/issues/721


thehesiod:
@terrycain you inspired me given you bumped to the next minor version :)

thehesiod:
btw previously it was building branches + pull requests, now it will only build pull requests

aiobotocore/args.py
```diff
@@ -45,13 +41,14 @@ def get_client_args(self, service_model, region_name, is_secure,
 
         config_kwargs['s3'] = s3_config
 
+        # aiobotocore addition
```
terrycain: :+1:

setup.py
```diff
@@ -4,20 +4,20 @@
...
-    'async_generator>=1.10',  # can remove if we move to py3.6+
+    'aioitertools>=0.5.1',
 ]
 
 PY_VER = sys.version_info
 
-if not PY_VER >= (3, 5, 3):
-    raise RuntimeError("aiobotocore doesn't support Python earlier than 3.5.3")
+if not PY_VER >= (3, 6, 0):
+    raise RuntimeError("aiobotocore doesn't support Python earlier than 3.6.0")
```
terrycain: So https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires exists, would using that be better?

tests/test_patches.py
```diff
@@ -54,22 +59,64 @@
...
+    ClientArgsCreator.get_client_args: {'e3a44e6f50159e8e31c3d76f5e8a1110dda495fa'},
+
+    # client.py
     ClientCreator._create_client_class: {'5e493d069eedbf314e40e12a7886bbdbcf194335'},
     ClientCreator._get_client_args: {'555e1e41f93df7558c8305a60466681e3a267ef3'},
-    BaseClient: {'d2e69f0184ae83df5b82284194ed1b72191f0161'},
+
+    BaseClient._make_api_call: {'0c59329d4c8a55b88250b512b5e69239c42246fb'},
+    BaseClient._make_request: {'033a386f7d1025522bea7f2bbca85edc5c8aafd2'},
+    BaseClient.get_paginator: {'c69885f5f73fae048c0b93b43bbfcd1f9c6168b8'},
```
terrycain: Yes, narrows down the diffing :smile:

thehesiod: actually I was thinking we can probably do this automatically via inspection

setup.py
thehesiod: great idea: https://github.com/aio-libs/aiobotocore/pull/769/commits/965700b457556be01b2e06c806e53c9b8149af0a

thehesiod:
thanks @terrycain !

deansg:
Can someone explain to me the dropping of the loop parameter in more detail? I tried to look into Python 3.8's release notes to see what changed regarding event loops, but didn't find anything that seemed relevant. What changed there that triggered the change here?

terrycain:
So there are few reasons. Partially this - https://bugs.python.org/issue34728 / https://bugs.python.org/issue36373
Mainly, back around Py3.5.2 I'm pretty sure there was a bug where getting the event loop would not always return the correct loop that is in use, so to solve that one passed the loop around everywhere, since we've dropped 3.5 support we can get the loop reliably now. (Might be totally wrong but for some reason I remember that)

deansg:
@terrycain Thank you for your response. I think it would've been clearer if the release notes/commit specifically mentioned that what triggered this change is merely a change in asyncio.sleep, and not some general Python 3.8 feature (which can be more worrying for Python <=3.7 users)

thehesiod:
not only sleep, see: https://docs.python.org/3/whatsnew/3.8.html and the second bug Terry mentioned.

```In asyncio, the explicit passing of a loop argument has been deprecated and will be removed in version 3.10 for the following: asyncio.sleep(), asyncio.gather(), asyncio.shield(), asyncio.wait_for(), asyncio.wait(), asyncio.as_completed(), asyncio.Task, asyncio.Lock, asyncio.Event, asyncio.Condition, asyncio.Semaphore, asyncio.BoundedSemaphore, asyncio.Queue, asyncio.create_subprocess_exec(), and asyncio.create_subprocess_shell().```

thehesiod:
basically loop param is deprecated as you can only have one loop per thread, so we're taking the plunge early :)

PR #773, opened 2020-02-23
add batch client + test

thehesiod:


dazza-codes:
Apache 2.0 code related to this is available in
- https://gitlab.com/dazza-codes/python-notes/-/tree/master/notes/aio_aws

But IMO that code is a higher-order concern than aiobotocore.  Also, that code may be split out of that project into its own project where additional features may be added, including small UI/app features.

thehesiod:
hijacking this PR for some more cleanup as well :)

thehesiod:
thanks @dazza-codes !

PR #786, opened 2020-04-01
Fixes get_frozen_credentials not being async

terrycain:
### Description of Change
If credentials are passed explicitly, the session calls self.set_credentials, which assigns `botocore.credentials.Credentials` to `self._credentials`. Later on when `session.get_credentials()` is called, a `botocore.credentials.Credentials` object will be passed to the signer, which will try and call `await creds.get_frozen_credentials()` which fails as `get_frozen_credentials` is not async.

  * [ ] Why needed

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,10 @@
 Changes
 -------
+
+1.0.1 (2020-04-01)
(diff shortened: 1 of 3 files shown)


thehesiod:
@terrycain thanks!

Issue #789, opened 2020-04-05
BaseClient.__getattr__ broken after emit_until_response switched to async


thehesiod:
@terrycain any ideas?  `__getattr__` on this class calls `emit_unti_response` and that method is now async.  I'm hitting the warning that the method is not being awaited all the time in the debugger because it's trying to figure out what attributes are available

thehesiod:
for now the debugger is just calling `__len__` so if we shortcut that at least debugging will be nicer

terrycain:
Yeah my debugger got rather unhappy with that function, really screwed up breakpoints in the past.

So, looks like the only _current_ use for that function is to provide an alias for list_hits_for_qualification_type when using mechanical turk - https://github.com/boto/botocore/blob/c58d7be987f2f568e51ad95f7a16e2aa6aa709f4/botocore/handlers.py#L919

For a quick fix you could do something like:
```python
def __getattr__(self, item):
    if self._service_model.service_id.hyphenize() == 'mturk' and item == 'list_hi_ts_for_qualification_type':
        return self.list_hits_for_qualification_type

    raise AttributeError(...)
```
But it does prevent the use of the event tree thingy which we're trying to avoid.

We could return some sort of async wrapper function that does the emit logic and raises attribute error when its awaited, not ideal but might work.

terrycain:
@thehesiod thoughts?

thehesiod:
@terrycain I was inspired :). https://github.com/aio-libs/aiobotocore/pull/803

thehesiod:
let me know what you think!  I thought it's kinda elegant

thehesiod:
bad news, going to have to revert the fix as it breaks ddtrace.  See https://github.com/aio-libs/aiobotocore/pull/814

thehesiod:
I don't think we'll ever be able to support it

terrycain:
Might as well ignore until someone really requires it

PR #790, opened 2020-04-05
disable client getattr

thehesiod:


thehesiod:
to temporarily resolve https://github.com/aio-libs/aiobotocore/issues/789

Issue #781, opened 2020-03-21
stubber for async HTTP mocking


dazza-codes:
RE https://botocore.amazonaws.com/v1/documentation/api/latest/reference/stubber.html
- https://github.com/boto/botocore/blob/develop/botocore/stub.py

Is there a similar async HTTP stubber available?  How could a unit test mock a client error?

For example, the following unit test runs OK and hits a 403-Forbidden client error (because this test hits real AWS services with fake credentials).  The intention is to use an HTTP stubber to return various kinds of client error responses so that unit tests can check handlers for many types of client error.

```python
import aiobotocore
import aiobotocore.config
import botocore.exceptions

@pytest.mark.asyncio
async def test_aio_s3_bucket_head_too_many_requests():

    session = aiobotocore.get_session()
    aioconfig = aiobotocore.config.AioConfig(max_pool_connections=1)
    session.set_default_client_config(aioconfig)
    session.set_credentials("fake_AWS_ACCESS_KEY_ID", "fake_AWS_SECRET_ACCESS_KEY")

    async with session.create_client("s3") as client:

        # TODO: HOW TO ADD HTTP STUBBER HERE, similar to:
        #   https://botocore.amazonaws.com/v1/documentation/api/latest/reference/stubber.html

        with pytest.raises(botocore.exceptions.ClientError) as err:
            await client.head_bucket(Bucket="missing-bucket")

        msg = err.value.args[0]
        assert "HeadBucket operation" in msg
        assert "403" in msg
        assert "Forbidden" in msg
```

terrycain:
The standard botocore stubber works fine:
```python
import aiobotocore
import asyncio
import datetime
from dateutil.tz import tzutc
from botocore.stub import Stubber, ANY
from pprint import pprint


async def main():
    session = aiobotocore.get_session()
    response = {'AcceptRanges': 'bytes',
                'ContentLength': 4,
                'ContentType': 'binary/octet-stream',
                'ETag': '"00000000000000000000000000000000"',
                'LastModified': datetime.datetime(2019, 8, 13, 14, 16, 31, tzinfo=tzutc()),
                'Metadata': {},
                'ResponseMetadata': {'HTTPHeaders': {'accept-ranges': 'bytes',
                                                     'content-length': '4',
                                                     'content-type': 'binary/octet-stream',
                                                     'date': 'Sun, 22 Mar 2020 19:57:03 GMT',
[...]

PR #803, opened 2020-04-19
fix getattr

thehesiod:


terrycain:
Quite genius :smile:

Couldn't the test for mturk be stubbed with something like https://github.com/aio-libs/aiobotocore/issues/781#issuecomment-602265458 ?

thehesiod:
awesome ya will do

PR #831, opened 2020-10-01
fix AioPageIterator search method

joseph-jones:
### Description of Change
the Base class PageIterator has a search method that does jmespath query on the result. The original search method can not be called from AioPageIterator instances. this is mirroring the behavior from PageIterator.

### Assumptions

  * [ ] Why needed


thehesiod:
thanks!  will look at this today, was moving last couple weeks

thehesiod (approved):

Issue #833, opened 2020-11-03
Use in AWS Lambda with minimal layer sizes


dazza-codes:
It would help if the test-CI matrix and the setup dependencies supported the current versions of AWS Lambda for boto3 and botocore.  It would be useful if one cell in a test-CI matrix pinned the boto3/botocore versions to the exact patch versions used in the current AWS Lambda runtime versions (and possibly ran in a container using the lambci runtime used by AWS Lambda for each python version in the runtimes available; excluding anything that does not support asyncio, of course).  At the time of writing, the setup.py does not allow a range of botocore or boto3 versions that are compatible with the current AWS Lambda versions; the current AWS Lambda runtime versions are listed in [1] as follows.

[1] https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

```
Python runtimes
--

Python 3.8 | python3.8 | boto3-1.14.40 botocore-1.17.40 | Amazon Linux 2
Python 3.7 | python3.7 | boto3-1.14.40 botocore-1.17.40 | Amazon Linux
Python 3.6 | python3.6 | boto3-1.14.40 botocore-1.17.40 | Amazon Linux
Python 2.7 | python2.7 | boto3-1.14.40 botocore-1.17.40 | Amazon Linux
```

The `setup.py` contains the following at the time of writing:

```python
install_requires = [
    # pegged to also match items in `extras_require`
    'botocore>=1.17.44,<1.17.45',
    'aiohttp>=3.3.1',
    'wrapt>=1.10.10',
    'aioitertools>=0.5.1',
]
extras_require = {
    'awscli': ['awscli==1.18.121'],
    'boto3': ['boto3==1.14.44'],
}
```

It could help if `setup.py` could keep track of AWS Lambda runtime versions and allow a range of versions that are compatible, e.g. maybe (?)

```python
    'botocore>=1.17.40,<1.17.45',

    'boto3': ['boto3>=1.14.40,<1.14.45'],
```

Of course, doing so would entail that a test-CI matrix uses the actual AWS Lambda pinned versions to check that everything works OK with those releases.  Then it should be possible to easily and confidently use aiobotocore in AWS Lambda layers and functions.

It might be interesting to try lambCI
- https://github.com/lambci/lambci
- https://stackshare.io/stackups/lambci-vs-travis-ci
- but it seems to require an AWS account for a CloudFormation deployment (it's not a service)

### Motivation for Specific AWS Lambda Compatibility

[...]

Issue #36, opened 2016-03-06
[Tracking] roll into botocore


thehesiod:
just wondering, thanks!

jettify:
I do not object, just do not see clean way how to accomplish this. Could be tricky because `botocore` also supports `python2.7` and `aiobotocore` uses `yield from` heavily.

thehesiod:
perhaps making the aio portion available for python3.x only? I think they do a similar thing with pyzmq: http://pyzmq.readthedocs.org/en/latest/api/zmq.asyncio.html

thehesiod:
I see this is being tracked in https://github.com/boto/botocore/issues/458

jettify:
the best way to handle this issue if client could be somehow abstracted from query/url/headers builder from IO. Nice approach also used here https://github.com/cablehead/python-consul

graingert:
@jettify an option for abstraction of IO from any code: https://gist.github.com/graingert/ca6cdd5d9ae2e18ca917b4594ac8a633

jettify:
We need to start thinking how to patch `botocore` so we can plugging our code using public api, for instance first easy change is to update https://github.com/boto/botocore/blob/develop/botocore/session.py#L108-L109 `Session` object to accept `ClientCreator`, so in `aiobotocore` we just pass `AioClientCreator` without coping screen of code.

thehesiod:
perhaps we can have a class by class goal of rolling into botocore to make it bite-sized chunks of work.  But I think before that we need a high-level strategy defined of how we want this to integrate into botocore to create a straw-man we can work towards and get botocore dev approval on the design.

thehesiod:
another option is switching to the c++ library: https://github.com/aws/aws-sdk-cpp this would greatly increase performance as well, mainly due to signing.

dazza-codes:
The general pattern in botocore is to register components or to hook up callbacks to an event-emitter (which itself could be a component).  In that context, aiobotocore might create and register components for:
- custom event emitter that uses coroutines that can be awaited
- custom client creator and client that can await event callbacks
- custom client event handlers/coroutines that can be awaited

It’s not clear whether this would be any different, in the end, than what aiobotocore is already.  This might be an alternative pattern to explore in aiobotocore as a POC, but it’s not clear whether it would work and it’s got little chance, if any, of ever getting merged into botocore (lots of useful PRs are stale on botocore).

PR #834, opened 2020-11-03
Use botocore and boto3 compatible with AWS Lambda

dazza-codes:
Partial fix for #833 but it doesn't modify any CI matrix settings.

This PR might get replaced by #835 but this PR is smaller and tries to use the current packaging setup.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -9,7 +9,7 @@ coverage = "==5.0.3"
 flake8 = "==3.7.9"
 flake8-formatter-abspath = "==1.0.1"
 docker = '<4'  # fix: client version 1.39 is too new. Maximum supported API version is 1.38
-moto = {extras = ["server"],version = "==1.3.14"}
+moto = {extras = ["server"],version = "==1.3.16"}
 idna = "==2.8"  # broken pipenv resolver
 pytest = "==5.3.5"
 pytest-cov = "==2.8.1"
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ from setuptools import setup, find_packages
 # NOTE: When updating botocore make sure to update awscli/boto3 versions below
 install_requires = [
     # pegged to also match items in `extras_require`
-    'botocore>=1.17.44,<1.17.45',
+    'botocore~=1.17.40',
     'aiohttp>=3.3.1',
     'wrapt>=1.10.10',
     'aioitertools>=0.5.1',
@@ -19,8 +19,8 @@ def read(f):
 
 
 extras_require = {
-    'awscli': ['awscli==1.18.121'],
-    'boto3': ['boto3==1.14.44'],
+    'awscli': ['awscli~=1.18.121'],
+    'boto3': ['boto3~=1.14.40']
 }
 
 


dazza-codes:
I don't use pipenv routinely and it looks like I missed updating a lock file or something in some of the travis builds:
```
[pipenv.exceptions.ResolutionFailure]: Warning: Your dependencies could not be resolved. You likely have a mismatch in your sub-dependencies.
  First try clearing your dependency cache with $ pipenv lock --clear, then try the original command again.
 Alternatively, you can use $ pipenv install --skip-lock to bypass this mechanism, then run $ pipenv graph to inspect the situation.
  Hint: try $ pipenv lock --pre if it is a pre-release dependency.
ERROR: Could not find a version that matches botocore<1.17.50,<1.18.0,==1.17.44,>=1.17.40,>=1.17.49 (from -r /tmp/pipenv2vfd09h5requirements/pipenv-hx71qvoy-constraints.txt (line 4))
```
I'm more familiar with using poetry for pure-pip projects.  I vaguely recall running into a ton of shell alias problems when using pipenv.  I started using it after working with rbenv for a long time, but switched to using conda to provide a python version and then using poetry to manage the dependencies.  Python packaging and packaging solutions are complicated, but the developments for better sem-ver resolutions are good.  (While learning scala, I really like the sbt tooling.)

mxr:
@dazza-codes I'm not super familiar with botocore & aibotocore so perhaps this advice is not helpful but I think the issue is a conflict between the botocore versions in `install_requires` vs `extras_require` in addition to a conflict between what `awscli~=1.18.121` needs and what botocore versions are specificed. I was able to get `pipenv lock` to succeed with this additional change

```diff
diff --git a/setup.py b/setup.py
index 12b00a7..46fd099 100644
--- a/setup.py
+++ b/setup.py
@@ -9,3 +9,3 @@ install_requires = [
     # pegged to also match items in `extras_require`
-    'botocore~=1.17.40',
+    'botocore~=1.19.0',
     'aiohttp>=3.3.1',
@@ -22,3 +22,3 @@ extras_require = {
     'awscli': ['awscli~=1.18.121'],
-    'boto3': ['boto3~=1.14.40']
+    'boto3': ['botocore~=1.19.0']
 }
```

Do you think it's worth trying?

toidi:
@dazza-codes this PR is stale. Dependencies were updated.

thehesiod:
wow sorry for not noticing this earlier, this has been done in https://github.com/aio-libs/aiobotocore/pull/844

thehesiod:
closing since already handled in above PR, also I updated CI to use github actions, I think just need to remove travis in another PR

mxr:
@thehesiod thanks for the fix!

thehesiod:
it helps when you need it for the company you work for as well =)

dazza-codes:
The point of this MR was to relax the very specific constraints on the versions that this library depends on.  I would drop the optional dependency on awscli entirely because that should never get pip installed (IMO).  Trying to satisfy the dependency versions of all the awscli and botocore can sometimes put pip or poetry into a spin to find all the 100s or 1000s of versions available for these packages.  Using a very tight pin in aiobotocore on very specific versions of botocore is a bit too safe, unless aiobotocore will be released in lock-step with botocore releases (1000s of releases).

thehesiod:
I just realized i ignored there ~= change, i think that's a good idea with the full version so ppl can use third party patches from custom servers

thehesiod:
actually the `~=` change is not possible.  The current versioning already allows users to have their own override.  For example with the current version: `'botocore>=1.22.8,<1.22.9',` you can publish a 1.22.8.1 version and it will be picked up and preferred

dazza-codes:
@thehesiod - I have run into circumstances where botocore has released a change to a private method that breaks aiobotocore, so I understand the concerns about tight pinning and I understand that it may be unfortunate for aiobotocore to patch private methods on botocore.  I think the bigger problem is that aiobotocore touches private details in botocore, and therefore we have to throw away any assumptions about sem-ver behavior :cry:.  So, even if the approach in this PR were adopted, the sem-ver assumptions are invalid anyway.  In this regard, although I would hope the `~=` version specifier could work, I'm not sure it would and I can understand closing this PR.  (Aside - going back to [earlier comments elsewhere](https://github.com/aio-libs/aiobotocore/issues/36#issuecomment-586645546) about event hooks some time ago, I still don't know if the event architecture is a better solution.)

Anyhow, just to add a brief comment here about understanding the `~=` operator.  In https://www.python.org/dev/peps/pep-0440/#version-specifiers the `~=` allows a "compatible" version.  For example, something like `botocore~=1.22.8` allows any release that is sem-ver "compatible".  Since this specifier includes a patch version, it should be equivalent to:  `>=1.22.8,==1.22.*`.  Assuming that botocore would not break any substantial methods in a `1.22.*` release, this might be OK.  

This is based on looking at the details in https://www.python.org/dev/peps/pep-0440/#compatible-release ... e.g.

```
~= 2.2
>= 2.2, == 2.*

~= 1.4.5
>= 1.4.5, == 1.4.*
```

PR #844, opened 2021-01-12
bump botocore

thehesiod:


thehesiod:
working on switching to github actions since travis is no longer working due to lack of credits in https://github.com/aio-libs/aiobotocore/pull/846

PR #863, opened 2021-04-23
TCP_CONNECTOR(ssl_context) deprecated. Change to ssl

slipovenko:
### Description of Change
Fixed aiohttp.TCPConnector deprecation warning. Must use ssl instead of ssl_context when instantiating one.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,9 @@
 Changes
 -------
+1.3.1 (2021-04-23)
+^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


thehesiod:
this may make aiobotocore incompatible with older versions of aiohttp, which versions support this new key?

slipovenko:
ssl_context is deprecated since version 3.0

thehesiod:
cool, could you add your change to CHANGES.rst?

thehesiod:
thank you!

Issue #872, opened 2021-07-10
AioJSONParser can't read the initial response from the stream


bashalex:
**Describe the bug**

The library fails to read (seems like any) Stream response.

In particular,  `AioJSONParser` fails to read the initial response from the stream.

When  `AioJSONParser.parse()` is called, internally it calls `_handle_event_stream()` which tries to call the `get_initial_response()` method of the `AioEventStream` which expectedly fails to execute `next(self._event_generator)` since it is an Async Generator.

To reproduce, you can call the `subscribe_to_shard()` method of the Kinesis client:
```
async with self._session.create_client(
        'kinesis',
        region_name=AWS_REGION_NAME,
        aws_access_key_id=AWS_ACCESS_KEY_ID,
        aws_secret_access_key=AWS_SECRET_ACCESS_KEY,
) as client:
    starting_position = {
        'Type': 'LATEST'
    }
    response = await client.subscribe_to_shard(
        ConsumerARN=CONSUMER_ARN,
        ShardId=SHARD_ID,
        StartingPosition=starting_position
    )
```

**Environment:**
 - Python Version: 3.9.5
 - OS name and version: WSL 2 (Ubuntu 18)

thehesiod:
will look asap

bashalex:
Thanks! Here is the stacktrace:
```
    ...
    response = await client.subscribe_to_shard(
  File "/home/bashalex/.cache/pypoetry/virtualenvs/asset-service-PCliylm9-py3.9/lib/python3.9/site-packages/aiobotocore/client.py", line 141, in _make_api_call
    http, parsed_response = await self._make_request(
  File "/home/bashalex/.cache/pypoetry/virtualenvs/asset-service-PCliylm9-py3.9/lib/python3.9/site-packages/aiobotocore/client.py", line 161, in _make_request
    return await self._endpoint.make_request(operation_model, request_dict)
  File "/home/bashalex/.cache/pypoetry/virtualenvs/asset-service-PCliylm9-py3.9/lib/python3.9/site-packages/aiobotocore/endpoint.py", line 89, in _send_request
    success_response, exception = await self._get_response(
  File "/home/bashalex/.cache/pypoetry/virtualenvs/asset-service-PCliylm9-py3.9/lib/python3.9/site-packages/aiobotocore/endpoint.py", line 122, in _get_response
    success_response, exception = await self._do_get_response(
[...]

PR #873, opened 2021-07-12
fix AioJSONParser

thehesiod:
#872


thehesiod:
FYI @terrycain, another expansion.  I found a couple missing global methods: `create_parser` + `get_response`

terrycain:
:smile: One day I'll be back using AWS again and can help out more

PR #877, opened 2021-08-20
fix retries and refactor httpsession

thehesiod:
~*supposedly* I can remove the identity fix given we now disable compression, we shall see if it works~

with freshening up of links and making API more in-line with botocore


thehesiod:
adding reviews in case you guys are free, otherwise will merge after some more local testing

thehesiod:
actually reading my own bug: https://github.com/boto/botocore/issues/1255 looks like we can't, will put back

CHANGES.rst
```diff
@@ -1,5 +1,9 @@
 Changes
 -------
+1.3.3 (2021-08-20)
```
mariaines, line 3: nit: 1.3.4 ?

thehesiod: 🤦🏼 looks like test-version wasn't run for some reason to catch this, trying rename

thehesiod:
ooo, we'll need a minor bump because the exceptions raised will be different

CHANGES.rst
thehesiod: hah, looks like files with dashes aren't tested

mariaines:
Hmm, I am getting SSLError:
```
SSL validation failed for https://dynamodb.us-west-2.amazonaws.com/ Cannot connect to host dynamodb.us-west-2.amazonaws.com:443 ssl:True [SSLCertVerificationError: (1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1131)')]
```

No changes except updating aiobotocore to this branch and botocore to 1.20.106:
Screen Shot 2021-08-19 at 9 58 57 PM

mariaines:
Might be https://github.com/boto/boto3/issues/2733, let me see if I can get around it

thehesiod:
@mariaines trying locally

thehesiod:
fyi on gitter if you need quick resolution

thehesiod:
repro'd, investigating

mariaines:
Same problem when I try pinning certifi==2020.11.8 or even certifi==2018.4.16 (as suggested [here](https://stackoverflow.com/a/60673056/2034082)) so I guess maybe not boto/boto3#2733

No rush, mostly just curious to get this working. Appreciate the investigation!

thehesiod:
fun-fun, just pushed up fix

CHANGES.rst
thehesiod: now @ 1.4.0, marking resolved

mariaines:
woot seems to be working! Thank you for the quick fixes!

thehesiod:
to all those following this, you now have to set the `AIOBOTOCORE_DEPRECATED_1_4_0_APIS` env var for the retries to fix

PR #887, opened 2021-10-03
Fixed pipfile and missing close() on httpsession

terrycain:
### Description of Change
AioBaseClient.close calls the close method of the internal http session... which doesn't exist.

### DIFF
--- a/aiobotocore/httpsession.py
+++ b/aiobotocore/httpsession.py
@@ -105,6 +105,9 @@ class AIOHTTPSession:
     async def __aexit__(self, exc_type, exc_val, exc_tb):
         await self._session.__aexit__(exc_type, exc_val, exc_tb)
 
+    async def close(self):
+        await self._session.__aexit__(None, None, None)
+
     def _get_ssl_context(self):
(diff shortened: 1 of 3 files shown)


terrycain:
How do we go about updating changes.rst so that the test doesnt get unhappy :smile:

aiobotocore/httpsession.py
```diff
@@ -105,6 +105,9 @@ async def __aenter__(self):
     async def __aexit__(self, exc_type, exc_val, exc_tb):
         await self._session.__aexit__(exc_type, exc_val, exc_tb)
 
+    async def close(self):
+        await self._session.__aexit__(None, None, None)
```
thehesiod, line 109: is aexit on this class still ok if close is called?

thehesiod:
can go ahead and bump the version in `__init__.py`

thehesiod:
actually I think I'm going to merge this into a new PR with the ssl fix

thehesiod:
moved to https://github.com/aio-libs/aiobotocore/pull/888

PR #888, opened 2021-10-05
fix verify + httpsession.close

thehesiod:


isidentical:
Thanks for the fix @thehesiod!

terrycain:
Nice 😄

PR #914, opened 2022-01-26
feat: add a non-blocking implementation of the adaptive retry strategy

### LINKED ISSUE #912: Event loop blocked during client-side rate limiting?
**Describe the bug**

This is more of a question than an actual bug at this point. I don't understand how aiobotocore avoids blocking when a retry handler is installed.

Let's assume we have an aiobotocore client with the "adaptive" retry handler installed:

```python
import asyncio
from aiobotocore.config import AioConfig

async def main():
    config = AioConfig(retries={"max_attempts": 5, "mode": "adaptive"})
    async with get_session().create_client("s3", config=config) as client:
        print(await client.get_object(Bucket="bucket", Key="key"))

asyncio.run(main())
```
When instantiating the client, [this code](https://github.com/aio-libs/aiobotocore/blob/0b71254446ef7ba6d93016d5a8b5b02c665a7b0b/aiobotocore/client.py#L41) is called in the `AioClientCreator.create_client` method:
```python
self._register_retries(service_client)
```
This method not overriden by aiobotocore, so from here we jump into botocore. Ultimately, we arrive at the `register_retry_handler` method [here](https://github.com/boto/botocore/blob/06b18608833ea59a501f54470efdf19926f701cd/botocore/retries/adaptive.py#L13):
```python
def register_retry_handler(client):
    # ... skipped some lines
    limiter = ClientRateLimiter(
        token_bucket=token_bucket,
    )
    client.meta.events.register(
        'before-send', limiter.on_sending_request,
    )
```

aiobotocore emits the ["before-send" event](https://github.com/aio-libs/aiobotocore/blob/0b71254446ef7ba6d93016d5a8b5b02c665a7b0b/aiobotocore/endpoint.py#L141) normally. Reading through the aiobotocore event machinery, it seems that it ends up directly calling the `limiter.on_sending_request` method registered above.

This [method](https://github.com/boto/botocore/blob/06b18608833ea59a501f54470efdf19926f701cd/botocore/retries/adaptive.py#L52) is:
```python
def on_sending_request(self, request, **kwargs):
    if self._enabled:
        self._token_bucket.acquire()
```
where `self._token_bucket.acquire()` comes from the [`TokenBucket` class](https://github.com/boto/botocore/blob/06b18608833ea59a501f54470efdf19926f701cd/botocore/retries/bucket.py#L19). The key elements are:

```python
class TokenBucket(object):

    def __init__(self, max_rate, clock, min_rate=_MIN_RATE):
        # ...
        self._lock = threading.Lock()
        self._new_fill_rate_condition = threading.Condition(self._lock)

    def acquire(self, amount=1, block=True):
        """Acquire token or return amount of time until next token available.

        If block is True, then this method will block until there's sufficient
        capacity to acquire the desired amount.

        If block is False, then this method will return True is capacity
        was successfully acquired, False otherwise.

        """
        with self._new_fill_rate_condition:
            # ...
```
so it seems to me that we may block on a `threading.Condition` here. Am I missing something? The `retry` parameter in the AioConfig object is explicitly covered by tests, so maybe there's a part of the code I'm overlooking.

**Checklist**
- [ ] I have reproduced in environment where `pip check` passes without errors
- [x] I have tried the same code in botocore to ensure this is an aiobotocore specific issue
- [x] I have tried similar code in aiohttp to ensure this is is an aiobotocore specific issue
- [ ] I have checked the latest and older versions of aiobotocore/aiohttp/python to see if this is a regression / injection

**pip freeze results**

```
aiobotocore==2.1.0
aiohttp==3.8.1
aioitertools==0.8.0
aiosignal==1.2.0
async-timeout==4.0.2
attrs==21.4.0
botocore==1.23.24
charset-normalizer==2.0.10
frozenlist==1.3.0
idna==3.3
jmespath==0.10.0
multidict==6.0.2
python-dateutil==2.8.2
six==1.16.0
urllib3==1.26.8
wrapt==1.13.3
yarl==1.7.2
```

**Environment:**
 - Python Version: [e.g. 3.6] 3.8.10
 - OS name and version: [e.g. darwin stretch] Ubuntu 20.04

**Additional context**

I stumbled into this while trying to understand in detail the retry behaviour of aiobotocore. In our production systems, we do observe that our event loop is sometimes blocked, but there's nothing that points so far to this specific issue. I'm asking this question purely based on reading the code.

Neki:
### Description of Change

Fix #912. Includes the test from #913. 

This is a draft pull request. I still need to add unit tests for the new Async classes introduced here.
If my understanding is correct, I don't need to update test_patches (these changes don't depend on any private attribute), although I rely on RateClocker not blocking despite using a `threading.Lock`.

Current tests (through `make cov` and `make mototest`) already pass.

### Assumptions

  * [ ] Why needed

### DIFF
--- a/aiobotocore/client.py
+++ b/aiobotocore/client.py
@@ -12,6 +12,7 @@ from .paginate import AioPaginator
 from .args import AioClientArgsCreator
 from .utils import AioS3RegionRedirector
 from .discovery import AioEndpointDiscoveryManager, AioEndpointDiscoveryHandler
+from .retries import adaptive
 from . import waiter
 
 history_recorder = get_global_history_recorder()
@@ -89,6 +90,43 @@ class AioClientCreator(ClientCreator):
         else:
             events.register('before-parameter-build',
                             block_endpoint_discovery_required_operations)
+
+    def _register_retries(self, client):
+        # botocore retry handlers may block. We add our own implementation here.
+        # botocore provides three implementations:
+        #
+        # 1) standard
+        # This one doesn't block. A threading.Lock is used in quota.RetryQuota,
+        # but it's only used to protect concurrent modifications of internal
+        # state inside multithreaded programs. When running under a single
+        # asyncio thread, this lock will be acquired and released in the same
+        # coroutine, and the coroutine will never block waiting for the lock.
+        # Thus, we don't need to redefine this strategy.
+        #
+        # 2) adaptive
+        # This one blocks when the client is applying self rate limiting.
+        # We override the corresponding definition to replace it with async
+        # objects.
+        #
+        # 3) legacy
+        # This one probably doesn't block.
+        #
+        # The code for this method comes directly from botocore. We could
+        # override `_register_v2_adaptive_retries` only. The override for
+        # `_register_retries` is only included for clarity.
+        retry_mode = client.meta.config.retries['mode']
+        if retry_mode == 'standard':
+            self._register_v2_standard_retries(client)
+        elif retry_mode == 'adaptive':
+            self._register_v2_standard_retries(client)
+            self._register_v2_adaptive_retries(client)
+        elif retry_mode == 'legacy':
+            self._register_legacy_retries(client)
+
+    def _register_v2_adaptive_retries(self, client):
+        # See comment in `_register_retries`.
+        # Note that this `adaptive` module is an aiobotocore reimplementation.
+        adaptive.register_retry_handler(client)
 
     def _register_s3_events(self, client, endpoint_bridge, endpoint_url,
                             client_config, scoped_config):
--- a/aiobotocore/retries/adaptive.py
+++ b/aiobotocore/retries/adaptive.py
@@ -1,0 +1,89 @@
+"""An async reimplementation of the blocking elements from botocore.retries.adaptive."""
+import asyncio
+import logging
+
+from botocore.retries import throttling
+from botocore.retries import standard
+
+# The RateClocker from botocore uses a threading.Lock, but in a single-threaded asyncio
+# program, the lock will be acquired then released by the same coroutine without
+# blocking.
+from botocore.retries.adaptive import RateClocker
+
+from . import bucket
+
+
+logger = logging.getLogger(__name__)
+
+
+def register_retry_handler(client):
+    clock = bucket.Clock()
+    rate_adjustor = throttling.CubicCalculator(starting_max_rate=0,
+                                               start_time=clock.current_time())
+    token_bucket = bucket.AsyncTokenBucket(max_rate=1, clock=clock)
+    rate_clocker = RateClocker(clock)
+    throttling_detector = standard.ThrottlingErrorDetector(
+        retry_event_adapter=standard.RetryEventAdapter(),
+    )
+    limiter = AsyncClientRateLimiter(
+        rate_adjustor=rate_adjustor,
+        rate_clocker=rate_clocker,
+        token_bucket=token_bucket,
+        throttling_detector=throttling_detector,
+        clock=clock,
+    )
+    client.meta.events.register(
+        'before-send', limiter.on_sending_request,
+    )
+    client.meta.events.register(
+        'needs-retry', limiter.on_receiving_response,
+    )
+    return limiter
+
+
+class AsyncClientRateLimiter:
+    """An async reimplementation of ClientRateLimiter."""
+    # Most of the code here comes directly from botocore. The main change is making the
+    # callbacks async.
+    # This doesn't inherits from the botocore ClientRateLimiter for two reasons:
+    # * the interface is slightly changed (methods are now async)
+    # * we rewrote the entirety of the class anyway
+
+    _MAX_RATE_ADJUST_SCALE = 2.0
+
+    def __init__(self, rate_adjustor, rate_clocker, token_bucket,
+                 throttling_detector, clock):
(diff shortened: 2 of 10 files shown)


thehesiod:
need to add overridden method to test_patches

tests/test_basic_s3.py
```diff
@@ -203,6 +205,29 @@ async def test_can_get_and_put_object(s3_client, create_object, bucket_name):
...
+
+    # Check that we did not call any blocking method
+    # Unfortunately can't patch threading.Lock.__enter__
+    with wrapt_function(
+        'botocore.retries.adaptive',
+        'ClientRateLimiter.on_sending_request',
+        _wrapt_failed
+    ), wrapt_function('botocore.retries.bucket', 'TokenBucket.acquire', _wrapt_failed):
+        await create_object('foobarbaz', body='body contents')
+        _ = await s3_client.get_object(Bucket=bucket_name, Key='foobarbaz')
```
thehesiod, line 226: we should have a positive test here as well to ensure the new methods are called

thehesiod: can be w/in this test, just wrap the new methods and ensure they're in called

aiobotocore/retries/bucket.py
```diff
@@ -0,0 +1,109 @@
+"""An async reimplementation of the blocking elements from botocore.retries.bucket."""
+import asyncio
+
+from botocore.exceptions import CapacityNotAvailableError
+from botocore.retries.bucket import Clock as Clock  # reexport # noqa
+
+
+class AsyncTokenBucket:
```
thehesiod, line 8: why no inherit? otherwise we won't pick up any methods they later add

Neki: The rationale was (I'll add a comment to this effect if you're OK with that):

The interface is different from TokenBucket. In TokenBucket, `max_rate` also has a setter. In `AsyncTokenBuket`, I switched to `async set_max_rate`. If using inheritance, we would have to override the setter to a method that raises an exception, since we no longer want anyone to use this setter.

I'm not sure we want to pick up automatically new methods here, as they may be blocking, and it's simpler to understand what's going on if the code fails with an AttributeError rather than if the event loop blocks in production.

thehesiod: that's fine if you leave a comment to this effect, however this also means in patches you should set the patch on the class itself and not the methods.  That way if anything changes in the class it will trigger and someone needs to look.   Please also put cmt in patch for them to look in the comment you'll add in AsyncTokenBucket.

aiobotocore/client.py
```diff
@@ -90,6 +91,43 @@ def _register_endpoint_discovery(self, client, endpoint_url, config):
...
+        retry_mode = client.meta.config.retries['mode']
+        if retry_mode == 'standard':
+            self._register_v2_standard_retries(client)
+        elif retry_mode == 'adaptive':
+            self._register_v2_standard_retries(client)
+            self._register_v2_adaptive_retries(client)
+        elif retry_mode == 'legacy':
+            self._register_legacy_retries(client)
+
+    def _register_v2_adaptive_retries(self, client):
```
thehesiod, line 126: if overriden these are missing in patches of these is missing in patches

aiobotocore/retries/adaptive.py
```diff
@@ -0,0 +1,86 @@
...
+    client.meta.events.register(
+        'before-send', limiter.on_sending_request,
+    )
+    client.meta.events.register(
+        'needs-retry', limiter.on_receiving_response,
+    )
+    return limiter
+
+
+class AsyncClientRateLimiter:
```
thehesiod, line 44: given you're not inheriting need to patch check the whole class

aiobotocore/retries/adaptive.py
```diff
@@ -0,0 +1,86 @@
...
+# blocking.
+from botocore.retries.adaptive import RateClocker
+
+from . import bucket
+
+
+logger = logging.getLogger(__name__)
+
+
+def register_retry_handler(client):
```
thehesiod, line 19: missing patch check on this method

thehesiod:
@Neki feel free to do what you think is best in RE to tests

aiobotocore/retries/adaptive.py
thehesiod: sorry, by patch I meant in test_patches.py make a ref to the class itself, ex:
```python
AsyncClientRateLimiter: {'xxx'},
```

thehesiod: no need for any other test changes

thehesiod: that way it will has the whole class, so if anything in the class changes it will trigger

Neki: Do you mean ClientRateLimiter (the botocore class), or AsyncClientRateLimiter (the aiobotocore one?). I added the former in test_patches.py.

Neki:
Hi @thehesiod, @simon-saliba is a coworker of mine and finished up this PR, it should be ready for review :)

thehesiod (approved):

thehesiod:
looks like there's a test failure

simon-saliba:
> looks like there's a test failure

Yes, I forgot to update to version 2.1.1 in __init__.py. All should be good now.

thehesiod:
thanks everyone!  I'll push out a release shortly

thehesiod:
hah, auto gen release notes is awesome

PR #934, opened 2022-05-05
AioAWSResponse and AioAWSRequest

thehesiod:
There are special decoding operations performed by these classes that we need in certain circumstances.  This moves us closer to botocore, however requires us to override even more methods :(

Missing:

**special.py**
- [x] RetryDDBChecksumError.is_retryable (needs RetryHandler.needs_retry down)

**retryhandler.py**
- [x] CRC32Checker._check_response


thehesiod:
@achimgaedke please let me know if this works for you

thehesiod:
holy cow is this a pain

achimgaedke:
> @achimgaedke please let me know if this works for you

Hey! Thanks for the hard work. It sounds like Whak-A-Mole...

I have taken one version 2 hours ago and used it with [DVC](https://dvc.org/doc/command-reference/pull), that's the realistic use case:

I got many of those errors:

> ERROR: failed to transfer 'md5: 37dee44a04f930b99d26ad16217dcfd1' - ClientResponseProxy.read() takes 1 positional argument but 2 were given                      
> ERROR: failed to transfer 'md5: b52cca673fe356892b47ea513c893efd' - ClientResponseProxy.read() takes 1 positional argument but 2 were given                      

I'll rebuild the DS workbench again with the updated branch and see how it goes.

and many more...

thehesiod:
I've fixed a bunch of bugs, have you tried the latest?

thehesiod:
ok @achimgaedke I've fixed the last tests

achimgaedke:
> ok @achimgaedke I've fixed the last tests

Yep, it works perfectly for me now.
Thanks a lot for all your work.

I just got aware that the increment from 2.2 to 2.3 will prevent this fix from flowing on.

```bash
pip install git+https://github.com/aio-libs/aiobotocore.git@thehesiod/awsresponse

...

s3fs 2022.3.0 requires aiobotocore~=2.2.0, but you have aiobotocore 2.3.0 which is incompatible.
Successfully installed aiobotocore-2.3.0
```

thehesiod:
I did 2.2 -> 2.3 because I fixed some of the exception conversions due to except ordering.  I'm going to merge and then if there are issues we can issue patch releases.

terrycain:
oof that did look painful, nice job though, looks great :)

thehesiod:
@terrycain thanks! ugh, just realized today that 3.6 tests weren't actually running in 3.6 due to pipenv "upgrading" the python env :(. Another pain.  Landing that shortly

PR #936, opened 2022-05-09
fix 3.6 support

thehesiod:


thehesiod:
ok biting the bullet and installing py3.6 ;)

Issue #937, opened 2022-05-09
Test files appearing in distribution package


slobberchops:
**Describe the bug**
The `tests` directory has been copied into the 2.3.2 distribution package. These files do not exist when downloading the 2.3.1 package. The existence of the test directory is likely unintentional and appears as a module when installed into a virtual environment.

This problem can be reproduced by creating a virtual environment and installing aiobotocore.

```
python3.9 -m venv aiobotocore
./aiobotocore/bin/pip install aiobotocore==2.3.2
 ls ./aiobotocore/lib/python3.9/site-packages/tests
... __init__.py  conftest.py       test_basic_s3.py  test_ec2.py	       test_mturk.py	 test_sns.py
... __pycache__  mock_server.py    test_batch.py	 test_eventstreams.py  test_patches.py	 test_sqs.py
... _helpers.py  moto_server.py    test_config.py	 test_lambda.py        test_response.py  test_version.py
... boto_tests   test_adaptive.py  test_dynamodb.py  test_monitor.py       test_session.py	 test_waiter.py
```

Installing 2.3.1 does not install the `tests` directory.

I also downloaded both 2.3.2 and 2.3.1 packages and verified that 2.3.2 contains `tests` whereas 2.3.1 does not.

**Checklist**
- [ ] I have reproduced in environment where `pip check` passes without errors
- [ ] I have provided `pip freeze` results
- [x] I have provided sample code or detailed way to reproduce
- [ ] I have tried the same code in botocore to ensure this is an aiobotocore specific issue
- [ ] I have tried similar code in aiohttp to ensure this is is an aiobotocore specific issue
- [ ] I have checked the latest and older versions of aiobotocore/aiohttp/python to see if this is a regression / injection

**pip freeze results**

**Environment:**
 - Python Version: 3.9
 - OS name and version: Darwin

**Additional context**
As an aside, installation of `tests` as a module appears to cause some confusion with IntelliJ which then has trouble when working on a project that also uses `tests` as its test directory.

thehesiod:
looks like tests folder is now gone in latest release

PR #938, opened 2022-05-09
fix Test files appearing in distribution package #937

slobberchops:
### Description of Change

Explicitly indicated `aiobotocore` package as intended target for installation.

Verified that `tests` directory does no appear in built distribution package.

  * [ ] Why needed


CHANGES.rst
```diff
@@ -1,5 +1,9 @@
 Changes
 -------
+2.3.3 (2022-05-09)
+^^^^^^^^^^^^^^^^^^
+* fix Test files appearing in distribution package
```
thehesiod, line 5: need to bump version in __init__.py as well

slobberchops: versions bumped

PR #941, opened 2022-05-20
fix connect timeout while getting IAM creds in Docker

aiudirog:
### Description of Change

While trying to use S3FS in a Docker container on my EC2, I found that `AioIMDSFetcher._fetch_metadata_token()` was throwing a `ConnectTimeoutError` trying to connect to the metadata service with `PUT http://169.254.169.254/latest/api/token/`. This came from an `aiohttp.ServerTimeoutError` while trying to submit the PUT.

 When I compared the logs to plain botocore, I found that it instead encountered a `ReadTimeoutError`, which it ignored and moved on to trying `GET http://169.254.169.254/latest/meta-data/iam/security-credentials/`, which succeeded. This error was redirected from a `urllib3.ReadTimeoutError` instead of a `ConnectTimeoutError`.

 I believe this particular difference comes from aiohttp and urllib3 handling the server connection issue differently. By handling the `ConnectTimeoutError`, aiobotocore is able to successfully get the credentials from the metadata service while in Docker.

 Note that this issue does not occur outside of Docker, even with an identical virtual env.

### Assumptions

- A `ConnectTimeoutError` from the metadata service shouldn't bubble up and end attempts to resolve credentials

  * [ ] Why needed


aiobotocore/utils.py
```diff
@@ -92,7 +92,7 @@ async def _fetch_metadata_token(self):
                         return None
                     elif response.status_code in (400,):
                         raise BadIMDSRequestError(request)
-                except ReadTimeoutError:
+                except (ReadTimeoutError, ConnectTimeoutError):
```
thehesiod, line 95: this is probably actually an issue with our exception mapping here: https://github.com/aio-libs/aiobotocore/blob/master/aiobotocore/httpsession.py#L197 mind investigating?

thehesiod: ya see: https://github.com/boto/botocore/blob/develop/botocore/utils.py#L461 I'm pretty sure we're just not mapping the exceptions correctly, we should probably port that relevant test over as well

aiudirog: Now that you mention it, I believe you're right: https://docs.aiohttp.org/en/stable/client_reference.html?highlight=servertimeouterror#aiohttp.ServerTimeoutError

The aiohttp docs say that `ServerTimeoutError` is also for read timeouts! I'll update my PR to reflect that.

---------------

Also, wow, that was the fastest response I've ever gotten to a PR! Thanks:)

aiudirog: Updated!

aiobotocore/httpsession.py
```diff
@@ -205,12 +205,10 @@ async def send(self, request):
                 request=request,
                 endpoint_url=request.url
             )
-        except ServerTimeoutError as e:
-            raise ConnectTimeoutError(endpoint_url=request.url, error=e)
```
thehesiod, line 209: we still do need to support ConnectTimeoutError: https://github.com/boto/botocore/blob/develop/botocore/httpsession.py#L478. so probably need something a little more refined

aiobotocore/utils.py
thehesiod: Welcoming a little distraction from this issue I'm working on :) added reply

aiobotocore/httpsession.py
aiudirog: Digging through the source code, it looks like we can differentiate by looking at the message:
- Connection timeout - https://github.com/aio-libs/aiohttp/blob/8e92c4c7c305f914c17f41539da20822ea7baddf/aiohttp/client.py#L505
- Read timeout - https://github.com/aio-libs/aiohttp/blob/8e92c4c7c305f914c17f41539da20822ea7baddf/aiohttp/client_proto.py#L197

How about:

```python
if str(e).lower().startswith('connect'):
    raise ConnectTimeoutError(...)
else:
    raise ReadTimeoutError(...)
```

aiudirog:
@thehesiod Are there any other changes you would like me to make?

thehesiod:
my apologies for delay, just been crazy here

CHANGES.rst
```diff
@@ -1,5 +1,9 @@
 Changes
 -------
+2.3.4 (2022-05-20)
```
thehesiod, line 3: forgot to bump version in `__init__.py`

aiudirog: Updated!

PR #974, opened 2022-10-17
bump

thehesiod:
TODO: research new token async paths like SSOTokenProvider._refresher


jakob-keller:
#999 is currently in draft and supersedes this PR.

Could you please elaborate on your TODO?

> TODO: research new token async paths like SSOTokenProvider._refresher

PR #835, opened 2020-11-03
Poetry packaging

dazza-codes:
Fix #833 although it does not use a lambci docker container in the travis-CI matrix.

This PR is a response to difficulty on a couple of other PRs when I can't grapple with pipenv for my dev-env.  There might be bits-n-pieces of this PR that are welcome and other bits that are not.  I'm open to discussion, testing and compromises.  I'm not having much success with pipenv.  Even if this PR is not adopted/merged, I will find it useful to work with this branch and rebase on master and manage a parallel set of poetry-deps as needed.

This is an attempt to use poetry packaging for the project and it might be a contentious PR/issue, depending on dev-team preferences for dev-env setups.  In using poetry for the past year or so, this PR represents some patterns that are working consistently for me.  Hopefully the added files are mostly self-explanatory, please ask any questions to clarify things and take it for a spin on your local to see how it works out.

## TODO

- update CI/CD to use poetry
- contributions require using python 3.8 now (?)
- tests might assume there is a default profile available to get credentials (?), lots of failures without it

PR #975, opened 2022-10-23
Replace setuptools with Poetry

lafrenierejm:
### Description of Change

Intended to obsolete #835.  This PR fixes tests and CI that are (as best as I can tell as of this writing) broken in #835.

### Assumptions

Compared to the existing build system using setuptools and Pipfile, Poetry makes it easier to
- Make builds fully reproducible (especially in e.g. CI) by automating the update of the lock file.
- Ensure that pinned versions of botocore, boto3, and awscli are compatible.

* [X] (NA) I have added change info to [CHANGES.rst](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.rst)
* [X] (NA) If this is resolving an issue (needed so future developers can determine if change is still necessary and under what conditions) (can be provided via link to issue with these details):
* [X] (NA) If this is providing a new feature  (can be provided via link to issue with these details):
  * [ ] Why needed

* [X] I have read and followed [CONTRIBUTING.rst](https://github.com/aio-libs/aiobotocore/blob/master/CONTRIBUTING.rst#how-to-upgrade-botocore)
* [X] I have updated test_patches.py where/if appropriate (also check if no changes necessary)
* [X] (NA) I have ensured that the awscli/boto3 versions match the updated botocore version


.github/workflows/python-package.yml
```diff
@@ -28,24 +28,20 @@ jobs:
...
-#        cache: 'pipenv'
     - name: Download dependencies
       run: |
         python -c "import sys; print(f'Python version: {sys.version}')"
-        python -m pip install -U setuptools pip pipenv codecov wheel
-        pipenv lock --python ${{ matrix.python-version }}
-        pipenv sync --dev
-        pipenv check || true
-        pipenv graph
+        curl -sSL https://install.python-poetry.org/ | python -
```
Fokko, line 34: You can also install poetry using pip on the line below 👍🏻

lafrenierejm: Implemented in https://github.com/lafrenierejm/aiobotocore/commit/caf724df187d1865c1ca8d0c49672e30489267af.

Fokko (approved):
Looks good! I'm a big fan of Poetry, but it is up to the maintainer to decide since it also impacts the release process.

lafrenierejm:
@thehesiod Any opinions on these changes?

hartungstenio:
If we are to change tools, maybe we should look at tools that support PEP 621. There are notable examples in the "PEP 517 ecosystem", like:
* [PDM](https://pdm.fming.dev/), which supports lock file and reproducible builds, and seem to have better dependency resolution than poetry
* [Hatch](https://hatch.pypa.io/), which have a great support for plugins. One good example is [hatch-vcs](https://github.com/ofek/hatch-vcs), a plugin able to use the current git tag as the `version` package metadata, and automatically update the `__version__` variable in `__init__.py`.

thehesiod:
cool thanks for the tips, worthy conversation

takeda:
> @lafrenierejm cool! we finally bumped, could you merge/re-base. Also how does this solution guarantee that each PR has a correct boto3/awscli match to the botocore version? Wouldn't you need to re-lock each time? I would think we shouldn't have a lockfile checked in in that case.

So the match is enforced via the versions provided in `pyproject.toml` if you specify exact version then `aiobotocore` will require it. Since your code is a package not an application. The lock file is not relevant when uploading the package to PyPI.

The `poetry.lock` is still useful for development, as it pins all of the lose versions specified in `pyproject.toml` to specific ones. It is essentially doing the same thing as when you run `pip-compile requirements-dev.in`

After trying to contribute a PR I think switching from `setuptools` to `poetry` would simplify development.

jakob-keller:
@thehesiod: Would you agree that this PR is stale and may be closed at this point?

lafrenierejm:
> Would you agree that this PR is stale and may be closed at this point?

I'm not @thehesiod, but this is definitely stale and should be closed. I'll go ahead and do so.

Issue #979, opened 2022-11-08
Integration with moto


yuvalshi0:
Hi maintainers,
I've been using `aiobocore` and wanted to mock s3 in my tests, I've you seen under [moto_server.py
](https://github.com/aio-libs/aiobotocore/blob/master/tests/moto_server.py) you actually run a moto service to test your services.
So to my suggesion: why not add it to moto? Or create a new package (i.e. aiobotomoto)?

thehesiod:
I think we have a request to add it as an extra, I'm game if someone comes up with PR :). Or may take me a bit to get to

PR #1004, opened 2023-04-05
Moto integration

akshara08:
### Description of Change
PR adds a patch to `endpoint.convert_to_response_dict` such that it can work with moto while not changing the expected behavior of the method. I also added tests to replicate the problem and fix the issue.

# Issue (#979)
The primary problem with moto integration is [this](https://github.com/aio-libs/aiobotocore/blob/72b8dd5d7d4ef2f1a49a0ae0c37b47e5280e2070/aiobotocore/endpoint.py#L23) method. There are 2 main issues with that method
1. It expects `raw` to have `raw_headers` just to convert the headers to lowercase. 
2. Moto passes [MockRawResponse](https://github.com/getmoto/moto/blob/79616e11e64add6b732006952abdef5c5d0914ef/moto/core/botocore_stubber.py#L9) which is not awaitable causing the breaks with lines containing `await`. 

## How to reproduce the error
```
# test_response.py

@pytest.fixture
def aws_credentials():
    """Mocked AWS Credentials for moto."""
    os.environ["AWS_ACCESS_KEY_ID"] = "testing"
    os.environ["AWS_SECRET_ACCESS_KEY"] = "testing"
    os.environ["AWS_SECURITY_TOKEN"] = "testing"
    os.environ["AWS_SESSION_TOKEN"] = "testing"


@pytest.fixture
def s3(aws_credentials):
    with mock_s3():
        conn = boto3.resource("s3")
        conn.create_bucket(Bucket='testbucket')
        yield conn


def test_moto(s3):
    for i in range(3):
        s3.Bucket('testbucket').put_object(
            Key=f'glob_{i}.txt', Body=f"test glob file {i}"
        )
    path = 's3://testbucket/glob_*.txt'
    fs = s3fs.S3FileSystem()
    files = fs.glob(path) # Fails 
    assert len(files) == 3
```

## Resolution
It's a two part solution
~~1. Remove the need for `raw_headers` and use the headers directly using `botocore.utils.lowercase_dict`~~
1. Since `raw_headers` are necessary, I've created a check to use headers if the `httpresponse` came from moto
2. Only `await` for `awaitable` objects. This check can be done using `inspect.isawaitable`

## Assumptions
~~1. This [patch](https://github.com/aio-libs/aiobotocore/blob/72b8dd5d7d4ef2f1a49a0ae0c37b47e5280e2070/aiobotocore/endpoint.py#L44) was added only to convert to lowercase and this [botocore util](https://github.com/boto/botocore/blob/29343bffa3150dfbd45ab2e5785a5b225d97f172/botocore/utils.py#L817) does the same~~
^Not valid anymore since `raw_headers` remain
1. Assigning non-awaitable objects to [response_dict['body']](https://github.com/akshara08/aiobotocore/blob/8c622779052b8a2ef90179586e1b202b1c2aa5cd/aiobotocore/endpoint.py#L49) without await is not expected to break any current behavior

## Alternative methods considered 
I considered added a raw_headers to [MockRawResponse](https://github.com/getmoto/moto/blob/79616e11e64add6b732006952abdef5c5d0914ef/moto/core/botocore_stubber.py#L9) but it seemed unnecessary especially if the point to convert to lowercase based on the [comment](https://github.com/aio-libs/aiobotocore/blob/72b8dd5d7d4ef2f1a49a0ae0c37b47e5280e2070/aiobotocore/endpoint.py#L40)

* [x] (NA) If this is providing a new feature  (can be provided via link to issue with these details):
  * [x] Why needed

# ‼️ Note ‼️ 
I've been getting `botocore.exceptions.NoCredentialsError: Unable to locate credentials` for some of the tests (5 to be exact). Ive got them even before I made changes, so I'm unsure if I'm breaking anything with this PR

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,9 @@
 Changes
 -------
+2.5.1 (2023-04-05)
+^^^^^^^^^^^^^^^^^^
+* integrate moto
+
 2.5.0 (2023-03-06)
 ^^^^^^^^^^^^^^^^^^
 * bump botocore to 1.29.76 (thanks @jakob-keller #999)
--- a/Pipfile
+++ b/Pipfile
@@ -18,7 +18,6 @@ pytest = "==6.2.4"
 pytest-cov = "==2.11.1"
 pytest-asyncio = "==0.14.0"
 pytest-xdist = "==2.2.1"
-s3fs = "==2023.3.0"
 
 # this is needed for test_patches
 dill = "==0.3.3"
--- a/Pipfile
+++ b/Pipfile
@@ -18,6 +18,7 @@ pytest = "==6.2.4"
 pytest-cov = "==2.11.1"
 pytest-asyncio = "==0.14.0"
 pytest-xdist = "==2.2.1"
+s3fs = "==2023.3.0"
 
 # this is needed for test_patches
 dill = "==0.3.3"
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -15,6 +15,7 @@ from botocore.endpoint import (
 )
 from botocore.hooks import first_non_none_response
 from botocore.utils import lowercase_dict
+from moto.core.botocore_stubber import MockRawResponse
 from urllib3.response import HTTPHeaderDict
 
 from aiobotocore.httpchecksum import handle_checksum_body
@@ -39,12 +40,28 @@ async def convert_to_response_dict(http_response, operation_model):
 
     """
     response_dict = {
-        'headers': HTTPHeaderDict(lowercase_dict(http_response.headers)),
         'status_code': http_response.status_code,
         'context': {
             'operation_name': operation_model.name,
         },
     }
+
+    if isinstance(http_response.raw, MockRawResponse):
+        # patch to integrate with moto
+        response_dict['headers'] = HTTPHeaderDict(
+            lowercase_dict(http_response.headers)
+        )
+    else:
+        # botocore converts keys to str, so make sure that they are in
+        # the expected case. See detailed discussion here:
+        # https://github.com/aio-libs/aiobotocore/pull/116
+        # aiohttp's CIMultiDict camel cases the headers :(
+        response_dict['headers'] = HTTPHeaderDict(
+            {
+                k.decode('utf-8').lower(): v.decode('utf-8')
+                for k, v in http_response.raw.raw_headers
+            }
+        )
     if response_dict['status_code'] >= 300:
         if isawaitable(http_response.content):
             response_dict['body'] = await http_response.content
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -1,4 +1,5 @@
 import asyncio
+from inspect import isawaitable
 
 from botocore.endpoint import (
     DEFAULT_TIMEOUT,
@@ -13,6 +14,7 @@ from botocore.endpoint import (
     logger,
 )
 from botocore.hooks import first_non_none_response
+from botocore.utils import lowercase_dict
 from urllib3.response import HTTPHeaderDict
 
 from aiobotocore.httpchecksum import handle_checksum_body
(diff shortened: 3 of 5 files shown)


Pipfile
```diff
@@ -18,6 +18,7 @@ pytest = "==6.2.4"
 pytest-cov = "==2.11.1"
 pytest-asyncio = "==0.14.0"
 pytest-xdist = "==2.2.1"
+s3fs = "==2023.3.0"
```
thehesiod, line 21: this introduces a circular dependency, can't do it as it may later be incompatible with the version of aiobotocore this becomes

aiobotocore/endpoint.py
```diff
@@ -37,30 +39,27 @@ async def convert_to_response_dict(http_response, operation_model):
...
     """
     response_dict = {
-        # botocore converts keys to str, so make sure that they are in
-        # the expected case. See detailed discussion here:
-        # https://github.com/aio-libs/aiobotocore/pull/116
-        # aiohttp's CIMultiDict camel cases the headers :(
-        'headers': HTTPHeaderDict(
-            {
-                k.decode('utf-8').lower(): v.decode('utf-8')
-                for k, v in http_response.raw.raw_headers
```
thehesiod, line 47: it's important we use raw_headers as the non-raw doesn't preserve casing of the value

tests/test_response.py
```diff
@@ -187,3 +192,44 @@ async def test_streaming_line_empty_body():
...
     await assert_lines(stream.iter_lines(), [])
+
+
+@pytest.fixture
+def aws_credentials():
+    """Mocked AWS Credentials for moto."""
+    os.environ["AWS_ACCESS_KEY_ID"] = "testing"
+    os.environ["AWS_SECRET_ACCESS_KEY"] = "testing"
+    os.environ["AWS_SECURITY_TOKEN"] = "testing"
+    os.environ["AWS_SESSION_TOKEN"] = "testing"
```
thehesiod, line 203: should be moved to conftest.py

aiobotocore/endpoint.py
akshara08: Alright, the alternative is kind of an ugly patch. Let me know what ya think

Pipfile
akshara08: Makes sense. Used `paginator` for testing

akshara08:
@thehesiod any update on this?

thehesiod:
re NOTE in your description you need to run with `-m moto` to limit to moto marked tests, all other tests are only to be run with private key as moto didn't have support for them when tests were originally written (we should revisit these tests and mark what exactly is missing for the test).

CHANGES.rst
```diff
@@ -1,5 +1,9 @@
 Changes
 -------
+2.5.1 (2023-04-05)
+^^^^^^^^^^^^^^^^^^
+* integrate moto
```
thehesiod, line 5: this is more of add support for moto in-place mock methods

aiobotocore/endpoint.py
```diff
@@ -13,6 +14,8 @@
     logger,
 )
 from botocore.hooks import first_non_none_response
+from botocore.utils import lowercase_dict
+from moto.core.botocore_stubber import MockRawResponse
```
thehesiod, line 18: moto is not a dependency of this library, so you can't do this.  I think this is really a bug in moto, it needs support for aiobotocore.  If you were to do this it would need to be a try/except ImportError.  Another option would be for aiobotocore to expose some moto helpers that correctly patch the response.  I'd do that instead as it will be a lot cleaner.

thehesiod:
sorry for delay, I think either moto needs to add async patches, or we ship like a moto_helpers.py which does the right thing.

aiobotocore/endpoint.py
akshara08: > moto is not a dependency of this library, so you can't do this. 
Ack! I missed that! 
I'm a little busy this week but I'll try to sneak in some time to dig into moto repo and see if there's a cleaner solution to this. I tried sending encoded `raw_headers` [here](https://github.com/akshara08/moto/blob/bbb07b4e7ede832d69b8fe0089e0707c915710fc/moto/core/botocore_stubber.py#L69) the last time I tried and it broke a bunch of tests. Let me try that again and see

thehesiod: no worries, thanks for working on this

akshara08:
Closing this to avoid piling on irrelevant commits. See #1007

PR #1008, opened 2023-04-20
Enter client CM in AioSSOTokenProvider

nickcollier:
### Description of Change
The client CM is never entered in AioSSOTokenProvider and when a token refresh is attempted create_token is called on the ClientCreatorContext rather than the actual initialized client

`AttributeError: 'ClientCreatorContext' object has no attribute 'create_token'`

### DIFF
--- a/aiobotocore/tokens.py
+++ b/aiobotocore/tokens.py
@@ -83,12 +83,13 @@ class AioDeferredRefreshableToken(DeferredRefreshableToken):
 
 class AioSSOTokenProvider(SSOTokenProvider):
     async def _attempt_create_token(self, token):
-        response = await self._client.create_token(
-            grantType=self._GRANT_TYPE,
-            clientId=token["clientId"],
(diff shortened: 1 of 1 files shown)


thehesiod:
thanks! hmm, now I'm wonder if the other places where we use self._client in these overridden classes

thehesiod:
I think these may be wrong too: https://github.com/aio-libs/aiobotocore/blob/72b8dd5d7d4ef2f1a49a0ae0c37b47e5280e2070/aiobotocore/utils.py#L483 and https://github.com/aio-libs/aiobotocore/blob/72b8dd5d7d4ef2f1a49a0ae0c37b47e5280e2070/aiobotocore/utils.py#L598

thehesiod:
ya, they're wrong I'm almost positive, could you add those to this pr too please, thanks!

thehesiod:
closing as I added the other two fixes in https://github.com/aio-libs/aiobotocore/pull/1030

PR #1011, opened 2023-04-25
add waiter.wait return

thehesiod:
hopefully botocore change coming from https://github.com/boto/botocore/issues/2913


thehesiod:
going to fix this in another pull request

PR #1033, opened 2023-08-11
bump minimum aiohttp req and add aiohttp tests

thehesiod:


tests/test_patches.py
```diff
@@ -115,16 +112,6 @@
...
 #    to the set
 
-# The follow is for our monkeypatches for read_timeout:
-#    github.com/aio-libs/aiobotocore/pull/248
-_AIOHTTP_DIGESTS = {
-    # for using _body
-    ClientResponse: {
-        'e178726065b609c69a1c02e8bb78f22efce90792',
-        '225e8033bfcff8cccbc2e975d7bd0c7993f14366',
-    },
```
thehesiod, line 125: is no longer needed as we don't reach into the private members anymore

PR #1034, opened 2023-08-16
Add support for Python 3.12 and drop EOL 3.7

hugovk:
### Description of Change

The [Python 3.12 release candidate is out](https://discuss.python.org/t/python-3-12-0-release-candidate-1-released/31137?u=hugovk) 🚀 

> ## Call to action
> 
> We strongly encourage maintainers of third-party Python projects to prepare their projects for 3.12 compatibilities during this phase, and where necessary publish Python 3.12 wheels on PyPI to be ready for the final release of 3.12.0.

Also EOL Python 3.7 was dropped in https://github.com/aio-libs/aiobotocore/pull/1033, this does some more cleanup.

### Assumptions

I assume you want to be ready for Python 3.12 :) 

* [ ] I have added change info to [CHANGES.rst](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.rst) - not done because tests would fail due to version mismatch between `CHANGES.rst` and `__init__.py`
  * [ ] Why needed


CHANGES.rst
```diff
@@ -1,5 +1,6 @@
 Changes
 -------
+
```
thehesiod, line 3: can you add a new line for the next release with this, should probably bump to 2.7.0 since we're removing 3.7 support

hugovk: Done! I didn't know what date to put, so used today as an arbitrary date.

Issue #939, opened 2022-05-13
upgrade from 2.2.X to 2.3.X breaks tests using botocore.stub.Stubber


akursar:
**Describe the bug**
Upgrading from 2.2.X to 2.3.X causes tests that use Stubber to fail. I have this example test, which is an aiobotocore version of the example described at https://botocore.amazonaws.com/v1/documentation/api/latest/reference/stubber.html#botocore.stub.Stubber

The test passes if I downgrade to 2.2.0, but fails with 2.3.0 (and 2.3.2).

I'm not sure if aiobotocore seeks to be compatible with Stubber, but it seemed worthwhile to report the issue.

```
import datetime

import pytest
from aiobotocore.session import get_session
from botocore.stub import Stubber


@pytest.mark.asyncio
async def test_example():
    session = get_session()
    async with session.create_client('s3') as s3:
        stubber = Stubber(s3)

        response = {
            'IsTruncated': False,
            'Name': 'test-bucket',
            'MaxKeys': 1000, 'Prefix': '',
            'Contents': [{
                'Key': 'test.txt',
                'ETag': '"abc123"',
                'StorageClass': 'STANDARD',
                'LastModified': datetime.datetime(2016, 1, 20, 22, 9),
                'Owner': {'ID': 'abc123', 'DisplayName': 'myname'},
                'Size': 14814
            }],
            'EncodingType': 'url',
            'ResponseMetadata': {
                'RequestId': 'abc123',
                'HTTPStatusCode': 200,
                'HostId': 'abc123'
            },
            'Marker': ''
        }

        expected_params = {'Bucket': 'test-bucket'}

        stubber.add_response('list_objects', response, expected_params)
        stubber.activate()

        service_response = await s3.list_objects(Bucket='test-bucket')
        assert service_response == response
```

It passes for 2.2.0 but fails for 2.3.0 with

```
aiohttp.client_exceptions.ServerTimeoutError: Connection timeout to host http://169.254.169.254/latest/api/token
```

**Checklist**
- [X] I have reproduced in environment where `pip check` passes without errors
- [X] I have provided `pip freeze` results
- [X] I have provided sample code or detailed way to reproduce
- [X] I have tried the same code in botocore to ensure this is an aiobotocore specific issue
- [X] I have tried similar code in aiohttp to ensure this is is an aiobotocore specific issue

sort of. i am not changing the version of aiohttp but am only updating aiobotocore

[...]

PR #1039, opened 2023-09-07
AioStubber that returns AioAWSResponse()

takeda:
### Description of Change
Adding AioStubber to provide async version (responding with AioAWSResponse()) of Stubber.

Hopefully solving issue raised in https://github.com/aio-libs/aiobotocore/issues/939

  * [ ] Why needed


thehesiod (commented):
need to add entry to test_patches

takeda:
@thehesiod updated

takeda:
@thehesiod I just realized that the CI tasks expect version to be provided, I did not do it because I thought that will be done after the PR is merged. I just updated the PR to include a version.

takeda:
@thehesiod please run it again

takeda:
@thehesiod I fixed the formatting and flake8 now doesn't report issues.

takeda:
*sigh* I wonder what will fail this time

takeda:
@thehesiod sorry for delay, but I had other things I needed to finish. I added some unit tests now.

aiobotocore/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '2.7.0'
+__version__ = '2.8.0'
```
thehesiod, line 1: I think we can just do a dot release, no need for major rev

takeda: @thehesiod updated to 2.7.1 also I don't know what's with coverity, because it was pointing that some parts were not covered while they were. I also updated the unit tests so I'm getting 100% for that file now. Can you run the pipeline again?

takeda:
@thehesiod can you run the pipeline again?

takeda:
@thehesiod I rebased the PR due to another PR that was merged recently

thehesiod:
whoops, forgot about this pr, thanks

takeda:
@thehesiod I see it failed, but it looks like it was during uploading the results to codecov?

takeda:
The codecov still makes nos sense, coverage told me the coverage is 100%, looking at codecov I see for example this (line 15):

https://github.com/aio-libs/aiobotocore/pull/1039/files#annotation_15037531067

But that line is always executed, so I don't see how it wouldn't be covered.

thehesiod:
> The codecov still makes nos sense, coverage told me the coverage is 100%, looking at codecov I see for example this (line 15):
> 
> https://github.com/aio-libs/aiobotocore/pull/1039/files#annotation_15037531067
> 
> But that line is always executed, so I don't see how it wouldn't be covered.

you need to add the moto marker to your tests, we only run things marked with [moto](https://github.com/aio-libs/aiobotocore/blob/master/Makefile#L25).  what would be better going forward is instead having a marker for tests that require a real AWS account, and then exclude those if you want to attempt that.  This has bit us a couple times with new contributors.

thehesiod:
looks like you can do this with `pytest -m "not MARKER"`

takeda:
@thehesiod I added the marker, I didn't use it, since stubber overlaps with functionality so I felt it was redundant.

PR #1041, opened 2023-09-19
Fix test_load_sso_credentials_without_cache

jakkdl:
### Description of Change
This test appears to have been completely broken since it was written in 2020, but it wasn't marked with any fixtures so it got ignored in CI and you only notice it if running all tests or you look at codecov for the lines in question: https://app.codecov.io/gh/aio-libs/aiobotocore/blob/master/tests%2Fboto_tests%2Ftest_credentials.py#L1256
I simply copied the relevant code from the test below it and it started working, so I assume this is what it was meant to be.

### Assumptions
That this test was intended to ... do something, anything.

* [ ] I have added change info to [CHANGES.rst](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.rst) - [I don't think this is required?]
* [x] One of several otherwise unrelated issues encountered in #965

### DIFF
--- a/tests/boto_tests/test_credentials.py
+++ b/tests/boto_tests/test_credentials.py
@@ -1253,10 +1253,14 @@ def _add_get_role_credentials_response(self):
     )
 
 
-def test_load_sso_credentials_without_cache(self):
-    self._add_get_role_credentials_response()
+@pytest.mark.moto
+@pytest.mark.asyncio
+async def test_load_sso_credentials_without_cache(sso_provider_setup):
+    self = sso_provider_setup
+    _add_get_role_credentials_response(self)
     with self.stubber:
-        credentials = self.provider.load()
+        credentials = await self.provider.load()
+        credentials = await credentials.get_frozen_credentials()
         self.assertEqual(credentials.access_key, 'foo')
         self.assertEqual(credentials.secret_key, 'bar')
         self.assertEqual(credentials.token, 'baz')


jakkdl:
The failing test seems unrelated to code changes - we get a warning about `path_to_write_report` [which was removed in 3.1.2](https://github.com/codecov/codecov-action/blob/3f71d3b2559351e4b67bb566281cc055ee038d9a/CHANGELOG.md?plain=1#L37), but that's not new and not the cause of the error.

But codecov uploading seemed to work in #1039, which was also opened from a fork (so it's not about secrets not being available in forks or something), so I'm not sure what might be different. Me being a first-time contributor to the repo?? Random fail??

jakkdl:
> re-ran, looks like error is `ImportError: cannot import name '_legacy_validators' from 'jsonschema'`

This seems to be jsonschema having renamed that in 4.19.1 (without mentioning in the changelog): https://github.com/python-jsonschema/jsonschema/compare/v4.19.0...v4.19.1#diff-1622fcab42cac2f075bbe401c3a9bd221ffb7757a029ccd207871bd12fd522b7

But that has been fixed in openapi-schema-validator 0.6.1 https://github.com/python-openapi/openapi-schema-validator/releases/tag/0.6.1 that was released two days ago, so rerunning now will fix that.


But that is all in the `Run unittests` step, whereas the first run failed in the `Upload to Codecov` step, so the original problem might still be there. But rerun and we will see

jakkdl:
:tada:

thehesiod (approved):

thehesiod:
ty!

Issue #1061, opened 2023-12-04
Aiohttp connection leak


grigoriev-semyon:
**Describe the bug**
When I use `get_object` my code starts to run very slowly. Many open aiohttp connections begin to be created. These connections cannot be released. The length of the aiohttp.BaseConnector._acquired set grows to the size of the connection pool and does not fall back.

**Checklist**
- [x] I have reproduced in environment where `pip check` passes without errors
- [x] I have provided `pip freeze` results
- [x] I have provided sample code or detailed way to reproduce
- [x] I have tried the same code in botocore to ensure this is an aiobotocore specific issue
- [x] I have tried similar code in aiohttp to ensure this is is an aiobotocore specific issue
- [x] I have checked the latest and older versions of aiobotocore/aiohttp/python to see if this is a regression / injection

**pip freeze results**
```

aiosignal==1.3.1
async-timeout==4.0.2
attrs==22.2.0
charset-normalizer==2.1.1
frozenlist==1.3.3
idna==3.4
multidict==6.0.4
yarl==1.8.2
aiohttp==3.8.4

certifi==2023.5.7

async-timeout==4.0.2
redis==5.0.1

PyYAML==6.0.1

flit_core==3.8.0
packaging==23.0
marshmallow==3.19.0
webargs==8.2.0

psycopg
asyncpg
sqlalchemy[asyncio]

aiobotocore==2.8.0

aiofiles
```

**Environment:**
 - Python Version: 3.11.4
 - OS name and version: CentOS

Code(connection):

```python
from contextlib import AsyncExitStack

from aiobotocore import session
from aiobotocore.config import AioConfig


class S3:
    def __init__(
        self,
        access_key_id: str,
        secret_access_key: str,
        endpoint_url: str,
        bucket: str,
    ) -> None:
        self._access_key_id = access_key_id
        self._secret_access_key = secret_access_key
        self._endpoint_url = endpoint_url
        self._bucket = bucket
        self._context_stack = AsyncExitStack()
        self._session = session.get_session()
        self.config = AioConfig(max_pool_connections=0)
        # self.config.connector_args = {"keepalive_timeout": None, "force_close": True}  

    @property
    def bucket(self) -> str:
        return self._bucket
  
    @property
    def client(self):
        return self._client


    async def action_connect(self, *args, **kwargs) -> None:
        self._client = await self._context_stack.enter_async_context(
            self._session.create_client(
                's3',
                aws_access_key_id=self._access_key_id,
[...]

PR #1062, opened 2023-12-05
New example of `get_object`

grigoriev-semyon:
### Description of Change
I've added a description of how to use the get_object method when using AsyncExitStack.

Fixes: #1061 

* [x] I have added change info to [CHANGES.rst](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.rst) (i dont think this is required when update exmaples)
  * [x] Why needed

### DIFF
--- a/examples/simple.py
+++ b/examples/simple.py
@@ -28,20 +28,14 @@ async def go():
         resp = await client.get_object_acl(Bucket=bucket, Key=key)
         print(resp)
         
-        try:
-            resp = await client.get_object(Bucket=bucket, Key=key)
+        resp = await client.get_object(Bucket=bucket, Key=key)
+        async with resp['Body'] as stream:
+            await stream.read()  # if you do not read the stream the connection cannot be re-used and will be dropped
             print(resp)
-        finally:
-            """
-            Required step if you use AsyncContextStack way.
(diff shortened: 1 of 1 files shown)


examples/simple.py
```diff
@@ -27,6 +27,22 @@ async def go():
...
+            """
+            Required step if you use AsyncContextStack way.
+            https://github.com/aio-libs/aiobotocore#context-manager-examples
+            
+            If you don't do this, the connections will not be closed
+            and eventually the pool of available connections will be empty
+            after a few requests.
+            All connections in aiohttp connection pool will be open.
+            """
+            resp["Body"].close() ##  Closes the aiohttp unreleased connection.
```
thehesiod, line 44: better way and updated comment:

```suggestion
        resp = await client.get_object(Bucket=bucket, Key=key)
        async with resp['Body'] as stream:
            await stream.read()  # if you do not read the stream the connection cannot be re-used and will be dropped
            print(resp)
        """
        This is to ensure the connection is returned to the pool as soon as possible.
        Otherwise the connection will be released after it is GC'd
        """
```

grigoriev-semyon:
thx to @SirGrigory

thehesiod:
ty!

PR #1073, opened 2024-01-16
Fix race condition in S3 Express identity cache

jakob-keller:
### Description of Change
A race condition in the S3 Express identity cache was introduced in 110b0a2a. This change modifies the cache to hold `asyncio.Task` instances instead of actual credentials to mitigate the issue without introducing locking.

### Assumptions
*Replace this text with any assumptions made (if any)*

### Checklist for All Submissions
* [x] I have added change info to [CHANGES.rst](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.rst)
* [x] If this is resolving an issue (needed so future developers can determine if change is still necessary and under what conditions) (can be provided via link to issue with these details): closes #1072
  * [ ] Detailed description of issue
  * [x] Alternative methods considered (if any): This race condition might also be [resolved through locking](https://github.com/aio-libs/aiobotocore/issues/1072#issuecomment-1894342999). This would result in additional complexity and runtime overhead.
  * [x] How issue is being resolved: The cache holds `asyncio.Task` instances that wrap the underlying `get_credentials` coroutine.
  * [ ] How issue can be reproduced
* [ ] If this is providing a new feature  (can be provided via link to issue with these details):
  * [ ] Detailed description of new feature
  * [ ] Why needed
  * [ ] Alternatives methods considered (if any)

### Checklist when updating botocore and/or aiohttp versions

* [x] I have read and followed [CONTRIBUTING.rst](https://github.com/aio-libs/aiobotocore/blob/master/CONTRIBUTING.rst#how-to-upgrade-botocore)
* [x] I have updated test_patches.py where/if appropriate (also check if no changes necessary)
* [x] I have ensured that the awscli/boto3 versions match the updated botocore version

Issue #1072, opened 2024-01-15
S3 Express Session opened for all asyncio calls


LouisAuneau:
Hello,

**Describe the bug**
We are aiming at improving our uploading speed using S3 Express One Zone. Running the following code using botocore, I manage to upload files in a reasonable time:
```
import botocore.session

bucket = 'my-bucket-eun1-az1--x-s3'
region = 'eu-north-1'
session = botocore.session.get_session()
client = session.create_client('s3', region_name=region)

for i, file in enumerate(dataset):
    client.put_object(
        Body=file,
        Bucket=bucket,
        Key=f'file_{i}.jpg'
    )
```

However, I wanted to improve perfomance by running async with the following code:
```
import aiobotocore.session

bucket = 'my-bucket-eun1-az1--x-s3'
region = 'eu-north-1'
session = aiobotocore.session.get_session()

async with session.create_client('s3', region_name=region) as client:
    tasks = []
    for i, file in enumerate(dataset):
        tasks.append(client.put_object(
            Body=file,
            Bucket=bucket,
            Key=f'file_{i}.jpg'
        ))

    await asyncio.gather(*tasks)
```

But it seems when running multiple `put_object` call asynchronously, the `CreateSession` endpoint is called at each call, leading to the following `ClientError`:
```
ClientError: An error occurred (SlowDown) when calling the CreateSession operation (reached max retries: 4): Reduce your request rate.
```

**Checklist**
- [x] I have reproduced in environment where `pip check` passes without errors
- [x] I have provided `pip freeze` results
- [x] I have provided sample code or detailed way to reproduce
- [x] I have tried the same code in botocore to ensure this is an aiobotocore specific issue
- [ ] I have tried similar code in aiohttp to ensure this is is an aiobotocore specific issue -> Time consuming as it requires to re-implement the authentication system of AWS.
- [x] I have checked the latest and older versions of aiobotocore/aiohttp/python to see if this is a regression / injection -> First and latest version that supports S3 Express One Zone.

**pip freeze results**
```
absl-py==1.4.0
aiobotocore==2.9.0
aiofile==3.8.8
aiofiles==23.2.1
aiohttp==3.9.1
aiohttp-retry==2.8.3
aioitertools==0.11.0
aiopath==0.5.12
aiosignal==1.3.1
anyio==3.7.1
asn1crypto==1.5.1
asttokens==2.4.1
async-timeout==4.0.3
asyncpg==0.29.0
attrs==23.2.0
[...]

PR #1074, opened 2024-01-17
update venv scripts

thehesiod:
ensures compiled requirements file always match sources used to generate it

### DIFF
--- a/aiobotocore/utils.py
+++ b/aiobotocore/utils.py
@@ -1,8 +1,9 @@
 import asyncio
-import functools
 import inspect
 import json
 import logging
+
+from async_lru import alru_cache
 
 import botocore.awsrequest
 from botocore.exceptions import (
@@ -367,30 +368,9 @@ class AioIdentityCache(IdentityCache):
 
 
 class AioS3ExpressIdentityCache(AioIdentityCache, S3ExpressIdentityCache):
-    @functools.cached_property
-    def _aio_credential_cache(self):
-        """Substitutes upstream credential cache."""
-        return {}
-
+    @alru_cache(maxsize=100)
     async def get_credentials(self, bucket):
-        # upstream uses `@functools.lru_cache(maxsize=100)` to cache credentials.
-        # This is incompatible with async code.
-        # We need to implement custom caching logic.
-
-        if (credentials := self._aio_credential_cache.get(bucket)) is None:
-            # cache miss -> get credentials asynchronously
-            credentials = await super().get_credentials(bucket=bucket)
-
-            # upstream cache is bounded at 100 entries
-            if len(self._aio_credential_cache) >= 100:
-                # drop oldest entry from cache (deviates from lru_cache logic)
-                self._aio_credential_cache.pop(
-                    next(iter(self._aio_credential_cache)),
-                )
-
-            self._aio_credential_cache[bucket] = credentials
-
-        return credentials
+        return await super().get_credentials(bucket=bucket)
 
     def build_refresh_callback(self, bucket):
         async def refresher():
--- a/aiobotocore/utils.py
+++ b/aiobotocore/utils.py
@@ -1,9 +1,8 @@
 import asyncio
+import functools
 import inspect
 import json
 import logging
-
-from async_lru import alru_cache
 
 import botocore.awsrequest
 from botocore.exceptions import (
@@ -368,9 +367,30 @@ class AioIdentityCache(IdentityCache):
 
 
 class AioS3ExpressIdentityCache(AioIdentityCache, S3ExpressIdentityCache):
-    @alru_cache(maxsize=100)
+    @functools.cached_property
+    def _aio_credential_cache(self):
+        """Substitutes upstream credential cache."""
+        return {}
+
     async def get_credentials(self, bucket):
-        return await super().get_credentials(bucket=bucket)
+        # upstream uses `@functools.lru_cache(maxsize=100)` to cache credentials.
+        # This is incompatible with async code.
+        # We need to implement custom caching logic.
+
+        if (credentials := self._aio_credential_cache.get(bucket)) is None:
+            # cache miss -> get credentials asynchronously
+            credentials = await super().get_credentials(bucket=bucket)
+
+            # upstream cache is bounded at 100 entries
+            if len(self._aio_credential_cache) >= 100:
+                # drop oldest entry from cache (deviates from lru_cache logic)
+                self._aio_credential_cache.pop(
+                    next(iter(self._aio_credential_cache)),
+                )
+
+            self._aio_credential_cache[bucket] = credentials
+
+        return credentials
 
     def build_refresh_callback(self, bucket):
         async def refresher():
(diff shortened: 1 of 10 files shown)


thehesiod:
@jakob-keller thoughts?  this uses another aio-libs module which seems to do what we want

build_and_sync_requirements.sh
thehesiod, line 1: should make local debugging less painful

thehesiod:
for some reason I can't add you as a reviewer

jakob-keller:
> for some reason I can't add you as a reviewer

I guess that's because I'm not a collaborator.

jakob-keller:
> @jakob-keller thoughts? this uses another aio-libs module which seems to do what we want

Good idea. I did not know about that library.

Do we understand the performance implications? AFAIK, S3 Express is all about lower latencies. It would be a shame, if that cache added more overhead than necessary. A benchmark would be great.

As a user, I would be slightly concerned about adding another dependency: Fewer is better, IMO. In particular, if there is another fairly straightforward solution available as #1073.

jakob-keller:
> Do we understand the performance implications? AFAIK, S3 Express is all about lower latencies. It would be a shame, if that cache added more overhead than necessary. A benchmark would be great.

Here's a simple benchmark.py:

```Python
import asyncio
import functools

import async_lru


async def main(acache, n: int = 1000) -> None:
    """Async main coroutine."""
    await asyncio.gather(*(acache("test") for _ in range(n)))


async def noop(_arg: str) -> None:
    """Trivial noop cooroutine."""


@functools.lru_cache(maxsize=100)
def _cache_1073(_arg: str) -> asyncio.Task[None]:
    return asyncio.create_task(noop(_arg))


async def cache_1073(_arg: str) -> None:
    """functools.lru_cache wraps asyncio.Tasks - #1073."""
    return await _cache_1073(_arg)


@async_lru.alru_cache(maxsize=100)
async def cache_1074(_arg: str) -> None:
    """async-lru based - #1074."""
    return await noop(_arg)
```

Run with

`python -m timeit -s "import asyncio; from benchmark import main, cache_1073" "asyncio.run(main(cache_1073))"`

vs.

`python -m timeit -s "import asyncio; from benchmark import main, cache_1074" "asyncio.run(main(cache_1074))"`

Under CPython 3.11.7 on macOS, #1073 turns out to be slightly faster than #1074: ~10%. That is not significant IMO and should not determine the decision.

thehesiod:
@jakob-keller I don't think #1073 is correct, it's avoiding a needed lock to ensure each lru entry is only awaited once

thehesiod:
IMO any extra effort should be put into speeding up/correcting https://github.com/aio-libs/async-lru. Looking at it I think there are many potential options. external deps aren't great indeed, but at least it's from the same ecosystem (aiolibs)

jakob-keller:
> @jakob-keller I don't think #1073 is correct, it's avoiding a needed lock to ensure each lru entry is only awaited once

I'm still convinced that both approaches are valid fixes for #1072: see https://github.com/aio-libs/aiobotocore/pull/1073#discussion_r1456336296

aiobotocore/utils.py
jakob-keller, line 1: I like the fact that this code looks almost the same as upstream.

install-requires.txt
jakob-keller, line 1: I'm in favor of aligning our "piping" to that of upstream as much as possible: less divergence should result in less trouble down the road.

In this case, botocore inlines its requirements: https://github.com/boto/botocore/blob/1.31.68/setup.py

What is the benefit of spinning this out?

setup.py
```diff
@@ -1,38 +1,32 @@
...
-install_requires = [
-    # pegged to also match items in `extras_require`
-    'botocore>=1.33.2,<1.33.14',
-    'aiohttp>=3.7.4.post0,<4.0.0',
-    'wrapt>=1.10.10, <2.0.0',
-    'aioitertools>=0.5.1,<1.0.0',
-]
+_this_dir = Path(__file__).parent
 
+# NOTE: When updating botocore make sure to update install-requires.txt
```
jakob-keller, line 7: We would need to also update documentation elsewhere.

Relying on simultaneous updates to multiple files is also more error prone IMO.

setup.py
```diff
@@ -1,38 +1,32 @@
...
 extras_require = {
     'awscli': ['awscli>=1.31.2,<1.31.14'],
     'boto3': ['boto3>=1.33.2,<1.33.14'],
 }
 
 
-def read(f):
-    return open(os.path.join(os.path.dirname(__file__), f)).read().strip()
+def read(file_name: str):
+    return (_this_dir / file_name).read_text().strip()
```
jakob-keller, line 15: `pathlib` sure is more idiomatic. And I love type hints :-) Why not annotate the return type (`-> str`), while we're at it?

jakob-keller (changes requested):
Thank you for the PR and all its contents. I have added a few comments.

IMO, the "piping" related changes should be in another PR. Would you mind splitting the PR up?

jakob-keller:
> > for some reason I can't add you as a reviewer
> 
> I guess that's because I'm not a collaborator.

Now I am. Thank you! 🎉

install-requires.txt
thehesiod: the benefit is that we can create a lock file based on the hash of the sources, so we know that the lockfile matches the source.  I suppose we can just hash the whole setup.py.  I can swap to that

aiobotocore/utils.py
thehesiod: going to revert in favor of your impl based on my updated understanding of how it works

thehesiod:
thanks for all the investigative work and instructiveness!

thehesiod:
@jakob-keller ok re-engineered based on discussions

build_and_sync_requirements.sh
```diff
@@ -0,0 +1,30 @@
...
+}
+
+trap cleanup EXIT SIGINT SIGTERM
+
+scratch_file=$(mktemp)
+
+envsubst < requirements-dev.in > ${scratch_file}
+
+# also hash setup.py as it has the install_requires and extras
+SHA_SUM=$(sha1sum ${scratch_file} setup.py | head -c 40)
```
jakob-keller, line 23: sha1sum produces two lines with two items each, which are piped to head which takes the first 40 bytes only. Looking at CI runs, this appears to be fragile and it completely ignores the hash of setup.py.

May I propose this alternative?

`cat ${scratch_file} setup.py | sha1sum | awk '{print $1}'`

1. Concatenate both inputs
2. Pipe into sha1sum
3. Extract first word (the 40 character hash)

thehesiod: 🤦🏼‍♂️ thanks, that is a very odd default behavior of sha1sum.  I'll re-work

setup.py
thehesiod: reverted file

setup.py
thehesiod: reverted file

install-requires.txt
thehesiod: reverted file

aiobotocore/utils.py
thehesiod: reverted file

PR #1077, opened 2024-01-20
Fix use of HTTPS proxies #1070

### LINKED ISSUE #1070: Proxy not working due to signature mismatch
**Description**

The method [`_setup_proxy_ssl_context`](https://github.com/aio-libs/aiobotocore/blob/110b0a2a8f18a8546e3a1a32f71810387a7d46a4/aiobotocore/httpsession.py#L144) expects a string but is passed a dict [here](https://github.com/aio-libs/aiobotocore/blob/110b0a2a8f18a8546e3a1a32f71810387a7d46a4/aiobotocore/httpsession.py#L100). At the `parse_url()` call, the connection establishment will fail as the dict cannot be parsed as url and does not even contain the url (which is why I can't think of an easy fix right now).

In commit [fbfeb1242ab1f378f6d4fe84b5d3dacdd8874189](https://github.com/aio-libs/aiobotocore/commit/fbfeb1242ab1f378f6d4fe84b5d3dacdd8874189), the signature was changed but the call was not updated. Reverting the changes applied to `_setup_proxy_ssl_context` does solve the problem.

In botocore, `_setup_proxy_ssl_context` is called in the `send` method where the request is available.

**Checklist**
- [ ] I have provided sample code or detailed way to reproduce
- [x] I have ~~tried~~ reviewed the same code in botocore to ensure this is an aiobotocore specific issue
- [x] I have ~~tried~~ reviewed similar code in aiohttp to ensure this is is an aiobotocore specific issue

**pip freeze results**
```
aiobotocore==2.9.0
aiohttp==3.9.1
aioitertools==0.11.0
aiosignal==1.3.1
attrs==23.2.0
botocore==1.33.13
certifi==2023.11.17
frozenlist==1.4.1
fsspec==2023.12.2
idna==3.6
jmespath==1.0.1
multidict==6.0.4
python-dateutil==2.8.2
s3fs==2023.12.2
six==1.16.0
urllib3==2.0.7
wrapt==1.16.0
yarl==1.9.4
```

**Environment:**
 - Python Version: 3.11.4
 - OS name and version: darwin 14.2.1 (23C71)

jakob-keller:
### Description of Change
This PR fixes support for HTTPS proxies.

### Assumptions
Connection to the HTTPS proxy and to the HTTPS endpoint may use a single `ssl.SSLContext`.

* [x] If this is resolving an issue (needed so future developers can determine if change is still necessary and under what conditions) (can be provided via link to issue with these details): closes #1070 
  * [x] How issue is being resolved: Use separate aiohttp sessions per proxy URL and merge any relevant configuration options into a single `ssl.SSLContext` to be used both for connections to the HTTPS proxy as well as the HTTPS endpoint
  * [ ] Why needed


aiobotocore/httpsession.py
```diff
@@ -93,25 +92,42 @@ def __init__(
...
         # request so don't need proxy manager
 
-        ssl_context = None
+        context = None
         if bool(verify):
+            context = self._get_ssl_context()
+            if self._cert_file:
+                context.load_cert_chain(self._cert_file, self._key_file)
+
+            # inline self._setup_ssl_cert
```
thehesiod, line 101: I'd prefer we kept this logic in `_setup_proxy_ssl_context` as it'll be easier to diff going forwards.  There's a flaw in this approach though because the proxy context depends on if the host we're connecting to is an IPV4 address or not: https://github.com/boto/botocore/blob/1.34.23/botocore/httpsession.py#L396-L397.  I think we need to implement a context per proxy url unfortunately.

thehesiod: which means a tcp connector per proxy url I think :(

jakob-keller: > which means a tcp connector per proxy url I think :(

I'm afraid, you're right. I'll try to address that in an updated PR.

thehesiod: sorry :]

thehesiod:
sorry for delay!  Kids got hand foot mouth sequentially and I've been working on getting our condo back on the market due to tenant destruction (pain) with no money (they didn't pay).  Will get to this ASAP.

jakob-keller:
> sorry for delay! Kids got hand foot mouth sequentially and I've been working on getting our condo back on the market due to tenant destruction (pain) with no money (they didn't pay). Will get to this ASAP.

Ouch, I feel ya! Take your time!

thehesiod:
is it possible to use their proxy manager concept to make the logic more similar?  would help to maintain going forwards.

jakob-keller:
> is it possible to use their proxy manager concept to make the logic more similar? would help to maintain going forwards.

I don't see how. We don't use (proxy) managers at all.

We could create a `._get_session()` method that extracts relevant logic from `.send()`, if that is what you propose. Should I give it a try?

jakob-keller:
I just rebased the PR onto main and added a commit that pulls out session creation into `._get_session()`. It makes the code a bit more readable, but introduces another round trip through the event loop, which could have a minor performance impact. Let me know, what you think. I can drop the commit easily.

thehesiod:
ok looks great, we really need to add a proxy test later though.  One suggestion, replace:
```python
            chunked = None
            if headers_.get('Transfer-Encoding', '').lower() == 'chunked':
                # aiohttp wants chunking as a param, and not a header
                headers_.pop('Transfer-Encoding', '')
                chunked = True
```
with something like this function after `_get_session`
```python
    def _chunked(self, headers):
        transfer_encoding = headers.get('Transfer-Encoding', '')
        if chunked := transfer_encoding.lower() == 'chunked':
            # aiohttp wants chunking as a param, and not a header
            headers.pop('Transfer-Encoding', '')
        return chunked
```
so it matches botocore some more.

thehesiod:
oh and then `chunked=self._chunked(headers_),`

thehesiod:
thank you and sorry for the delays!  finally finished the painting, now for all the other 1000 tiny things to fix lol

jakob-keller:
I'm not sure we need to deal with the chunked parameter at all. To me it seems that [aithttp is able to handle that internally](https://github.com/aio-libs/aiohttp/blob/0e91eb0bd94298e501084f245df44cc1772c49e5/aiohttp/client_reqrep.py#L435) at least since https://github.com/aio-libs/aiohttp/commit/2c858346b6b82a703e9b187ee26a95f3f6772a67.

aiobotocore/httpsession.py
```diff
@@ -195,20 +215,15 @@ async def send(self, request):
...
                 data = _IOBaseWrapper(data)
 
             url = URL(url, encoded=True)
-            response = await self._session.request(
+            session = await self._get_session(proxy_url)
+            response = await session.request(
                 request.method,
                 url=url,
-                chunked=chunked,
+                chunked=self._chunked(headers_),
```
jakob-keller, line 226: I'm not sure we need to deal with the chunked parameter at all. To me it seems that [aithttp is able to handle that internally](https://github.com/aio-libs/aiohttp/blob/0e91eb0bd94298e501084f245df44cc1772c49e5/aiohttp/client_reqrep.py#L435) at least since https://github.com/aio-libs/aiohttp/commit/2c858346b6b82a703e9b187ee26a95f3f6772a67.

jakob-keller:
I have no idea, why CI is failing. This should be unrelated to this PR.

aiobotocore/httpsession.py
thehesiod: hmm, seems like there's test coverage so it seems like you could pull it out to see what happens (assuming it's an integration test to ensure it's testing w/ aiohttp and moto)

thehesiod: although I'd open an issue with comments for better testing in this area as well to do later unless you feel very motivated :)

thehesiod:
hmm: 
```            if report.has_errors():
>               raise ParamValidationError(report=report.generate_report())
E               botocore.exceptions.ParamValidationError: Parameter validation failed:
E               Invalid type for parameter TopicArn, value: None, type: <class 'NoneType'>, valid types: <class 'str'>
```
https://github.com/aio-libs/aiobotocore/actions/runs/8129847790/job/22217412080?pr=1077

thehesiod:
hmm, also seeing an xml parse error, I wonder if it's moto that got upgraded

aiobotocore/httpsession.py
jakob-keller: > although I'd open an issue with comments for better testing in this area as well to do later unless you feel very motivated :)

Unit tests in what's still an unfamiliar codebase are not my strong suit :-/

jakob-keller: > Let's see what happens: #1096

CI doesn't complain, but I'm not sure that's enough evidence. Maybe aiohttp folks can weigh in?

jakob-keller:
> I have no idea, why CI is failing. This should be unrelated to this PR.

For some reason chunked=None is not equivalent to chunked=False. I just reverted to the old logic and now CI is passing.

aiobotocore/httpsession.py
thehesiod: we can just add a test later, i think werkzerg via moto should be able to handle it

thehesiod (approved):

thehesiod:
thank you so much, sorry for delay!

jakob-keller:
Thanks for the thorough review! Much appreciated!

Issue #749, opened 2020-01-11
support pluggable http clients/coroutine frameworks (anyio/httpx)


graingert:
I'd like to be able to use aiobotocore in a project that uses trio and httpx

anyio provides a generic interface to curio/trio/asyncio that libraries can use to provide generic support for all async/await eventloops

aiobotocore would require users to pass a "request" coroutine function that provides all the http functionality that aiobotocore needs for http requests

graingert:
As far as I'm aware aiobotocore would need to replace all asyncio calls with the relevant anyio equivalent, and any gather/ensure_future with a nursary/spawn pattern instead

thehesiod:
should be done with https://github.com/aio-libs/aiobotocore/issues/619 which will cause a TON of refactoring and perhaps re-thinking this whole project

graingert:
I don't think #619 and this issue are really that coupled and could be completed in any order

dazza-codes:
botocore/boto3 already supports an event callback for extensions.  See
- https://boto3.amazonaws.com/v1/documentation/api/latest/guide/events.html
- use aiobotocore as a starting point to create a new base class and hook it up to a boto3 event when it creates clients.

thehesiod:
they reason I said #619 is because it's going to greatly expand this codebase, and thus parts which would need to be refactored to support whats mentioned in this issue.  IIRC there are sleeps/locks for example in the new credentials file we need to replace components in.  @dazza-codes that's great insight for that one particular event, I just logged https://github.com/aio-libs/aiobotocore/issues/774 so we don't forget about it and apply it where possible.  Given #619 moves us to async events a lot more will be possible.

toidi:
What is the roadmap here. I would contribute if there's a plan.

`aiohttp` dependency is discouraged in projects that use a different server web framework.

graingert:
@toidi step 1 would be to find any imports/calls to asyncio and replace it with the anyio equivalent

toidi:
@thehesiod are you ok with the proposal above?

thehesiod:
Let me take a look at anyio and report back

thehesiod:
[...]

PR #1081, opened 2024-01-29
Address breaking change introduced in `aiohttp==3.9.2`

### LINKED ISSUE #882: potential aiohttp>=4.0.0 breaking changes
**Describe the bug**
I have a suspicion we'll hit some issues when aiohttp 4.0.0 comes out judging by what I've seen in this issue https://github.com/terrycain/aioboto3/issues/237 where the user had installed 4.0.0a1.

The issue was using aiobotocore 1.3.3, where verify_ssl was an unexpected argument passed to aiohttp.TCPConnector, aiohttp.TCPConnector has since moved [here](https://github.com/aio-libs/aiobotocore/blob/master/aiobotocore/httpsession.py#L90) but I reckon it would still have the same issue.

I shall look into this in more depth later today. Might be worth updating the dep to `aiohttp>=3.3.1,<4.0.0`.

jakob-keller:
### Description of Change
`aiohttp==3.9.2` introduces a change in handling SSL/TLS-related arguments that breaks our code. This PR applies current `aiohttp` best-practices to SSL/TLS-related arguments.

### Assumptions
The dependency specification for `aiohttp` was bumped to prevent regressions with our updated code and `aiohttp<3.9.2`.

* [x] If this is resolving an issue (needed so future developers can determine if change is still necessary and under what conditions) (can be provided via link to issue with these details):  closes #882
  * [ ] Why needed


jakob-keller:
The CI failure appears to be unrelated

dacevedo12:
Seems like they addressed this in 3.9.3 https://github.com/aio-libs/aiohttp/pull/8098

jakob-keller:
> Seems like they addressed this in 3.9.3 [aio-libs/aiohttp#8098](https://github.com/aio-libs/aiohttp/pull/8098)

Indeed, still we should cease using the deprecated `verify_ssl` argument to prevent this issue from popping up again in the future.

jakkdl:
> > Seems like they addressed this in 3.9.3 [aio-libs/aiohttp#8098](https://github.com/aio-libs/aiohttp/pull/8098)
> 
> Indeed, still we should cease using the deprecated `verify_ssl` argument to prevent this issue from popping up again in the future.

perhaps of note is that #749 (which I've been slowly chugging away on) wants to replace aiohttp with httpx, so there might not be a very long future for aiohttp.

aiobotocore/httpsession.py
```diff
@@ -205,7 +204,18 @@ async def send(self, request):
...
+
+            session = self._sessions.get(proxy_url)
+            if session is None:
+                connector = self._create_connector(proxy_url)
+                self._sessions[proxy_url] = session = aiohttp.ClientSession(
+                    connector=connector,
+                    timeout=self._timeout,
+                    skip_auto_headers={'CONTENT-TYPE'},
+                    auto_decompress=False,
+                )
```
thehesiod, line 216:
```suggestion
            if not (session := self._sessions.get(proxy_url)):
                connector = self._create_connector(proxy_url)
                self._sessions[proxy_url] = session = await self._exit_stack.enter_async_context(aiohttp.ClientSession(
                    connector=connector,
                    timeout=self._timeout,
                    skip_auto_headers={'CONTENT-TYPE'},
                    auto_decompress=False,
                ))
```

Then you don't need to iterate over the sessions when doing cleanup

jakob-keller: Great idea! I added the `AsyncExitStack` over at #1077 and will push here, too.

thehesiod (approved):

setup.py
```diff
@@ -8,7 +8,7 @@
 install_requires = [
     # pegged to also match items in `extras_require`
     'botocore>=1.34.41,<1.34.52',
-    'aiohttp>=3.7.4.post0,<4.0.0',
```
thehesiod, line 11: need to update github actions (python-package.yml) to only test with aiohttp > 3.9.2 and update

thehesiod (changes requested):

setup.py
jakob-keller: Haven't I done that already?

thehesiod (approved):

jakob-keller:
CI doesn't allow for two releases dated the same in CHANGES.txt :-(

thehesiod:
will fix test here, one sec

PR #1154, opened 2024-08-19
Switch to PyPI trusted publishing (#1116)

jakob-keller:
### Description of Change
Switch to PyPI trusted publishing

### Assumptions
Trusted publisher is configured for this project.

### Checklist for All Submissions
* [ ] I have added change info to [CHANGES.rst](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.rst)
* [x] If this is resolving an issue (needed so future developers can determine if change is still necessary and under what conditions) (can be provided via link to issue with these details): closes #1116
  * [ ] Detailed description of issue
  * [ ] Alternative methods considered (if any)
  * [ ] How issue is being resolved
  * [ ] How issue can be reproduced
* [ ] If this is providing a new feature  (can be provided via link to issue with these details):
  * [ ] Detailed description of new feature
  * [ ] Why needed
  * [ ] Alternatives methods considered (if any)

### Checklist when updating botocore and/or aiohttp versions

* [ ] I have read and followed [CONTRIBUTING.rst](https://github.com/aio-libs/aiobotocore/blob/master/CONTRIBUTING.rst#how-to-upgrade-botocore)
* [ ] I have updated test_patches.py where/if appropriate (also check if no changes necessary)
* [ ] I have ensured that the awscli/boto3 versions match the updated botocore version

PR #1158, opened 2024-08-20
Rename CI/CD action to conform to aio-libs standard

jakob-keller:
### Description of Change
Partially adresses https://github.com/aio-libs/aiobotocore/pull/1154#issuecomment-2297940743 by renaming our CI GitHub workflow.

Since we do not use GitHub environments and trusted publisher expects `pypi`, this might still fail. @webknjaz: Is the environment really required or could you remove it, at least temporarily?

### Assumptions
None

  * [ ] Why needed

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -1,4 +1,4 @@
-name: CI
+name: CI/CD
 
 on:
   push:
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,7 @@
 aiobotocore
 ===========
-.. |ci badge| image:: https://github.com/aio-libs/aiobotocore/actions/workflows/python-package.yml/badge.svg?branch=master
-    :target: https://github.com/aio-libs/aiobotocore/actions/workflows/python-package.yml
+.. |ci badge| image:: https://github.com/aio-libs/aiobotocore/actions/workflows/ci-cd.yml/badge.svg?branch=master
+    :target: https://github.com/aio-libs/aiobotocore/actions/workflows/ci-cd.yml
     :alt: CI status of master branch
 .. |coverage badge| image:: https://codecov.io/gh/aio-libs/aiobotocore/branch/master/graph/badge.svg
     :target: https://codecov.io/gh/aio-libs/aiobotocore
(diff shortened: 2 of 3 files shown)


jakob-keller:
If we need to create the `pypi` deployment environment, that would [need to be done by the repo owner](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment#creating-an-environment). Which means, I am blocked.

webknjaz:
Yes, the requirement is required for tighter security. And no, creating environments does not block you, technically, because they are auto-created when the corresponding job using them runs.

That said, I'll still look into creating one in order to apply protection rules and provide the last approval gate through a button on GH.

webknjaz:
Plz also add `environment.url` with `github.ref_name` rendered in. It will show up in a few places in the UI.

Additionally, I recommend running dist build unconditionally, at the very beginning of the workflow + running `twine check --strict`. It's not pleasant when the metadata validation fails at the last moment when releasing.
Besides, it'd provide a good indicator of the ability to build it continuously.
I personally even run tests against those. But this is something that could be explored in a follow-up.

Another thing I noticed is that building the dists uses a deprecated method.
https://packaging.python.org/en/latest/guides/modernize-setup-py-project/

jakob-keller:
I will look into that! Your help is much appreciated. We have accumulated a bit of technical debt here and there :-)

webknjaz:
Feel free to ask me about CI. I'm trying to streamline it across the org. I think that some good examples can be found in aiomonitor/aiomysql/yarl/frozenlist/multidict. Some aren't fully synchronized, though.

P.S. I checked the repo settings only to find out that I've already configured the environment long ago 😂

thehesiod:
am i ok to approve, anything left?

jakob-keller:
This is part of restoring publishing to PyPI and may be approved (and merged). There will be two more related PRs: #1160 and another one that introduces the pypi deployment environment. But we can split those to keep it simple.

thehesiod (approved):

jakob-keller:
#1163 will fix the missing deployment environment. After that, we should be able to publish again.

jakob-keller:
I opened #1164 to track metadata validation

PR #1163, opened 2024-08-21
Modernize CI/CD workflow

jakob-keller:
### Description of Change
Modernize CI/CD workflow based upon recommendations from https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-python#publishing-to-pypi and @webknjaz
* Replace `python setup.py` by `build`
* Use deployment environment `pypi`

### Assumptions
None

  * [ ] Why needed


.github/workflows/ci-cd.yml
```diff
@@ -19,98 +19,106 @@ jobs:
...
+      # IMPORTANT: this permission is mandatory for trusted publishing
       id-token: write
+
+    # Dedicated environments with protections for publishing are strongly recommended.
+    environment:
+      name: pypi
+      url: https://pypi.org/project/aiobotocore/${{github.ref_name}}
+
     steps:
-    - name: Download dists
```
webknjaz, line 110: Could you revert to no-indent sequences? I'm standardizing around this style in many places.

It'd be nice to drop in this config for consistency:
```console
$ cat .yamllint
---

extends: default

rules:
  indentation:
    level: error
    indent-sequences: false
  truthy:
    allowed-values:
    - >-
      false
    - >-
      true
    - >-  # Allow "on" key name in GHA CI/CD workflow definitions
      on

...
```

.github/workflows/ci-cd.yml
```diff
@@ -19,98 +19,106 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: [ '3.8', '3.9', '3.10', '3.11', '3.12', '3.13' ]
```
webknjaz, line 22: let's list things on separate lines
```suggestion
        python-version:
        - 3.8
        - 3.9
        - >-
          3.10
        - 3.11
        - 3.12
        - 3.13
```

.github/workflows/ci-cd.yml
jakob-keller: My PyCharm begs to differ ;-) I will give it a look.

.github/workflows/ci-cd.yml
```diff
@@ -19,98 +19,106 @@ jobs:
...
-  build-tarball:
-    name: Tarball
+
+  # based on https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-python#publishing-to-pypi
+  release-build:
     runs-on: ubuntu-latest
-    needs: pre-deploy
+
+    needs:
+      - test
```
webknjaz, line 73: Don't depend on the tests. If anything, it's the tests that should depend on buildig the dists. This way you'd be able to test said dists instead of building things in those jobs again.

.github/workflows/ci-cd.yml
```diff
@@ -19,98 +19,106 @@ jobs:
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: release-dists
+          path: dist/
+
+  pypi-publish:
     runs-on: ubuntu-latest
+
+    needs:
+      - release-build
```
webknjaz, line 102:
```suggestion
    - release-build
    - test
```

.github/workflows/ci-cd.yml
```diff
@@ -19,98 +19,106 @@ jobs:
...
+    if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')
+
     permissions:
+      # IMPORTANT: this permission is mandatory for trusted publishing
       id-token: write
+
+    # Dedicated environments with protections for publishing are strongly recommended.
+    environment:
+      name: pypi
+      url: https://pypi.org/project/aiobotocore/${{github.ref_name}}
```
webknjaz, line 114: mustache is sparse in all the other places
```suggestion
      url: https://pypi.org/project/aiobotocore/${{ github.ref_name }}
```

.github/workflows/ci-cd.yml
```diff
@@ -19,98 +19,106 @@ jobs:
...
+      - name: Setup Python ${{ matrix.pyver }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          allow-prereleases: true
+      - name: Lock and sync dependencies
+        env:
+          AIOHTTP_VERSION: ${{ matrix.aiohttp-version }}
+        run: |
+          python -c "import sys; print(f'Python version: {sys.version}')"
```
webknjaz, line 43: *(out of the scope, but this should really just be `python -V -V`)*

.github/workflows/ci-cd.yml
```diff
@@ -19,98 +19,106 @@ jobs:
...
+
+      - name: Check release distributions
+        run: |
+          python -m pip install twine
+          python -m twine check --strict dist/*
+
+      - name: Upload distributions
+        uses: actions/upload-artifact@v4
+        with:
+          name: release-dists
```
webknjaz, line 95: My guide https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ uses
```suggestion
          name: python-package-distributions
```
because it represents what it is better. The possible use for these artifacts is wider. They are published on workflow runs and can be downloaded for inspection even when not releasing.
Additionally, if you get to testing them at some point, they'll be downloaded into the test job. So I'd not use the word “release” in the name.

P.S. Because they might end up being passed around into more places, I tend to use the name into a variable (like in the `${{ env }}` context) and just use that var everywhere instead of hardcoding the name in multiple places.

.github/workflows/ci-cd.yml
webknjaz: yeah, that's why there should be a shared config so that every PR author wouldn't reformat by accident. I usually have yamllint wired via the pre-commit.com tool

jakob-keller: Turns out PyCharm supports that, too: https://www.jetbrains.com/help/pycharm/code-style-yaml.html#tabs_and_indents

jakob-keller: Opened #1168 for standardization

.github/workflows/ci-cd.yml
jakob-keller: I remove the dependency and opened #1169

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools"]
+requires = ["setuptools>=72.2.0"]
```
webknjaz, line 2: Let's make it a bit more sparse:
```suggestion
requires = ["setuptools >= 72.2.0"]
```

Also, what are the differences with older versions? If they output comparable results (just not byte-to-byte), it might be fine to have a lower value here.

jakob-keller: Yes, I believe it would be fine, but lack the experience and time to confirm that. I see no harm in aiming for a relatively high lower bound. There shouldn't be any conflicts due to build isolation, right?

.github/workflows/ci-cd.yml
```diff
@@ -1,31 +1,61 @@
...
+        - '3.11'
+        - '3.12'
+        - '3.13'
         include:
-          - aiohttp-version: '==3.9.2'
-          - aiohttp-version: '<4.0.0'
-            python-version: '3.11'
+        - aiohttp-version: '==3.9.2'
+        - aiohttp-version: '<4.0.0'
+          python-version: '3.11'
```
webknjaz, line 56: I personally dislike quoting all the unnecessary places. It's inconsistent: we don't quote `name:` or `runs-on:` values, but do this for arbitrary strings — that's two styles in one.

Whenever I need something as an explicit string, I use a `>-` which allows me to put stuff on a separate line, drop the quotes and I don't have to escape any quotes that may be met inside such strings so it looks cleaner.

.github/workflows/ci-cd.yml
```diff
@@ -1,31 +1,61 @@
+---
 name: CI/CD
 
 on:
   push:
```
webknjaz, line 5: In a follow-up PR, I'd also enable merge queues.

pyproject.toml
webknjaz: For as long as it supports all required Python versions, this should be fine. The only case where conflicts may arise is downstream packaging, which you'll hear about if somebody won't be able to repackage it for Debian or something else with conservative software upgrades policies.

.github/workflows/ci-cd.yml
```diff
@@ -1,31 +1,62 @@
...
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version:
+        - '3.8'
+        - '3.9'
+        - '3.10'
+        - '3.11'
+        - '3.12'
+        - '3.13'
```
webknjaz, line 52: it is possible to unquote these as follows:
```suggestion
        - 3.8
        - 3.9
        - >-
          3.10
        - 3.11
        - 3.12
        - 3.13
```

.github/workflows/ci-cd.yml
```diff
@@ -57,60 +88,32 @@ jobs:
...
-
-  deploy:
-    name: Deploy
-    needs: [build-tarball]
+  pypi-publish:
+    name: Publish Python 🐍 distribution 📦 to PyPI
+    if: startsWith(github.ref, 'refs/tags/')  # only publish on tag pushes
+    needs:
+    - build
+    - test
```
webknjaz, line 103: I just realized that this project doesn't integrate alls-green (yet). If you add it in a follow-up, it'll become possible to have a unified gate for use for branch protection, which would in turn let us enable safe auto-merges. Example: https://github.com/cherrypy/cheroot/blob/1ff20b1/.github/workflows/ci-cd.yml#L1147-L1166.

.github/workflows/ci-cd.yml
```diff
@@ -57,60 +88,32 @@ jobs:
...
-      with:
-        name: dist
-        path: dist
-
-  deploy:
-    name: Deploy
-    needs: [build-tarball]
+  pypi-publish:
+    name: Publish Python 🐍 distribution 📦 to PyPI
+    if: startsWith(github.ref, 'refs/tags/')  # only publish on tag pushes
```
webknjaz, line 100: This check is the old way, I think GH implemented a helper variable after I added this to the guide. I should probably update the examples :)

```suggestion
    if: github.ref_type == 'tag'
```

.github/workflows/ci-cd.yml
```diff
@@ -1,31 +1,62 @@
...
+        - '3.13'
         include:
-          - aiohttp-version: '==3.9.2'
-          - aiohttp-version: '<4.0.0'
-            python-version: '3.11'
+        - aiohttp-version: ==3.9.2
+        - aiohttp-version: <4.0.0
+          python-version: '3.11'
       fail-fast: false
     timeout-minutes: 15
```
webknjaz, line 58: This reminded me of a linting check that could be integrated in a follow-up: https://github.com/aio-libs/frozenlist/blob/826e261/.pre-commit-config.yaml#L29-L45.

.github/workflows/ci-cd.yml
```diff
@@ -1,31 +1,62 @@
...
 name: CI/CD
 
 on:
   push:
     branches:
-      - 'master'
-    tags: [ '[0-9]+.[0-9]+.[0-9]+' ]
+    - master
+    tags:
+    - '[0-9]+.[0-9]+.[0-9]+'
```
webknjaz, line 9: I usually do this
```suggestion
    - >-
      [0-9]+.[0-9]+.[0-9]+
```

webknjaz (approved):
This seems good enough. Feel free to postpone addressing any comments to follow-ups, if you want.

.github/workflows/ci-cd.yml
jakob-keller: Opened as #1171

.github/workflows/ci-cd.yml
jakob-keller: Opened as #1172

jakob-keller:
This should restore our ability to release to PyPI! Thank you @webknjaz for all the great input. I definitely learned a lot this week :-)

jakob-keller:
@thehesiod: From my point of view, now would be a good time to cut a release. What do you think?

jakob-keller:
With regards to the previously used PyPI credentials: we should both revoke them over at PyPI as well as remove them from this repo's settings.

thehesiod:
> @thehesiod: From my point of view, now would be a good time to cut a release. What do you think?

are there any substantial changes for users that would be included?

jakob-keller:
> > @thehesiod: From my point of view, now would be a good time to cut a release. What do you think?
> 
> are there any substantial changes for users that would be included?

Anything strictly user-facing is included in `CHANGES.rst`. Beyond that, there are plenty of contributor-facing changes which should be addressed by updated documentation. Not sure if those need to go into `CHANGES.rst`, too.

jakob-keller:
It's quite a lot of commits and sizeable diff: https://github.com/aio-libs/aiobotocore/compare/2.13.2...master

thehesiod:
> It's quite a lot of commits and sizeable diff: [2.13.2...master](https://github.com/aio-libs/aiobotocore/compare/2.13.2...master)

ya we have real changes, sounds good!

webknjaz:
@jakob-keller in other projects, we sometimes make an alpha-release to safely test out the pipeline. I think you could do that here as well, to get familiar with the updates. You can do that from a separate branch if you don't want a commit on main.

jakob-keller:
2.13.3-alpha it is: https://github.com/aio-libs/aiobotocore/releases/tag/2.13.3-alpha

jakob-keller:
Guess it didn't trigger due to the restrictive on push tags filter. Will do some experiments...

jakob-keller:
Actually, our release process is pretty coarse: It does not involve any commits at all, just a tag. I would skip the alpha and go straight to 2.13.3. The only thing that could go wrong IMO is the publishing to PyPI, most likely due to permissions. In that case, we can fix it and try again, YOLO!

webknjaz:
I deleted the repository secret, however the bot account on PyPI did not have a token. So it must be under somebody else's human account.

jakob-keller:
There we go: https://github.com/aio-libs/aiobotocore/releases/tag/2.13.3

@thehesiod: Feel free to edit the `What's Changed` section, if it looks to verbose.

jakob-keller:
Nice, I didn't know there was a manual review step involved. LGTM!

webknjaz:
> 2.13.3-alpha

That's not compliant with PEP 440. It would be 2.13.3.a0 in Python ecosystem.

webknjaz:
> Actually, our release process is pretty coarse: It does not involve any commits at all, just a tag.

Well, since this isn't using `setuptools-scm` (maybe it should be integrated — I usually default to having it), I would imagine that the version number bump would have to be committed and then that commit would need to be tagged. Otherwise, they'd be out of sync.

webknjaz:
> I would imagine that the version number bump would have to be committed and then that commit would need to be tagged. Otherwise, they'd be out of sync.

Additionally, that commit might make the change log consistent. In projects with Towncrier, it'd be generated from fragments.

jakob-keller:
> > 2.13.3-alpha
> 
> That's not compliant with PEP 440. It would be 2.13.3.a0 in Python ecosystem.

Good to know, I simply went with GitHub UI's suggestion.

> > Actually, our release process is pretty coarse: It does not involve any commits at all, just a tag.
> 
> Well, since this isn't using `setuptools-scm` (maybe it should be integrated — I usually default to having it), I would imagine that the version number bump would have to be committed and then that commit would need to be tagged. Otherwise, they'd be out of sync.

We still do that manually, prior to release.

jakob-keller:
> > I would imagine that the version number bump would have to be committed and then that commit would need to be tagged. Otherwise, they'd be out of sync.
> 
> Additionally, that commit might make the change log consistent. In projects with Towncrier, it'd be generated from fragments.

I agree, but that's probably #1167.

webknjaz:
In projects, where the version is hardcoded, it's usually bumped to the next `.dev0` after each release, and then, turned to stable before the next one.

jakob-keller:
It worked like a charm: https://pypi.org/project/aiobotocore/

webknjaz:
> > > 2.13.3-alpha

> > That's not compliant with PEP 440. It would be 2.13.3.a0 in Python ecosystem

> Good to know, I simply went with GitHub UI's suggestion.

If you didn't update the version in code, that what would've ended up being used. And if you'd have burned that version number, and you'd need to bump to the next stable because the release artifacts are immutable on PyPI. Good thing it didn't trigger.

thehesiod:
> There we go: https://github.com/aio-libs/aiobotocore/releases/tag/2.13.3
> 
> @thehesiod: Feel free to edit the `What's Changed` section, if it looks too verbose.

I typically don't really include internal related changes as they're not relevant for the consumer of the library, just add noise to end-users

thehesiod:
> > 2.13.3-alpha
> 
> That's not compliant with PEP 440. It would be 2.13.3.a0 in Python ecosystem.

@jakob-keller we should have a test to ensure the CI fails if the version is invalid (can parse with disttools or something similar)

jakob-keller:
> > There we go: https://github.com/aio-libs/aiobotocore/releases/tag/2.13.3
> > @thehesiod: Feel free to edit the `What's Changed` section, if it looks too verbose.
> 
> I typically don't really include internal related changes as they're not relevant for the consumer of the library, just add noise to end-users. Perhaps we can add internal changes in a collapsible section

I split the release notes as suggested.

jakob-keller:
> > > 2.13.3-alpha
> > 
> > 
> > That's not compliant with PEP 440. It would be 2.13.3.a0 in Python ecosystem.
> 
> @jakob-keller we should have a test to ensure the CI fails if the version is invalid (can parse with disttools or something similar). Can do this next to our existing version checks

Opened as #1177

webknjaz:
Here, I came up with some extra granular sections that may target some changelog reading audiences that would otherwise be treated as internal: https://yarl.aio-libs.org/en/latest/changes/#id7.

pyproject.toml
jakob-keller: You were right, of course: #1193 got hit by this.

PR #1165, opened 2024-08-21
Modernize docs

### LINKED ISSUE #1161: Retrieve version via importlib.metadata in docs/conf.py
Technically, it's best to have the project installed into the same venv as Sphinx and then, there's an stdlib library to retrieve the installed metadata. That would be `importlib.metadata.version('aiobotocore')`. That would be the most accurate info you can get.

_Originally posted by @webknjaz in https://github.com/aio-libs/aiobotocore/pull/1160#discussion_r1724139667_

jakob-keller:
### Description of Change
Closes #1161

### Assumptions
None

  * [ ] Why needed

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -10,6 +10,7 @@ sphinx:
 
 python:
   install:
+    - path: .
     - requirements: docs/requirements.txt
 
 formats:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ boto3 = [
 
 [project.urls]
 Repository = "https://github.com/aio-libs/aiobotocore"
-Documentation = "https://aiobotocore.readthedocs.io"
+Documentation = "https://aiobotocore.aio-libs.org"
 
 [tool.setuptools.dynamic]
 version = { attr = "aiobotocore.__version__" }
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,26 +27,19 @@
 #
 # needs_sphinx = '1.0'
 
-import pathlib
+import importlib.metadata
 import re
 
-_docs_path = pathlib.Path(__file__).parent
-_version_path = _docs_path / '../aiobotocore/__init__.py'
-
-
-with _version_path.open() as fp:
-    try:
-        _version_info = re.search(
-            r"^__version__ = '"
-            r"(?P<major>\d+)"
-            r"\.(?P<minor>\d+)"
-            r"\.(?P<patch>\d+)"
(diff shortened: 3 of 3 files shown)


pyproject.toml
```diff
@@ -42,6 +42,9 @@ awscli = [
 boto3 = [
     "boto3 >=1.34.70, <1.34.163",
 ]
+docs = [
```
webknjaz, line 45: Please, don't abuse extras as dependency groups. Why do you want to provide the end-users with a `pip install 'aiobotocore [docs]'`? It's a public API that you'll end up having to maintain. Besides, putting pinned deps into the installed distribution metadata is a bad tone for libs. Additionally, no dependency management tools (like `pip-tools`) would be able to sanely manage environment pins.

.readthedocs.yml
webknjaz, line 1: FWIW, it's nice to run the same automation @ RTD, CI, and locally. I like using tox for this (some people go for nox).

You can make RTD call tox as follows:
```console
$ cat .readthedocs.yaml                                                                  
# Read the Docs configuration file
# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
---

version: 2

build:
  os: ubuntu-24.04
  tools:
    python: >-  # has to be parsed as a YAML string
      3.11
  commands:
  - >-
    PYTHONWARNINGS=error
    python3 -Im venv "${READTHEDOCS_VIRTUALENV_PATH}"
  - >-
    PYTHONWARNINGS=error
    "${READTHEDOCS_VIRTUALENV_PATH}"/bin/python -Im
    pip install tox
  - >-
    PYTHONWARNINGS=error
    "${READTHEDOCS_VIRTUALENV_PATH}"/bin/python -Im
    tox -e build-docs --notest -vvvvv
  - >-
    PYTHONWARNINGS=error
    "${READTHEDOCS_VIRTUALENV_PATH}"/bin/python -Im
    tox -e build-docs --skip-pkg-install -q
    --
    "${READTHEDOCS_OUTPUT}"/html -b dirhtml
...
```

Another example: https://github.com/pytest-dev/pytest-asyncio/pull/920

This would allow wiring such a self-install into tox that could be run locally and in any other envs the same way.

webknjaz (approved):
I think this, is good assuming that consolidating sphinx runs would be done separately.

pyproject.toml
jakob-keller: Good point! I reverted that part and managed to install the project itself all the same. Looks much better to me now.

.readthedocs.yml
jakob-keller: opened as #1166

Issue #1169, opened 2024-08-21
Test distribution packages in CI


jakob-keller:
If anything, it's the tests that should depend on building the dists. This way you'd be able to test said dists instead of building things in those jobs again.

_Originally posted by @webknjaz in https://github.com/aio-libs/aiobotocore/pull/1163#discussion_r1725178624_

PR #1185, opened 2024-08-23
Streamline dev dependency management

jakob-keller:
### Description of Change
Prepares CI/CD workflow for testing of distribution packages (#1169)

### Assumptions
* `aiohttp` version constraint in CI/CD workflow is no longer required
* `time` statements in CI/CD workflow are no longer required

  * [ ] Why needed


.gitignore
```diff
@@ -1,5 +1,5 @@
 # generated by users and intentionally not pushed to the repo
```
webknjaz, line 1: This is weird. pip-tools based workflows are specifically built around having "lock files" checked in for reproducibility..

webknjaz:
What's the game plan here? This is definitely different from anything I've ever seen in the wild…

.gitignore
thehesiod: to make this legit @jakob-keller we should have one lockfile per python version and delete this cmt and txt

thehesiod:
yea, was fast and lose, we should tighten it up

.gitignore
webknjaz: Not just per python but per each permutation of python+OS+arch. This is what extensive locking looks like: https://github.com/cherrypy/cheroot/tree/main/requirements. However, to implement such an integrated environment, a lot of effort is required, so I'd postpone.

jakob-keller: I was aware of that and agree that it should be addressed at some point. As it stands, we could throw out pip-tools from CI/CD and just run `pip install .[awscli,boto3] -r requirements-dev.in` or similar. But I suggest to leave it as it is for now and have a separate effort to improve that aspect, maybe even switch dependency management tool again (poetry, hatch, ...).

jakob-keller:
I've opened #1188 for tracking the step-change improvement in dependency management.

Nonetheless, I believe it makes sense to do this PR now to throw out some technical debt.

webknjaz (approved):

PR #1196, opened 2024-09-03
Load botocore i/o methods in executor

chemelli74:
### Description of Change

Follow-up of #1120

Added the option to load in the executor `botocore` methods that do I/O.
To enable the new approach is enough to call

```python
AioSession(load_executor=True)
```

nothing changes for all current downstream implementations of this library as default is False and thus is not needed to be specified.

### Assumptions

  * [ ] Why needed

### DIFF
--- a/aiobotocore/client.py
+++ b/aiobotocore/client.py
@@ -1,3 +1,5 @@
+import asyncio
+
 from botocore.auth import resolve_auth_type
 from botocore.awsrequest import prepare_request_dict
 from botocore.client import (
@@ -41,26 +43,23 @@ class AioClientCreator(ClientCreator):
         api_version=None,
         client_config=None,
         auth_token=None,
+        load_executor=False,
     ):
         responses = await self._event_emitter.emit(
             'choose-service-name', service_name=service_name
         )
         service_name = first_non_none_response(responses, default=service_name)
-        service_model = self._load_service_model(service_name, api_version)
-        try:
-            endpoints_ruleset_data = self._load_service_endpoints_ruleset(
-                service_name, api_version
+        logger.debug(
+            "AioClientCreator - Method load_service_model[botocore] could generate I/O. Running in executor: %s",
+            load_executor,
+        )
+        if load_executor:
+            model_data = await asyncio.get_running_loop().run_in_executor(
+                None, self._load_models, service_name, api_version
             )
-            partition_data = self._loader.load_data('partitions')
-        except UnknownServiceError:
-            endpoints_ruleset_data = None
-            partition_data = None
-            logger.info(
-                'No endpoints ruleset found for service %s, falling back to '
-                'legacy endpoint routing.',
-                service_name,
-            )
-
+        else:
+            model_data = self._load_models(service_name, api_version)
+        service_model, endpoints_ruleset_data, partition_data = model_data
         cls = await self._create_client_class(service_name, service_model)
         region_name, client_config = self._normalize_fips_region(
             region_name, client_config
@@ -108,6 +107,23 @@ class AioClientCreator(ClientCreator):
             service_client, endpoint_url, client_config
         )
         return service_client
+
+    def _load_models(self, service_name, api_version):
+        service_model = self._load_service_model(service_name, api_version)
+        try:
+            endpoints_ruleset_data = self._load_service_endpoints_ruleset(
+                service_name, api_version
+            )
+            partition_data = self._loader.load_data('partitions')
+        except UnknownServiceError:
+            endpoints_ruleset_data = None
+            partition_data = None
+            logger.info(
+                'No endpoints ruleset found for service %s, falling back to '
+                'legacy endpoint routing.',
+                service_name,
+            )
+        return service_model, endpoints_ruleset_data, partition_data
 
     async def _create_client_class(self, service_name, service_model):
         class_attributes = self._create_methods(service_model)
--- a/aiobotocore/session.py
+++ b/aiobotocore/session.py
@@ -1,7 +1,12 @@
+import asyncio
+import functools
+
 from botocore import UNSIGNED
 from botocore import __version__ as botocore_version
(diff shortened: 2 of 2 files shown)


aiobotocore/client.py
```diff
@@ -41,26 +43,23 @@ async def create_client(
...
-            endpoints_ruleset_data = None
-            partition_data = None
-            logger.info(
-                'No endpoints ruleset found for service %s, falling back to '
-                'legacy endpoint routing.',
-                service_name,
+        logger.debug(
+            "AioClientCreator - Method load_service_model[botocore] could generate I/O. Running in executor: %s",
+            load_executor,
+        )
```
bdraco, line 55: I think it would be better to add this to the documentation instead

aiobotocore/session.py
```diff
@@ -103,9 +110,25 @@ async def get_service_data(self, service_name, api_version=None):
         Retrieve the fully merged data associated with a service.
         """
         data_path = service_name
-        service_data = self.get_component('data_loader').load_service_model(
-            data_path, type_name='service-2', api_version=api_version
+        data_loader: Loader = self.get_component('data_loader')
+        logger.debug(
+            "AioSession - Method load_service_model[botocore] could generate I/O. Running in executor: %s",
+            self.load_executor,
         )
```
bdraco, line 117: As above

jakob-keller (commented):
Thank you for creating the PR.

Having taken a brief look, several concerns prevent me from supporting this proposal at this time:

1. Unless absolutely necessary, we should avoid making any changes to the signature of upstream functions and classes, in particular public APIs. This proposal introduces several API changes, including to `create_client()`. While they might well be backwards compatible, they introduce drift and might make it more difficult to stay up-to-date with future botocore releases.
2. botocore (and by extension aiobotocore) has a powerful configuration mechanism that could (should?) be used instead.
3. As far as I can tell, the botocore loader is far more complex and has a lot more blocking codepaths than are considered here. Having a fully async loader would be preferable to me. I have started work in that area, but it will be a massive change and I am not sure I can pull through and finish it.

jakob-keller:
Here's a recent example of adding a configuration option: #1102

jakob-keller:
Check https://github.com/boto/botocore/blob/develop/botocore/loaders.py to get a glimpse of the number of I/O operations that exist and might need to be run in the executor. If you then identify all calling code throughout aiobotocore/botocore the task of avoiding any blocking calls becomes daunting, at least to me.

jakob-keller:
For your reference, I have pushed a branch with progress on fully async loaders to https://github.com/jakob-keller/aiobotocore/tree/async-loaders. This is nowhere near complete, though.

chemelli74:
> Here's a recent example of adding a configuration option: #1102

So if I go that way, are you willing to add the change ?

jakob-keller:
> > Here's a recent example of adding a configuration option: #1102
> 
> So if I go that way, are you willing to add the change ?

That would only address 1. and 2. I feel we should have a discussion about 3. before more effort is invested in this or any other PR. But that is just my opinion.

chemelli74:
> > > Here's a recent example of adding a configuration option: #1102
> > 
> > 
> > So if I go that way, are you willing to add the change ?
> 
> That would only address 1. and 2. I feel we should have a discussion about 3. before more effort is invested in this or any other PR. But that is just my opinion.

I would say that is a start.
So if we go that way current discovered issues are fixed.
We can always extend or as you are doing rework it later on.

What you think ?

jakob-keller:
> > > > Here's a recent example of adding a configuration option: #1102
> > > 
> > > 
> > > So if I go that way, are you willing to add the change ?
> > 
> > 
> > That would only address 1. and 2. I feel we should have a discussion about 3. before more effort is invested in this or any other PR. But that is just my opinion.
> 
> I would say that is a start. So if we go that way current discovered issues are fixed. We can always extend or as you are doing rework it later on.
> 
> What you think ?

What specific issue are you referencing? When #1120 was discussed, no consensus was reached that this is even an issue with aiobotocore.

chemelli74:
> > > > > Here's a recent example of adding a configuration option: #1102
> > > > 
> > > > 
> > > > So if I go that way, are you willing to add the change ?
> > > 
> > > 
> > > That would only address 1. and 2. I feel we should have a discussion about 3. before more effort is invested in this or any other PR. But that is just my opinion.
> > 
> > 
> > I would say that is a start. So if we go that way current discovered issues are fixed. We can always extend or as you are doing rework it later on.
> > What you think ?
> 
> What specific issue are you referencing? When #1120 was discussed, no consensus was reached that this is even an issue with aiobotocore.

I don't know what to say or add.

The issue is made evident by HA, `botocore` maintainer wrote that it is a problem with the async code.
The above code fix the issue with the blocking calls inside the event loop

Please help me understanding what is the missing information.

jakob-keller:
First of all, I am not aware that a related issue has been raised with aiobotocore. This might well be a technicality, but that would be the ideal place to clarify where the issue lies, if there are any work arounds available and/or whether and how to best address it. Discussing the apparent issue, potential solutions and a proposed fix in the context of a PR has not worked well for #1120 and I feel this PR might be on a similar path. I suggest to open an issue and continue the discussion there. Thank you!

chemelli74:
> I suggest to open an issue and continue the discussion there. Thank you!

Done, #1199

aiobotocore/session.py
```diff
@@ -37,6 +42,7 @@ def __init__(
         event_hooks=None,
         include_builtin_handlers=True,
         profile=None,
+        load_executor=False,
```
thehesiod, line 45: instead of this being a bool, you should just pass in the executor you want to use for the loader, that way you avoid defaulting to creating a ton of threads

thehesiod:
deprecating in favor of more targeted https://github.com/aio-libs/aiobotocore/pull/1451/

PR #1218, opened 2024-10-22
delete stale for now until we have more resources

thehesiod:
seems to be causing more work than solving


jakob-keller (approved):
Fine with me. Alternatively, we could just deactivate that workflow.

thehesiod:
ah ok, did that for now

Issue #1251, opened 2024-12-31
Improve handling of upstream dev dependencies


jakob-keller:
If you mix files from upstream with local changes, I recommend vendoring them separately and including into yours. This allows you to sync the contents by copying blindly.

_Originally posted by @webknjaz in https://github.com/aio-libs/aiobotocore/pull/1246#discussion_r1900144472_

PR #1257, opened 2025-01-07
Leverage `uv`

### LINKED ISSUE #1188: Use best practice dependency management
Not just per python but per each permutation of python+OS+arch. This is what extensive locking looks like: https://github.com/cherrypy/cheroot/tree/main/requirements. However, to implement such an integrated environment, a lot of effort is required, so I'd postpone.

_Originally posted by @webknjaz in https://github.com/aio-libs/aiobotocore/pull/1185#discussion_r1729649752_

jakob-keller:
### Description of Change
Leverage `uv` for local development environment and CI/CD pipeline.

### Assumptions

* [x] If this is resolving an issue (needed so future developers can determine if change is still necessary and under what conditions) (can be provided via link to issue with these details): closes #1188
  * [ ] Why needed


pyproject.toml
```diff
@@ -53,6 +53,45 @@ boto3 = [
...
+    "coverage==7.2.7",
+    # "setuptools==71.1.0;python_version>='3.12'",
+    # "packaging==24.1;python_version>='3.12'", # Requirement for setuptools>=71
+
+    # Pytest specific deps
+    "pytest==8.1.1",
+    "pytest-cov==5.0.0",
+    "pytest-xdist==3.5.0",
+    # "atomicwrites>=1.0", # Windows requirement
+    # "colorama>0.3.0", # Windows requirement
```
thehesiod, line 71: what are these commented out packages for?

pyproject.toml
```diff
@@ -53,6 +53,45 @@ boto3 = [
...
+    "dill ~=0.3.3", # Used in test_patches.py
+    "docker ~=7.1",
+    "docutils ~=0.16", # Used in test_version.py
+    "moto[server,s3,sqs,awslambda,dynamodb,cloudformation,sns,batch,ec2,rds] ~=4.2.9",
+    "packaging ~=24.1", # Used in test_version.py
+    "pip ~=24.3.1", # Used in test_version.py
+    "pre-commit ~=3.5.0",
+    "pytest-asyncio ~=0.23.8",
+    "requests ~=2.32.3", # Used in test_version.py
+    "time-machine ~=2.15.0",
```
thehesiod, line 83: this is new to me, needed?

pyproject.toml
```diff
@@ -53,6 +53,45 @@ boto3 = [
...
+    "tomli; python_version<'3.11'", # Used in test_version.py
+    "werkzeug ~=3.0.6", # Used in moto_server.py
```
thehesiod, line 85: this is a side-effect of moto, if moto changes werkzeug we need to change our impl as well.  IOW this is not our choice

thehesiod: if it makes you feel better can import this from moto directly ;)

CONTRIBUTING.rst
```diff
@@ -13,31 +13,24 @@ First of all, clone the repository::
 
     $ git clone git@github.com:aio-libs/aiobotocore.git
 
-Create virtualenv with at least python3.8 (older versions are not supported).
```
webknjaz, line 16: I don't think it's a good idea to dictate contributing workflows in general. Less vendor locking is better.

CONTRIBUTING.rst
```diff
@@ -13,31 +13,24 @@ First of all, clone the repository::
...
 
+Create a virtual environment::
 
-After that, please install libraries required for development::
-
-    $ pip install pip-tools
-    $ pip-compile --all-extras pyproject.toml requirements-dev.in
-    $ pip-sync
-    $ pip install -e .
+   $ cd aiobotocore
```
webknjaz, line 20: This probably shouldn't be dedented. Does it render well?

webknjaz (commented):
My biggest concern is forcing yet another tool on contributors.

pyproject.toml
jakob-keller: As far as I can tell, those are dev dependencies from upstream that we do not currently need.

This PR does not change the dev dependencies in any way, it merely replaces the existing requirements-dev.in file into the [PEP 735](https://peps.python.org/pep-0735/) compliant [dependency-groups] table in pyproject.toml.

pyproject.toml
jakob-keller: Again, these are the exact same dev dependencies that we already use and which have been specified in `dependencies-dev.in`. This is now migrated to [dependency-groups] in `pyproject.toml`.

pyproject.toml
jakob-keller: Sorry, if I'm mistaken, but I thought we had resolved that as part of #1246. We use `werkzeug` directly in unit tests and that specification guards us from regressions should moto drop that dependency. I agree that we might then subsequently also drop werkzeug, but that's a separate issue, IMO.

jakob-keller:
> My biggest concern is forcing yet another tool on contributors.

My sales pitch would be that in our context `uv` is primarily a direct replacement of `pip-tools`, but  also (optionally) substitutes:
* `pyenv`, installers from python.org, ... (to provision Python toolchain)
* `virtualenvwrapper`, `venv`, ...? (for managing virtual environments)
* `pip` (as package installer)
* `build` (as build frontend in CI/CD)

For me personally, having worked mostly on poetry based projects, it's a single, well-known tool, that is trivial to adopt and improves developer experience by a lot. And it may be used in conjuncture with existing tools, if one prefers to take that route.

CONTRIBUTING.rst
jakob-keller: That's a valid point and I strongly convinced our documentation may be improved upon. Since I'm not a technical writer nor an English native, I hesitate to take the initiative.

What I like about `uv` is that it works well out-of-the-box and makes many of the previously documented steps obsolete.

jakob-keller: If there was a standard lock file format and tooling that supported it, we would be in a different place, but as of now, I believe uv offers the best overall experience and is a major step forward for this project.

I there is no consensus, we may well close this PR. No hard feelings, it was not a lot of effort to put this together.

pyproject.toml
webknjaz: I'm wondering if there's merit in having a dependency group for the upstream deps. You could just keep a requirements file in sync with the other repo if it's truly needed and then, extract whatever is relevant into this group, I suppose...

pyproject.toml
webknjaz: Exactly: https://github.com/aio-libs/aiobotocore/blob/7432338/tests/moto_server.py#L12C1-L12C16

CONTRIBUTING.rst
jakob-keller: It was rendering well (https://aiobotocore--1257.org.readthedocs.build/en/1257/contributing.html), but thanks for pointing out the mixed up indentations. It's fixed now.

CONTRIBUTING.rst
webknjaz: I think that it's good for CI. I'm just concerned that telling people to install a tool that messes with their userspace by default is not to be taken lightly.
Personally, I tend to be cautious when it comes to such things. I'm comfortable with `tox` because I know it won't affect my env beyond the project dir, for example.
I've been playing with pipx/uvx recently, but really carefully.
I think that there would be a portion of contributors that would feel like we're sending them off to learn an entirely new tool and its implications. So if there's a way to improve this, it should be taken into account.
This was one of my problems w/ Poetry, by the way. To this day, I avoid interacting with projects that don't let me do my thing w/o touching it.
OTOH, it's probably fine where there's CI to validate changes — in such cases, I just edit the files, push the thing and hope to see some feedback from the interface that isn't as eager to interfere with how I get to use my computer.

webknjaz:
Well, let's try it out. I've outlined my concerns and I think that `uv pip-compile` would've been less intrusive. But I suppose it's okay to attempt doing this and revert if it doesn't work out.

pyproject.toml
jakob-keller: We already have #1251 to track improvement of handling of upstream dev dependencies.

The motivation of following relevant upstream dev dependencies closely is that we attempt to keep upstream code in our codebase as close to verbatim as possible to simplify maintenance. Any drift in formatting, linter suppression comments and so forth is undesired.

I am proposing to split the existing `requirements-dev.in` file into two separate dependency groups to make the distinction between upstream dev dependencies and aiobotocore dev dependencies explicit for the first time. Just to avoid any confusion: These are not "extras" that leak into the published distribution packages in any way, it is merely a logical grouping of dev dependencies, intended to simplify their handling.

CONTRIBUTING.rst
jakob-keller: What is your take on dev containers? Would that address your objections? I want to learn about those this year.

jakob-keller: Honestly, all active contributors have already commented on this PR. Not sure if there is anyone else that feels strongly about these issue - or at all.

webknjaz: > What is your take on dev containers? Would that address your objections? I want to learn about those this year.

I've been meaning to look deeper into these as well. I think, they might be a good interface to declare (and keep working through CI). They'd enable more contributors, especially those from the VS Code land or GitHub Codespaces.

I'm a NeoVim user (AstroNvim mostly) and I'd like to try them out too, I've heard there's integrations.

But I think the main benefit is that enabling more contributors to start seamlessly is not blocking anybody who wouldn't want to opt in.


> Honestly, all active contributors have already commented on this PR. Not sure if there is anyone else that feels strongly about these issue - or at all.

Fair. Although, I'm mostly thinking of potential contributors, not the existing ones. The existing contributors are more likely to adapt and try out different things. But if for somebody the entry barrier is becoming higher, it might gate them and prevent them from trying.

pyproject.toml
thehesiod: please don't resolve conversations started by reviewers, bad etiquette.  Anywho this seems like it was added in https://github.com/aio-libs/aiobotocore/commit/a73f8c26192b190f503791b309124cb32eac44c8 for seemingly no reason.  We should not be adding random requirements that aren't needed

thehesiod: if it is needed it needs a comment as to why it's here

pyproject.toml
thehesiod: no, see my reply: https://github.com/aio-libs/aiobotocore/pull/1246#discussion_r1906210969

thehesiod: if moto stops using that library, we NEED to update our code, and having the import fail is a feature letting us know something needs to be re-written to match moto

thehesiod: if we include that library we're going to have some weird side-effect behavior of moto moves away from using the library

webknjaz: It's a direct dependency right now: https://github.com/aio-libs/aiobotocore/pull/1246#discussion_r1906234448. Pretending that it isn't wouldn't change the fact that it is, it'd only make it more confusing.

pyproject.toml
jakob-keller: > please don't resolve conversations started by reviewers, bad etiquette.

Sorry, I was not aware and will not do so again.

> Anywho this seems like it was added in [a73f8c2](https://github.com/aio-libs/aiobotocore/commit/a73f8c26192b190f503791b309124cb32eac44c8) for seemingly no reason. We should not be adding random requirements that aren't needed

It's a valid requirement: https://github.com/aio-libs/aiobotocore/blob/74323389b8fb6fcbe415f27765e7d7712f440050/tests/boto_tests/unit/test_signers.py#L75

> if it is needed it needs a comment as to why it's here

OK, I will add that comment referring to `test_signers.py`. My hesitation is due to the fact that those cross references do not scale well, i.e. for dev dependencies that are used in multiple or many modules.

jakob-keller: The comment is now included in this PR.

thehesiod: ok cool, thanks for that!

pyproject.toml
thehesiod: following rules for rules sake is fools errand, there will always be exceptions

jakob-keller: In the end, it should be @thehesiod who makes the call what qualifies as a dev dependency. I was working under the wrong assumption that we had reached consensus and that's why I jumped the gun in #1246. If you ask for it, I will revert dev dependencies to the previous state.

CONTRIBUTING.rst
jakob-keller: @thehesiod: Do you have an opinion on this topic, which IMO is the deciding factor for this PR? If there is also hesitation, we should just shelve the PR and I am OK with that outcome.

pyproject.toml
thehesiod: how about this, we pull werkzerg import from moto?  then it's not a direct dependency

jakob-keller: > how about this, we pull werkzerg import from moto? then it's not a direct dependency

I can live with that :-)

CONTRIBUTING.rst
thehesiod: I'd like to try it

thehesiod: this is not going to be the deciding factor for contributing believe me, so I wouldn't worry about it :)

pyproject.toml
thehesiod: cool once that's done I'll approve

jakob-keller: Well, we cannot switch to importing indirectly from moto, because:
1. `moto` uses the `from werkzeug.foo import bar` pattern.
2. In `moto_server.py`, we use `werkzeug.serving.generate_adhoc_ssl_context()`.
3. That function is not used in `moto==4.2.14`, i.e. that function does not become part of the moto namespace.

How does that change you assessment?

thehesiod (approved):

jakob-keller:
Friendly reminder for reviewers to close any open conversations. Thanks!

thehesiod:
hmm maybe I'll remove that req of unresolved conversations

CONTRIBUTING.rst
webknjaz: Personally, when I say “contributors”, I don't mean “core” but external non-maintainer folks. But alright. Marking this thread as resolved.

webknjaz:
> hmm maybe I'll remove that req of unresolved conversations

That's a good requirement to have. Keep it. It facilitates more transparency in comms.

PR #1288, opened 2025-02-09
trim usage of the old Python 2 vs 3 helper

a-detiste:
botocore only supports Python 3 anyway.

https://github.com/boto/botocore/blob/develop/botocore/compat.py

This showed up while doing an audit with 'Debian Code Search'

https://codesearch.debian.net/search?q=botocore.compat&literal=1&perpkg=1

https://wiki.debian.org/Python3-six-removal

### DIFF
--- a/aiobotocore/credentials.py
+++ b/aiobotocore/credentials.py
@@ -7,7 +7,6 @@ import subprocess
 from copy import deepcopy
 from hashlib import sha1
 
-import botocore.compat
 from botocore import UNSIGNED
 from botocore.compat import compat_shell_split
 from botocore.config import Config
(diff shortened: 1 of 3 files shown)


jakob-keller:
Thank you for proposing to modernize the `botocore` codebase! However, this is not the `botocore` repository. Please check out https://github.com/boto/botocore and offer your contribution there. It might be best to check whether they are open for this improvement, before putting in the effort for another PR. Good luck!

a-detiste:
Hi.

I already also proposed a much longer PR too. I reviewed all the 5 projects I could find in "Debian Code Search" using botocore.compat and sent PR for all.

Now I m asking to spend a few minutes to read this botocore.compat module and understand what it does: not much.

Well, not much _anymore_

At the times of Python2 vs 3 it did somewhat the same thing as "six" ... plus also import six itself. 

Importing json from botocore.compat has not any purpose left nowadays. 

Greetings

a-detiste:
So this PR is not improving botocore itself but more the ecosystem as a whole

thehesiod:
we ensure our codebase matches botocore for compatibility reasons, so anything non async related comes directly from botocore.

PR #1295, opened 2025-02-14
Exclude unit tests from coverage

jakob-keller:
### Description of Change
Exclude unit tests from coverage

### Assumptions
Reported Coverage should not include unit test code

  * [ ] Why needed


thehesiod:
we had this debate at work, we kept it because then it assures all your test code is being called as well

thehesiod:
there are cases where tests incorrectly aren't called and code coverage
would point that out. the issues with test code coverage can easily be
resolved with codecov exclusions.

PR #1297, opened 2025-02-15
bump moto

thehesiod:


tests/conftest.py
```diff
@@ -126,7 +126,7 @@ def alternative_region():
 
 @pytest.fixture
 def signature_version():
-    return 's3'
+    return 'v4'
```
thehesiod, line 129: this was critical to get the new ThreadedServer to recognize the correct service to route the request to

tests/boto_tests/unit/test_signers.py
```diff
@@ -41,15 +45,19 @@ async def test_signers_generate_db_auth_token(rds_client):
...
-    )
-    assert result2.startswith(
-        'prod-instance.us-east-1.rds.amazonaws.com:3306/?AWSAccessKeyId=xxx&'
+    expected_result = (
+        'prod-instance.us-east-1.rds.amazonaws.com:3306/?Action=connect'
+        '&DBUser=someusername&X-Amz-Algorithm=AWS4-HMAC-SHA256'
+        '&X-Amz-Date=20161107T173933Z&X-Amz-SignedHeaders=host'
+        '&X-Amz-Expires=900&X-Amz-Credential=akid%2F20161107%2F'
+        'us-east-1%2Frds-db%2Faws4_request&X-Amz-Signature'
+        '=d1138cdbc0ca63eec012ec0fc6c2267e03642168f5884a7795320d4c18374c61'
```
thehesiod, line 54: copied from the botocore test

pyproject.toml
```diff
@@ -71,18 +71,17 @@ botocore-dev = [
...
-    "pre-commit >=3.5.0,<5",
-    "pytest-asyncio >=0.23.8,<0.26",
-    "requests ~=2.32.3", # Used in test_version.py
-    "time-machine ~=2.15.0", # Used in test_signers.py
+    "pre-commit >= 3.5.0,<5",
+    "pytest-asyncio >= 0.23.8,<0.26",
+    "requests ~= 2.32.3", # Used in test_version.py
+    "time-machine ~= 2.15.0", # Used in test_signers.py
     "tomli; python_version<'3.11'", # Used in test_version.py
-    "werkzeug ~=3.0.6", # Used in moto_server.py
```
thehesiod, line 85: yay no more werkzeug ;)

thehesiod:
ok finally ready

thehesiod:
btw updated the merge to be squash and merge to avoid master pollution

jakob-keller:
> ok finally ready

I’ll try to do the review by tomorrow.

tests/test_basic_s3.py
```diff
@@ -174,14 +174,28 @@ async def test_result_key_iters(s3_client, bucket_name, create_object):
...
 
+    # now test checksum'd file
+    key_name = 'foobarbaz_cs'
+    await create_object(key_name, b'abcd', ChecksumAlgorithm="sha1")
+    resp = await s3_client.get_object(
+        Bucket=bucket_name, Key=key_name, ChecksumMode="ENABLED"
+    )
+    data = await resp['Body'].read()
+    assert data == b'abcd'
+    print()
```
jakob-keller, line 197: Probably left over and should be removed?

tests/test_basic_s3.py
```diff
@@ -174,14 +174,28 @@ async def test_result_key_iters(s3_client, bucket_name, create_object):
     assert 'CommonPrefixes' in response
 
 
-async def test_can_get_and_put_object(s3_client, create_object, bucket_name):
-    await create_object('foobarbaz', body='body contents')
-    resp = await s3_client.get_object(Bucket=bucket_name, Key='foobarbaz')
+async def test_can_get_and_put_object(
+    s3_client, create_object, bucket_name: str
```
jakob-keller, line 178: Why that single type annotation?

tests/mock_server.py
```diff
@@ -1,20 +1,32 @@
...
 
+host = '127.0.0.1'
+
+
+def get_free_tcp_port(release_socket: bool = False):
+    sckt = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sckt.bind((host, 0))
+    addr, port = sckt.getsockname()
+    if release_socket:
+        sckt.close()
```
jakob-keller, line 25: Amazon Q Developer pointed me to this issue: https://docs.aws.amazon.com/amazonq/detector-library/python/socket-close-platform-compatibility/

tests/conftest.py
```diff
@@ -468,8 +483,8 @@ def tempdir():
 
 
 @pytest.fixture
-def create_object(s3_client, bucket_name):
-    async def _f(key_name, body='foo'):
+def create_object(s3_client, bucket_name: str):
```
jakob-keller, line 486: Why add partial type annotations?

pyproject.toml
```diff
@@ -71,18 +71,17 @@ botocore-dev = [
     # "colorama>0.3.0", # Windows requirement
 ]
 dev = [
-    "dill ~=0.3.3", # Used in test_patches.py
-    "docker ~=7.1",
-    "docutils >=0.16,<0.22", # Used in test_version.py
-    "moto[server,s3,sqs,awslambda,dynamodb,cloudformation,sns,batch,ec2,rds] ~=4.2.9",
-    "packaging ~=24.1", # Used in test_version.py
+    "dill ~= 0.3.3", # Used in test_patches.py
```
jakob-keller, line 74: @webknjaz had suggested to adopt the sparse notation. Why revert this?

jakob-keller (changes requested):
Awesome, this one is much overdue!

I have a few minor questions and suggestions, but this should be easy to follow up on.

Thanks for the effort!

pyproject.toml
thehesiod: I just added whitespace to make it more readable, what do you mean by sparse notation, never heard of it.  PEP 440 suggests having spaces as well: https://peps.python.org/pep-0440/#version-specifiers

tests/test_basic_s3.py
thehesiod: because s3_client is a dynamic object and create_object is a function.  Just stepping into the right direction

tests/test_basic_s3.py
thehesiod: ah ya, from original commit from way back, thanks, removed

tests/mock_server.py
thehesiod: we're not using socket file descriptors, we're using stream...and never going to run these tests on windows ;)

pyproject.toml
jakob-keller: I don't care either way, just want to make sure we don't convert back and forth all the time, depending on who authors or does reviews.

jakob-keller (approved):

pyproject.toml
thehesiod: I care in that I want it readable :)

thehesiod:
thanks for the review! lemme c what I can do about this coverage

jakob-keller:
> thanks for the review! lemme c what I can do about this coverage

The merge is blocked due to missing verified commit signatures.

thehesiod:
will fix

thehesiod:
https://github.com/mattlqx/pre-commit-sign?

jakob-keller:
> @jakob-keller crap it's the pre-commit thing, how do we fix that thing so it's signed?

I guess the easiest way would be for you to squash the unverified commit and force push your branch.

jakob-keller:
> https://github.com/mattlqx/pre-commit-sign?

The unverified commit originates from https://pre-commit.ci. I don't see a way to make that spit out verified commits. Going forward, I guess we need to either drop pre-commit.ci or drop the verification requirement. I'd be in favor of the former.

thehesiod:
@jakob-keller ya I say drop the pre-commit from being able to modify the PR.  Failing is fine, then having the user fix it is better otherwise author may not notice something changing their code on their behalf potentially leading to bugs...like especially since we want to mostly keep formatting from botocore

PR #1301, opened 2025-02-24
upstream header fixes

thehesiod:
fix to https://github.com/aio-libs/aiobotocore/discussions/1300


aiobotocore/endpoint.py
```diff
@@ -13,7 +13,6 @@
     logger,
 )
 from botocore.hooks import first_non_none_response
-from urllib3.response import HTTPHeaderDict
```
jakob-keller, line 16: That removes our only direct urllib3 dependency and it could be dropped from pyproject.toml

thehesiod: done

thehesiod:
btw ran the local only tests as well and they passed

aiobotocore/endpoint.py
thehesiod: ya, that should have been an error, hmmm

aiobotocore/httpsession.py
jakob-keller, line 1: This would be a breaking change, right? `response.headers` would no longer accept case-insensitive keys. I am not sure If anyone relies on that, but it could be risky and might need to be documented and reflected in the version bump.

tests/test_lambda.py
```diff
@@ -41,7 +45,11 @@ def lambda_handler(event, context):
     return _process_lambda(lambda_src)
 
 
-async def test_run_lambda(iam_client, lambda_client, aws_lambda_zip):
```
jakob-keller, line 44: https://github.com/getmoto/moto/pull/8603 should take care of the root cause of our failing tests.

aiobotocore/httpsession.py
thehesiod: response.headers is a case insensitive dict (CIMultiDict), just that now we'd store them with their original case, if you go to AWSResponse it then wraps it in its own case insensitive dict ([HttpHeaders](https://github.com/boto/botocore/blob/develop/botocore/awsrequest.py#L460C14-L460C21))

tests/test_lambda.py
thehesiod: ok lemme re-lock, I'll keep the env var in case we need it later

thehesiod: ah 29 is not out yet, added cmt to remove once it's released

jakob-keller: I believe moto dropped support for Python 3.8 in the same release they shipped that fix. Since we still support Python 3.8, we might be blocked from upgrading for the time being.

aiobotocore/httpsession.py
thehesiod: basically we'd be preserving the case earlier and in more places, but it maintains being a case insensitive dict

tests/test_lambda.py
jakob-keller: > ok lemme re-lock, I'll keep the env var in case we need it later

Could you at least do that work in a separate PR? I don't think this is related to the original work.

thehesiod: py3.8 is no longer supported, we should drop it: https://devguide.python.org/versions/

jakob-keller: #1302 bumps dev dependencies, including moto. That takes care of failing tests, except for Python 3.8 on arm64.

thehesiod: we're kinda unique requirements as we require more things than botocore which could cause us to be a tighter version set, however meh, might as well keep it for now

jakob-keller: The underlying issue is not specific to GitHub CI: If you run tests locally with Python 3.8 on Apple silicon, you'd also be affected, i.e. 50% of runs would fail.

If you really want to address this, we should look at sys.version and something like https://stackoverflow.com/questions/71548156/how-to-get-architecture-of-running-python-interpreter-on-macos-with-apple-silico. 

Since Python 3.8 support will be dropped rather sooner than later, I'd say ignore it a tiny bit longer.

The light-weight fix would be to ignore platforms, and simply skip that test on Python 3.8 (via pytest decorator using sys.version).

In any case, I strongly suggest to open a separate PR.

Makefile
```diff
@@ -18,7 +18,6 @@ cov cover coverage: pre-commit
 
 mototest:
 	python -Wd -X tracemalloc=5 -X faulthandler -m pytest -vv -m "not localonly" -n auto --cov-report term --cov-report html --cov-report xml --cov=aiobotocore --cov=tests --log-cli-level=DEBUG $(FLAGS) aiobotocore tests
-	@echo "open file://`pwd`/htmlcov/index.html"
```
jakob-keller, line 21: Is this cosmetic, i.e. reduces clutter in standard output? Ideally, it would be in a separate PR ;-)

.github/workflows/ci-cd.yml
```diff
@@ -77,12 +77,15 @@ jobs:
       uses: astral-sh/setup-uv@v5
       with:
         python-version: ${{ matrix.python-version }}
+        enable-cache: true
```
jakob-keller, line 80: I believe this is unnecessary, because the setup-uv caches by default since https://github.com/astral-sh/setup-uv/releases/tag/v5.0.0. Also the default glob should be sufficient for our case.

Makefile
thehesiod: it's a security issue, we should not be calling "open" on a url during the build.  It's so minor not sure it warrants another PR

tests/test_lambda.py
thehesiod: I actually prefer it to fail when run locally as this needs to be resolved.  I only care during prod builds so it doesn't interrupt PRs from getting merged.  Given I needed it to get this PR to pass I suggest keeping here

.github/workflows/ci-cd.yml
thehesiod: cool, reverted, thanks

thehesiod: they should update the docs to stay the default is true, would have prevented me from going down this path

tests/test_lambda.py
jakob-keller: > I actually prefer it to fail when run locally as this needs to be resolved.
Well it won't, unless moto backports my PR and does another 4.0.x release with Python 3.8 support. But ok.

> I only care during prod builds so it doesn't interrupt PRs from getting merged. Given I needed it to get this PR to pass I suggest keeping here
That is incorrect: We treat arm jobs as experimental, i.e. they DO NOT BLOCK CI.

BTW, I am currently prepping a PR that "graduate" arm from experimental stage, except for Python 3.8. That way, if something else fails, it would block CI going forward.

jakob-keller: There we go: #1303

thehesiod: > BTW, I am currently prepping a PR that "graduate" arm from experimental stage, except for Python 3.8. That way, if something else fails, it would block CI going forward.

hmm well they were showing as failed here so if it didn't block CI was still confusing.  But better just not run it like you did

thehesiod: reverted per your PR

jakob-keller: > hmm well they were showing as failed here so if it didn't block CI was still confusing.

I agree, GitHub UI is confusing in that regard.

jakob-keller (approved):

thehesiod:
hmm, perhaps we should try publishing a dev version of this for people to test with?

jakob-keller:
> hmm, perhaps we should try publishing a dev version of this for people to test with?

Sounds reasonable, but I have no experience with it. Would you expect a significant amount of people to notice and participate?

jakob-keller:
As expected, the arm64 job on Python 3.8 is still flaky, but since we marked it as experimental, that does not block the PR. You could merge it now, if you like.

thehesiod:
> > hmm, perhaps we should try publishing a dev version of this for people to test with?
> 
> Sounds reasonable, but I have no experience with it. Would you expect a significant amount of people to notice and participate?

I'm thinking someone in that discussion group linked, had a post recently: https://github.com/aio-libs/aiobotocore/discussions/1300.  Let me change this to a dev version instead then

jakob-keller:
I

> > > hmm, perhaps we should try publishing a dev version of this for people to test with?
> > 
> > 
> > Sounds reasonable, but I have no experience with it. Would you expect a significant amount of people to notice and participate?
> 
> I'm thinking someone in that discussion group linked, had a post recently: #1300. Let me change this to a dev version instead then

They could get the wheel from https://github.com/aio-libs/aiobotocore/actions/runs/13508980065 without us having to publish to PyPI.

jakob-keller:
> Let me change this to a dev version instead then

Thinking about it, I don't think our CI will work for a dev release. For starters, the CI trigger would not even fire: https://github.com/aio-libs/aiobotocore/blob/7cb07f6f5c2048cd4d0edea1ccac72be3479c44b/.github/workflows/ci-cd.yml#L11

thehesiod:
> > Let me change this to a dev version instead then
> 
> Thinking about it, I don't think our CI will work for a dev release. For starters, the CI trigger would not even fire:
> 
> https://github.com/aio-libs/aiobotocore/blob/7cb07f6f5c2048cd4d0edea1ccac72be3479c44b/.github/workflows/ci-cd.yml#L11

fixed (theoretically)

jakob-keller (approved):

maresb:
I'm seeing a regression related to this change on some old code:

```
TypeError: object bytes can't be used in 'await' expression
```

triggered by https://github.com/aio-libs/aiobotocore/blob/3f15b219e0961a2611b507bf9577c10190bed6ba/aiobotocore/endpoint.py#L54-L55

I'm having some trouble with my debugger at the moment, so I'll try to provide more info once I get the chance to look around.

maresb:
That particular line has not changed for several years, but this PR modifies the code path immediately before it.

It seems that `else` is being used as an implicit fallback, and it's clearly not meant to handle bytes, but something leading up to that recently changed and is now allowing bytes in my particular case.

thehesiod:
ok if you could provide a testcase that would be greatly appreciated, thanks!

kjagiello:
I was able to reproduce it, but only when running aiobotocore under moto.

repro.py
```python
import asyncio
from moto import mock_aws

session = get_session()


async def main():
    with mock_aws():
        async with session.create_client("ec2", region_name="eu-west-1") as client:
            await client.describe_instances()


asyncio.run(main())
```

Run with `uv`:

```sh
uvx --with aiobotocore==2.21.1 --with moto==5.1.1 -- python repro.py
```

maresb:
Thanks @kjagiello that's a great MWE! The code I have also uses `moto`.

maresb:
Now that this regression has been verified by @kjagiello, I created issue #1320. Thanks so much @kjagiello!!!

PR #1308, opened 2025-02-27
Align with upstream

jakob-keller:
### Description of Change
During preparation for compatibility with botocore 1.37.x, I noticed and fixed a few discrepancies with upstream:
* fix class hierarchies of `AioDeferredRefreshableCredentials` and `AioSSOCredentialFetcher` to avoid unnecessary patches
* add missing patches to `test_patches.py`
* reduce minor drift in patched code (newlines, comments etc.)

### Assumptions
We strive to keep aiobotocore closely aligned with upstream.

  * [ ] Why needed


aiobotocore/credentials.py
```diff
@@ -376,11 +376,6 @@ def __init__(self, refresh_using, method, time_fetcher=_local_now):
         self.method = method
         self._frozen_credentials = None
 
-    def refresh_needed(self, refresh_in=None):
-        if self._frozen_credentials is None:
-            return True
-        return super().refresh_needed(refresh_in)
-
```
thehesiod, line 383: this method is still in botocore: https://github.com/boto/botocore/blob/1941b2b67bd033957c5ddd4855868fe7e010a174/botocore/credentials.py#L686-L689

aiobotocore/credentials.py
```diff
@@ -364,7 +362,9 @@ async def get_frozen_credentials(self):
         return self._frozen_credentials
 
 
-class AioDeferredRefreshableCredentials(AioRefreshableCredentials):
+class AioDeferredRefreshableCredentials(
+    DeferredRefreshableCredentials, AioRefreshableCredentials
+):
```
thehesiod, line 367: little value in inherting from parent class as effictively it's a re-write bug 🤷🏼  i guess useful in case they do an isinstance/issubclass on it

aiobotocore/credentials.py
jakob-keller: Yes, but since we fix the class hierarchy, we no longer need to patch it. We simply use upstream's implementation. It's non-blocking anyway.

aiobotocore/credentials.py
jakob-keller: In ned to differ: by subclassing, we avoid having to copy over a bunch of non-blocking code from upstream.

jakob-keller: Well, in this case "a bunch" is just the `refresh_needed()` below, but still.

thehesiod: I wonder if I left the parent out due to the parent having the method implemented and slicing or something 🤔

aiobotocore/credentials.py
thehesiod: oo, wait, why is this not async, hmm

thehesiod: ah ok, doesn't need to be async, yay, ignore!

aiobotocore/credentials.py
jakob-keller: I wondered, too, but couldn't find anything. It's really just copy and paste, but please double check!

thehesiod (approved):

jakob-keller:
@thehesiod Thanks for review, as always!

thehesiod:
> @thehesiod Thanks for review, as always!

thank you for all the hard work! 🚀  If it breaks we'll need to make sure to put a comment why it is the way it is.

Issue #1313, opened 2025-03-03
[regression] 2.21 introduces credential refresh assert


thehesiod:
```
    response = await dynamodb_client.get_item(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "aiobotocore/client.py", line 362, in _make_api_call
    ) = await self._resolve_endpoint_ruleset(
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "aiobotocore/client.py", line 510, in _resolve_endpoint_ruleset
    endpoint_info = await self._ruleset_resolver.construct_endpoint(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "aiobotocore/regions.py", line 24, in construct_endpoint
    provider_params = await self._get_provider_params(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "aiobotocore/regions.py", line 85, in _get_provider_params
    param_val = self._resolve_param_as_builtin(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "botocore/regions.py", line 626, in _resolve_param_as_builtin
    return builtin()
           ^^^^^^^^^
  File "botocore/credentials.py", line 350, in get_property
    return getattr(self, property_name, None)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "aiobotocore/credentials.py", line 305, in account_id
    raise NotImplementedError(
NotImplementedError: missing call to self._refresh. Use get_frozen_credentials instead
```

thehesiod:
repro'd

thehesiod:
suggest we expose a new `get_account_id` method that does the refresh, and replace this entry: https://github.com/boto/botocore/blob/develop/botocore/args.py#L821-L823.  Then replace `_resolve_param_as_builtin` probably in `AioEndpointRulesetResolver` and then we can make the call async in `_get_provider_params`.  I can work on this later today probably if no one gets to it

jakob-keller:
Fixed by #1314

PR #1314, opened 2025-03-04
fix refreshable credentials injection

thehesiod:
fixes issue #1313


thehesiod:
the base class impls may never get called, just there for correctness

jakob-keller (approved):
Nice and clean. Thanks for the fix!

PR #1315, opened 2025-03-05
Changes for botocore 1.37.2 support

dym-ok:


jakob-keller:
Please make sure to check out https://github.com/aio-libs/aiobotocore/discussions/1287. Bumping the botocore pin can be a deeply involved process, especially for 1.37.2 with its new serializers.

nalbion:
Can we get an update to https://pypi.org/project/botocore/1.37.25

PR #1325, opened 2025-04-13
Use SPDX license identifier for project metadata

cdce8p:
### Description of Change
PEP 639 specified the use of SPDX license expressions for `license` metadata.
Setuptools `v77` added full support for it. However that version requires Python `>=3.9`.

Until `3.8` is dropped, I believe it makes sense to just use the SPDX identifier with the old `license.text` field.

https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files

  * [ ] Why needed

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,9 +1,10 @@
 Changes
 -------
 
-2.22.0 (2025-04-09)
+2.22.0 (2025-04-15)
 ^^^^^^^^^^^^^^^^^^^
 * patch `ResponseParser`
+* use SPDX license identifier for project metadata
 
 2.21.1 (2025-03-04)
 ^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


jakob-keller (changes requested):
Thank you, that makes sense to me. Could you please do a rebase on main?

jakob-keller (changes requested):
Sorry, our approach to releasing and versioning is a bit weird. Could you please revert the version bump (2.22.0 is currently unreleased)?

jakob-keller:
This project requires verified commit signatures: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

Could you please make sure your commits can be verified by GitHub? Otherwise, if you want, I could re-commit the changes for you to pass that blocker.

cdce8p:
@jakob-keller Force pushed the commit to add the signature. Should be good now. Not sure why the one job is failing though. Looks unrelated.

PR #1323, opened 2025-04-07
Update botocore to 1.37.3

stj:
### Description of Change

Update the `botocore` to version 1.37.3. 
Includes changes to support for the smithy-rpc-v2-cbor protocol.

### Assumptions
*Replace this text with any assumptions made (if any)*

### Checklist for All Submissions
* [x] I have added change info to [CHANGES.rst](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.rst)
* [ ] If this is resolving an issue (needed so future developers can determine if change is still necessary and under what conditions) (can be provided via link to issue with these details):
  * [ ] Detailed description of issue
  * [ ] Alternative methods considered (if any)
  * [ ] How issue is being resolved
  * [ ] How issue can be reproduced
* [ ] If this is providing a new feature  (can be provided via link to issue with these details):
  * [ ] Detailed description of new feature
  * [ ] Why needed
  * [ ] Alternatives methods considered (if any)

### Checklist when updating botocore and/or aiohttp versions

* [x] I have read and followed [CONTRIBUTING.rst](https://github.com/aio-libs/aiobotocore/blob/master/CONTRIBUTING.rst#how-to-upgrade-botocore)
* [x] I have updated test_patches.py where/if appropriate (also check if no changes necessary)
* [x] I have ensured that the awscli/boto3 versions match the updated botocore version

PR #1312, opened 2025-03-01
Patch `ResponseParser`

jakob-keller:
### Description of Change
Patch `parsers.ResponseParser` to avoid duplicate code.

### Assumptions
*Replace this text with any assumptions made (if any)*

### Checklist for All Submissions
* [x] I have added change info to [CHANGES.rst](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.rst)
* [ ] If this is resolving an issue (needed so future developers can determine if change is still necessary and under what conditions) (can be provided via link to issue with these details):
  * [ ] Detailed description of issue
  * [ ] Alternative methods considered (if any)
  * [ ] How issue is being resolved
  * [ ] How issue can be reproduced
* [ ] If this is providing a new feature  (can be provided via link to issue with these details):
  * [ ] Detailed description of new feature
  * [ ] Why needed
  * [ ] Alternatives methods considered (if any)

### Checklist when updating botocore and/or aiohttp versions

* [ ] I have read and followed [CONTRIBUTING.rst](https://github.com/aio-libs/aiobotocore/blob/master/CONTRIBUTING.rst#how-to-upgrade-botocore)
* [ ] I have updated test_patches.py where/if appropriate (also check if no changes necessary)
* [ ] I have ensured that the awscli/boto3 versions match the updated botocore version

PR #1328, opened 2025-04-21
Create common interface for aio response classes

stj:
Having the `parse` method async on all parsers means the need for checking that `parse` is a coroutine is not needed anymore. The check for coroutine or not moved into the parser where a private method can be async or sync. This cleans up the public interface of the aio response classes.

### Description of Change

Change is extracted out of pull #1323 after a review comment ask for it.

### Assumptions

  * [ ] Why needed

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,8 @@ Changes
 ^^^^^^^^^^^^^^^^^^^
 * patch `ResponseParser`
 * use SPDX license identifier for project metadata
-* refactor `AioResponseParser` to have a consistent async implementation of the `parse` method.
+* refactor ``AioResponseParser`` to have a consistent async implementation of the `parse` method.
+  sync `parse` methods are deprecated and will be removed in a future release.``
 
 2.21.1 (2025-03-04)
 ^^^^^^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ Changes
 ^^^^^^^^^^^^^^^^^^^
 * patch `ResponseParser`
 * use SPDX license identifier for project metadata
+* refactor `AioResponseParser` to have a consistent async implementation of the `parse` method.
 
 2.21.1 (2025-03-04)
 ^^^^^^^^^^^^^^^^^^^
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -1,4 +1,6 @@
 import asyncio
+import inspect
+import warnings
 
 from botocore.endpoint import (
     DEFAULT_TIMEOUT,
@@ -204,9 +206,15 @@ class AioEndpoint(Endpoint):
         )
         parser = self._response_parser_factory.create_parser(protocol)
 
-        parsed_response = await parser.parse(
+        parsed_response = parser.parse(
             response_dict, operation_model.output_shape
         )
+        if inspect.isawaitable(parsed_response):
+            parsed_response = await parsed_response
+        else:
+            warnings.deprecated(
+                f"The use of non-async `parse` function on {parser.__class__.__name__} is deprecated."
+            )
         parsed_response.update(customized_response_dict)
 
         if http_response.status_code >= 300:
@@ -234,7 +242,13 @@ class AioEndpoint(Endpoint):
         if error_shape is None:
             return
 
-        modeled_parse = await parser.parse(response_dict, error_shape)
+        modeled_parse = parser.parse(response_dict, error_shape)
+        if inspect.isawaitable(modeled_parse):
+            modeled_parse = await modeled_parse
+        else:
+            warnings.deprecated(
+                f"The use of non-async `parse` function on {parser.__class__.__name__} is deprecated."
+            )
         # TODO: avoid naming conflicts with ResponseMetadata and Error
         parsed_response.update(modeled_parse)
 
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -204,15 +204,9 @@ class AioEndpoint(Endpoint):
         )
         parser = self._response_parser_factory.create_parser(protocol)
 
-        if asyncio.iscoroutinefunction(parser.parse):
-            parsed_response = await parser.parse(
-                response_dict, operation_model.output_shape
-            )
-        else:
-            parsed_response = parser.parse(
-                response_dict, operation_model.output_shape
-            )
-
+        parsed_response = await parser.parse(
+            response_dict, operation_model.output_shape
+        )
         parsed_response.update(customized_response_dict)
 
         if http_response.status_code >= 300:
@@ -240,10 +234,7 @@ class AioEndpoint(Endpoint):
         if error_shape is None:
             return
 
-        if asyncio.iscoroutinefunction(parser.parse):
-            modeled_parse = await parser.parse(response_dict, error_shape)
-        else:
-            modeled_parse = parser.parse(response_dict, error_shape)
+        modeled_parse = await parser.parse(response_dict, error_shape)
         # TODO: avoid naming conflicts with ResponseMetadata and Error
         parsed_response.update(modeled_parse)
 
--- a/aiobotocore/parsers.py
+++ b/aiobotocore/parsers.py
@@ -1,3 +1,5 @@
+import inspect
+
 from botocore.parsers import (
     LOG,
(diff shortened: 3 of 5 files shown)


aiobotocore/parsers.py
```diff
@@ -31,6 +33,43 @@ def _create_event_stream(self, response, shape):
...
+            elif self._is_modeled_error_shape(shape):
+                parsed = self._do_modeled_error_parse(response, shape)
+                # We don't want to decorate the modeled fields with metadata
+                return parsed
+            else:
+                parsed = self._do_error_parse(response, shape)
+        else:
+            parsed = self._do_parse(response, shape)
+            if inspect.iscoroutine(parsed):
+                parsed = await parsed
```
stj, line 53: The `iscoroutinefunction` checks previously in `endpoint.py` are handled here now.

jakob-keller:
Could you please take a look here, too?

https://github.com/aio-libs/aiobotocore/blob/d7c109294b69e5d286b2270dd318b7a730fb2e7f/aiobotocore/response.py#L152-L158

We should be able to avoid the check there as well.

aiobotocore/parsers.py
```diff
@@ -31,6 +33,43 @@ def _create_event_stream(self, response, shape):
         name = response['context'].get('operation_name')
         return AioEventStream(response['body'], shape, parser, name)
 
+    # Only JSONParser needs this to be async. Override it here at the base
```
jakob-keller, line 36: Feel free to remove the comment

aiobotocore/parsers.py
jakob-keller: I love that you switched from deprecated `asyncio.iscoroutinefunction` to `inspect.iscoroutine`.

stj: I'll try to understand the difference of `iscouroutine` and `isawaitable` now. The helpers use `isawaitable`.
If I understand the [documentation](https://docs.python.org/3/library/inspect.html#inspect.iscoroutine) correct, `isawaitable` support functions decorated with `types.coroutine` and aren't defined by `async def`. If I understand that right, `isawaitable` covers a larger set of functions. 

I am going to change it to `_helpers.resolve_awaitable`.

jakob-keller:
Looks good to me! Would you mind adding an entry to CHANGES.rst?

stj:
How does 

> refactor `AioResponseParser` to have a consistent async implementation of the `parse` method.

sound for a changelog entry? It communicates a change and where it is, though doesn't say what the implications are for aiobotocore users. Is `parse` a public interface?

jakob-keller:
> How does
> 
> > refactor `AioResponseParser` to have a consistent async implementation of the `parse` method.
> 
> sound for a changelog entry?

Sounds good to me.

> Is `parse` a public interface?

I'd say, it might, but stick with the 2.22.0 version bump.

jakob-keller (dismissed):

aiobotocore/endpoint.py
```diff
@@ -204,15 +204,9 @@ async def _do_get_response(self, request, operation_model, context):
...
-                response_dict, operation_model.output_shape
-            )
-        else:
-            parsed_response = parser.parse(
-                response_dict, operation_model.output_shape
-            )
-
+        parsed_response = await parser.parse(
+            response_dict, operation_model.output_shape
+        )
```
thehesiod, line 209: have we validated across all the parsers this is now true?  Are there parsers that don't inherit from our baseclass?

thehesiod: i see everything parents to ResponseParser 👍🏼

jakob-keller: Good call!

Upstream has a number of parsers that we have not patched:
* `BaseXMLResponseParser`
* `BaseJSONParser`
* `BaseEventStreamParser`
* `EventStreamJSONParser`
* `EventStreamXMLParser`
* `BaseRestParser`

The `Base*` parsers should be fine, since we patch any relevant child parsers. But I believe the `EventStream*` parsers might be problematic. They are used indirectly and might need to be patched to work correctly.

jakob-keller: I could prepare a follow-up of #1312 to fully patch the parser class hierarchy. That would be a prerequisite for this PR, IMO. Any thoughts?

jakob-keller: Work-in-progress: #1329

jakob-keller: #1329 makes sure that `EVENT_STREAM_PARSER_CLS` class attributes ("indirect" uses) point to patched parsers. Once that is merged, we can continue with this one.

stj: Interesting, I didn't look at `EVENT_STREAM_PARSER_CLS` and how they are used and inherited. If I understand them right, they are based on the sync implementation for parser. `aiobotocore` has no async version of them, probably because these event parsers don't do I/O, right? As far as I understand, they are used by `AioEventStream` to parse each received "chunk". My belief atm is that not having an async version for `EventStream*` is fine. 

Just the classes in the `PROTOCOL_PARSERS` mapping are changed here to provide a consistent async interface for the `parser` method. What would be the benefit of fully patching the parser class hierarchy?

jakob-keller: > My belief atm is that not having an async version for `EventStream*` is fine.

I guess it would be, as long as we continue to allow for both sync and async variants of `parse()`. However, this PR takes away that flexibility and we need to be careful to handle all parser classes, as @thehesiod rightfully pointed out. As far as I understand, that includes parser classes that we have not patched previously, hence #1329. I could well be on the wrong track here, though.

An alternative would be to simply keep the current implementation of `parse()` and move it up the `AioResponseParser`, but leave everything else as it is.

thehesiod: ya that's why I was waiting to do a review of this PR when I had some time, we need to thoroughly look at how these classes are used

thehesiod: lets report back when one of us does that

stj: How about allowing sync `parse` methods but emit a deprecation warning?

thehesiod: sounds like a plan, would be safer

CHANGES.rst
```diff
@@ -5,6 +5,7 @@ Changes
 ^^^^^^^^^^^^^^^^^^^
 * patch `ResponseParser`
 * use SPDX license identifier for project metadata
+* refactor `AioResponseParser` to have a consistent async implementation of the `parse` method.
```
webknjaz, line 8: It's RST
```suggestion
* refactor ``AioResponseParser`` to have a consistent async implementation of the ``parse`` method.
```

aiobotocore/endpoint.py
```diff
@@ -204,15 +206,15 @@ async def _do_get_response(self, request, operation_model, context):
...
+        parsed_response = parser.parse(
+            response_dict, operation_model.output_shape
+        )
+        if inspect.isawaitable(parsed_response):
+            parsed_response = await parsed_response
         else:
-            parsed_response = parser.parse(
-                response_dict, operation_model.output_shape
+            warnings.deprecated(
+                f"The use of non-async `parse` function on {parser.__class__.__name__} is deprecated."
```
jakob-keller, line 216: Why not use `type(parser).__name__`?

aiobotocore/response.py
```diff
@@ -150,10 +150,5 @@ async def get_response(operation_model, http_response):
         response_dict['body'] = await http_response.content
 
     parser = parsers.create_parser(protocol)
-    if asyncio.iscoroutinefunction(parser.parse):
-        parsed = await parser.parse(
-            response_dict, operation_model.output_shape
-        )
-    else:
-        parsed = parser.parse(response_dict, operation_model.output_shape)
+    parsed = await parser.parse(response_dict, operation_model.output_shape)
```
jakob-keller, line 153: Apply deprecation pattern from `endpoints.py`

tests/test_patches.py
jakob-keller, line 1: That test was slightly refactored in the default branch. Please do a rebase or merge in new commits.

jakob-keller (changes requested):

jakob-keller:
@stj: I discovered a few more quirks with regards to event streams and decided to created #1347 which is meant to supersede this PR. I hope you don't mind and thanks again for the work!

jakob-keller:
Superseded by #1347

PR #1330, opened 2025-04-23
Prevent swallowing of test cases

### LINKED ISSUE #1316: `_API_DIGESTS ` in `test_patches.py` swallowed test cases
**Describe the bug**
`test_patches.py` maintains a collection of upstream code that is patched by aiobotocore:
https://github.com/aio-libs/aiobotocore/blob/d145aea37ee05fff16a71c7c420f703aa794440d/tests/test_patches.py#L142

The patched objects are used as dictionary keys, which results in multiple test cases being swallowed due to hash collisions, e.g.:
https://github.com/aio-libs/aiobotocore/blob/d145aea37ee05fff16a71c7c420f703aa794440d/tests/test_patches.py#L427-L438

Solution: Either those redundant entries are merged or the dict is replaced, i.e. by a tuple of tuples.

**Checklist**
- [ ] I have reproduced in environment where `pip check` passes without errors
- [ ] I have provided `pip freeze` results
- [ ] I have provided sample code or detailed way to reproduce
- [ ] I have tried the same code in botocore to ensure this is an aiobotocore specific issue
- [ ] I have tried similar code in aiohttp to ensure this is is an aiobotocore specific issue
- [ ] I have checked the latest and older versions of aiobotocore/aiohttp/python to see if this is a regression / injection

**pip freeze results**

**Environment:**
 - Python Version: [e.g. 3.6]
 - OS name and version: [e.g. darwin stretch]

**Additional context**

jakob-keller:
### Description of Change
Replace `_API_DIGESTS` dict by tuple to avoid hash collisions resulting in swallowed test cases.

### Assumptions
None

* [ ] If this is resolving an issue (needed so future developers can determine if change is still necessary and under what conditions) (can be provided via link to issue with these details): closes #1316
  * [ ] Why needed


jakob-keller:
c9ca006cf03382e6feba1d0c74be3d0b1924823e demonstrates the need for the PR: The hash was outdated, but did not fail the test because it was shadowed by the following test case.

tests/test_patches.py
```diff
@@ -801,7 +1583,7 @@ def test_patches():
     print(f"Botocore version: {botocore.__version__}")
 
     success = True
-    for obj, digests in _API_DIGESTS.items():
+    for obj, digests in _API_DIGESTS:
```
thehesiod, line 1586: can you add an assert that we don't encounter the same item twice, also a comment about why this is a tuple and not dict, otherwise someone may try converting it back to a dict later

thehesiod: cmt should go at top of declaration

thehesiod:
I had similar issue with two tests with the same name in the same file, perhaps add pytest-error-for-skips or ensure linter catches this case too.  Ideally this particular case would also be caught by linter..obviously wouldn't work if it was generated dynamically.  Thinking about it I think we should add linting to ensure literal dict keys in general aren't declared more than once in case this type of issue is anywhere else in our code

tests/test_patches.py
jakob-keller: The test now uses pytest parametrization. Does that address your concerns?

thehesiod (approved):

PR #1344, opened 2025-04-25
Zizmor integration

### LINKED ISSUE #1334: Add zizmor GitHub workflow
Might be a good idea to add a zizmor workflow..

_Originally posted by @webknjaz in https://github.com/aio-libs/aiobotocore/issues/1333#issuecomment-2829751866_

jakob-keller:
### Description of Change
Integrate zizmor via GitHub Actions 

* https://woodruffw.github.io/zizmor/usage/#use-in-github-actions

### Assumptions
None

* [x] If this is resolving an issue (needed so future developers can determine if change is still necessary and under what conditions) (can be provided via link to issue with these details): closes #1334 
  * [ ] Why needed

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -36,8 +36,9 @@ jobs:
       # yamllint disable-line rule:line-length
       uses: astral-sh/setup-uv@c7f87aa956e4c323abf06d5dec078e358f6b4d04  # v6.0.0
       with:
-        # prevent cache poisoning
-        enable-cache: ${{ github.ref_type == 'tag' && 'false' || 'auto' }}
+        # yamllint disable-line rule:line-length
+        enable-cache: | # zizmor: ignore[cache-poisoning] cache is disabled when publishing to prevent poisoning
+          ${{ github.ref_type == 'tag' && 'false' || 'auto' }}
     - name: Build distribution 📦
       run: uv build
     - name: Check distribution 📦
@@ -90,8 +91,9 @@ jobs:
       uses: astral-sh/setup-uv@c7f87aa956e4c323abf06d5dec078e358f6b4d04  # v6.0.0
       with:
         python-version: ${{ matrix.python-version }}
-        # prevent cache poisoning
-        enable-cache: ${{ github.ref_type == 'tag' && 'false' || 'auto' }}
+        # yamllint disable-line rule:line-length
+        enable-cache: | # zizmor: ignore[cache-poisoning] cache is disabled when publishing to prevent poisoning
+          ${{ github.ref_type == 'tag' && 'false' || 'auto' }}
     - name: Run pre-commit hooks
       run: |
         uv run make pre-commit
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -114,11 +114,17 @@ jobs:
         fail_ci_if_error: true  # optional (default = false)
         verbose: true  # optional (default = false)
 
+  zizmor:
+    uses: ./.github/workflows/reusable-zizmor.yml
+    permissions:
+      security-events: write
+
   check:  # This job does nothing and is only used for the branch protection
     if: always()
     needs:
     - build
     - test
+    - zizmor
     runs-on: ubuntu-24.04
     timeout-minutes: 5
 
--- a/.github/workflows/reusable-zizmor.yml
+++ b/.github/workflows/reusable-zizmor.yml
@@ -1,0 +1,39 @@
+---
+name: Reusable zizmor
+
+permissions: {}
+
+on:
+  workflow_call:
+
+env:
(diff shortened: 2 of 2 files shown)


jakob-keller:
@webknjaz: Should we make this part of the allsgreen check? If so, fold the workflow into the existing ci-cd workflow?

webknjaz (commented):
> @webknjaz: Should we make this part of the allsgreen check? If so, fold the workflow into the existing ci-cd workflow?

Didn't think about it. But perhaps, it's a good idea. Though, there's merge queue and I don't know if we should be running the check with those. I guess, we'll find out.

Regarding adding more jobs, while integrating alls-green into CPython, I had to deal with multiple workflows. And I ended up coming up with a modularization idea — different jobs (but not matrixes)  can go into in-repo reusable workflows. Their names are prefixed with `reusable-` and they are only allowed to have the `workflow_call` trigger setup. Then, the main workflow has all the real triggers + defines relationships between the jobs and matrixes on the high level. This top-level definition also contains alls-green and pypi-publish (due to trusted publishing limitations, it's not in a reusable workflow).

Having been using this approach for almost 2 years now, I find it a really good structure as it allows separating the abstraction layers which is good for the organization.

jakob-keller:
> > @webknjaz: Should we make this part of the allsgreen check? If so, fold the workflow into the existing ci-cd workflow?
> 
> Didn't think about it. But perhaps, it's a good idea. Though, there's merge queue and I don't know if we should be running the check with those. I guess, we'll find out.
> 
> Regarding adding more jobs, while integrating alls-green into CPython, I had to deal with multiple workflows. And I ended up coming up with a modularization idea — different jobs (but not matrixes) can go into in-repo reusable workflows. Their names are prefixed with `reusable-` and they are only allowed to have the `workflow_call` trigger setup. Then, the main workflow has all the real triggers + defines relationships between the jobs and matrixes on the high level. This top-level definition also contains alls-green and pypi-publish (due to trusted publishing limitations, it's not in a reusable workflow).
> 
> Having been using this approach for almost 2 years now, I find it a really good structure as it allows separating the abstraction layers which is good for the organization.

I like the modular design and applied it to the best of my knowledge. That should have been a separate PR, but ¯\_(ツ)_/¯

webknjaz:
Yeah, I thought you'd just add the new workflow that way and separate the rest...

I think you should resubmit this PR from a fork (actually, I think any topic branches should live outside upstream but that's a separate topic).

I'm concerned about most of the jobs not having their permissions settings. And I think they might have higher privileges within upstream but when a contributor comes in, they'd break. So I'd rather test that right away.

jakob-keller:
> Yeah, I thought you'd just add the new workflow that way and separate the rest...

Reverted all but the new workflow.

> I think you should resubmit this PR from a fork (actually, I think any topic branches should live outside upstream but that's a separate topic).

I thought, I was doing that all along. Not sure what you mean?

> I'm concerned about most of the jobs not having their permissions settings. And I think they might have higher privileges within upstream but when a contributor comes in, they'd break. So I'd rather test that right away.

I'm afraid, I don't follow you. But that relates to the other jobs, right?

webknjaz:
Oh, I got confused reviewing from mobile. You're right. No need to resubmit.

webknjaz:
> I'm concerned about most of the jobs not having their permissions settings. And I think they might have higher privileges within upstream but when a contributor comes in, they'd break. So I'd rather test that right away.\n\nI'm afraid, I don't follow you. But that relates to the other jobs, right?

I thought that maybe that build job might have a difficulty uploading artifacts. But if the CI doesn't fail, than it's fine.

webknjaz (approved):

webknjaz:
@jakob-keller I did something experimental, and you can now stop hosting an in-tree reusable workflow:
```yaml
---

name: GitHub Actions Security Analysis with zizmor 🌈

on:  # yamllint disable-line rule:truthy
  push:
  pull_request:

jobs:
  zizmor:
    name: 🌈 zizmor

    permissions:
      security-events: write

    uses: zizmorcore/workflow/.github/workflows/reusable-zizmor.yml@3bb5e95068d0f44b6d2f3f7e91379bed1d2f96a8

...
```

jakob-keller:
> @jakob-keller I did something experimental, and you can now stop hosting an in-tree reusable workflow:
> 
> ```yaml
> ---
> 
> name: GitHub Actions Security Analysis with zizmor 🌈
> 
> on:  # yamllint disable-line rule:truthy
>   push:
>   pull_request:
> 
> jobs:
>   zizmor:
>     name: 🌈 zizmor
> 
>     permissions:
>       security-events: write
> 
>     # yamllint disable-line rule:line-length
>     uses: zizmorcore/workflow/.github/workflows/reusable-zizmor.yml@1ae473d8672fe7613e809d86d202a35063736e16
> 
> ...
> ```

Picked this up in #1355

PR #1345, opened 2025-04-25
Refactor build and test CI/CD jobs into reusable workflows

jakob-keller:
### Description of Change
Refactor build and test CI/CD jobs into reusable workflows (follow on of #1344)

### Assumptions
None

  * [ ] Why needed

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -24,35 +24,9 @@ env:
 
 jobs:
   build:
-    name: Build distribution 📦
-    runs-on: ubuntu-24.04
-    timeout-minutes: 5
-
-    steps:
-    - uses: actions/checkout@v4
-      with:
-        persist-credentials: false
-    - name: Install uv
-      # yamllint disable-line rule:line-length
-      uses: astral-sh/setup-uv@c7f87aa956e4c323abf06d5dec078e358f6b4d04  # v6.0.0
-      with:
-        # yamllint disable-line rule:line-length
-        enable-cache: | # zizmor: ignore[cache-poisoning] cache is disabled when publishing to prevent poisoning
-          ${{ github.ref_type == 'tag' && 'false' || 'auto' }}
-    - name: Build distribution 📦
-      run: uv build
-    - name: Check distribution 📦
-      run: uvx twine check --strict dist/*
-    - name: Upload distribution 📦
-      uses: actions/upload-artifact@v4
-      with:
-        name: python-package-distributions
-        path: dist/
+    uses: ./.github/workflows/reusable-build.yml
 
   test:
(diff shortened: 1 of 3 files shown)


.github/workflows/reusable-test.yml
```diff
@@ -0,0 +1,76 @@
...
+env:
+  FORCE_COLOR: 1
+
+jobs:
+  test:
+    name: Test Python ${{ matrix.python-version }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.experimental }}
+    strategy:
+      matrix:
```
webknjaz, line 18: Don't put the matrix inside the reusable workflows, keep it in the top-level one and pass params as workflow inputs.

webknjaz (commented):
This idea helps separating data from logic and keep all the params in one place.

.github/workflows/reusable-test.yml
```diff
@@ -0,0 +1,68 @@
...
+on:
+  workflow_call:
+    inputs:
+      python-version:
+        type: string
+        required: true
+      os:
+        type: string
+        required: true
+      continue-on-error:
```
webknjaz, line 15: This is an implementation detail so I started calling it `xfail` to match pytest's semantics. Could be `allow-failures` or something like that.

webknjaz (approved):
Just one optional tip.

.github/workflows/reusable-test.yml
jakob-keller: I chose `continue-on-error` because it matches [GitHub semantics](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error).

.github/workflows/ci-cd.yml
```diff
@@ -24,35 +24,9 @@ env:
...
-      run: uvx twine check --strict dist/*
-    - name: Upload distribution 📦
-      uses: actions/upload-artifact@v4
-      with:
-        name: python-package-distributions
-        path: dist/
+    uses: ./.github/workflows/reusable-build.yml
 
   test:
-    name: Test Python ${{ matrix.python-version }} on ${{ matrix.os }}
```
webknjaz, line 53: @jakob-keller I missed this earlier, but the name should be set statically. Something like
```yaml
    name: 🧪 Tests${{ '' }}  # nest jobs under the same sidebar category
```
https://github.com/aio-libs/aiosignal/blob/0d5691b/.github/workflows/ci-cd.yml#L232

This will make sure that all the matrix jobs are grouped together in the left sidebar of the CI logs view. Currently, each of the matrix jobs is nested under a dedicated category, unnecessarily.

jakob-keller: #1354 addresses this

PR #1347, opened 2025-04-26
Patch `ResponseParser.parse()` and others to prepare for upstream bump

jakob-keller:
### Description of Change
Patch `ResponseParser.parse()` as well as `ClientArgsCreator.get_client_args()`, `Endpoint.__init__()`, and `EventStream._parse_event()`  in preparation for #1323.

Based on #1328 by @stj 

### Assumptions
* Thanks to #1329, we may be sure that all parsers have been patched. 
* `parsers.AioResponseParser` and many of its subclasses have not yet been released to PyPI. It is therefore considered safe to patch `.parse()` and remove checks in calling codepaths.

* [x] I have added change info to [CHANGES.rst](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.rst): implied in first item for pending release 2.22.0
  * [ ] Why needed

### DIFF
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -226,16 +226,9 @@ class AioEndpoint(Endpoint):
             customized_response_dict=customized_response_dict,
         )
         parser = self._response_parser_factory.create_parser(protocol)
-
-        if asyncio.iscoroutinefunction(parser.parse):
-            parsed_response = await parser.parse(
-                response_dict, operation_model.output_shape
-            )
-        else:
-            parsed_response = parser.parse(
-                response_dict, operation_model.output_shape
-            )
-
+        parsed_response = await parser.parse(
+            response_dict, operation_model.output_shape
+        )
         parsed_response.update(customized_response_dict)
 
         if http_response.status_code >= 300:
@@ -262,11 +255,7 @@ class AioEndpoint(Endpoint):
         error_shape = service_model.shape_for_error_code(error_code)
         if error_shape is None:
             return
-
-        if asyncio.iscoroutinefunction(parser.parse):
-            modeled_parse = await parser.parse(response_dict, error_shape)
-        else:
-            modeled_parse = parser.parse(response_dict, error_shape)
+        modeled_parse = await parser.parse(response_dict, error_shape)
         # TODO: avoid naming conflicts with ResponseMetadata and Error
         parsed_response.update(modeled_parse)
 
--- a/aiobotocore/endpoint.py
+++ b/aiobotocore/endpoint.py
@@ -16,6 +16,7 @@ from botocore.hooks import first_non_none_response
 
 from aiobotocore.httpchecksum import handle_checksum_body
 from aiobotocore.httpsession import AIOHTTPSession
+from aiobotocore.parsers import AioResponseParserFactory
 from aiobotocore.response import StreamingBody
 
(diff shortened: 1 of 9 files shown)


aiobotocore/endpoint.py
```diff
@@ -203,16 +203,9 @@ async def _do_get_response(self, request, operation_model, context):
...
         parser = self._response_parser_factory.create_parser(protocol)
-
-        if asyncio.iscoroutinefunction(parser.parse):
-            parsed_response = await parser.parse(
-                response_dict, operation_model.output_shape
-            )
-        else:
-            parsed_response = parser.parse(
-                response_dict, operation_model.output_shape
-            )
```
thehesiod, line 214: just make sure all possible parsers will be patched now.  Previously I had it to limit the scope of which parsers were async to those actually needed.  From what I see parsers can be created also here: https://github.com/aio-libs/aiobotocore/blob/master/aiobotocore/args.py#L97 so that will need to be fixed as well.  I think it would be safer to use my `resolve_awaitable`. btw theoretically people could pass their own `response_parser_factory` arg to the `class Endpoint` constructor as well.

jakob-keller: > just make sure all possible parsers will be patched now.

To the best of my knowledge, they all are. I just triple-checked all calls to `.parse()` and `._parse_event()` in upstream as well as aiobotocore and am very confident now.

> Previously I had it to limit the scope of which parsers were async to those actually needed.

That worked well and is a good design, as long as the class hierarchy remains small with few types that need patching. Recent updates to upstream introduced multiple new classes, some of which require asynchronous patches, which leads me to conclude that it will be easier to switch to an async-only API going forward. That's also the way most other functions are patched, right?

> From what I see parsers can be created also here: https://github.com/aio-libs/aiobotocore/blob/master/aiobotocore/args.py#L97 so that will need to be fixed as well.

True and the type of parser created there is picked from the patched `PROTOCOL_PARSERS` mapping. This is safe, unless that mapping is modified.

> btw theoretically people could pass their own `response_parser_factory` arg to the `class Endpoint` constructor as well.

They could and would simply need to adopt our new API, i.e. it's a breaking change. I find it reasonable to expect to provide an AioResponseParser when working with low-level aiobotocore primitives.

thehesiod: it's' not because it's calling the version in botocore which is not patched, we need to call the version of that API in our file: https://github.com/aio-libs/aiobotocore/blob/fbfeb1242ab1f378f6d4fe84b5d3dacdd8874189/aiobotocore/parsers.py#L23

jakob-keller: All calls to `create_parser` are already patched, if you're worried about that. Otherwise, I don't understand.

thehesiod: this is not patched: https://github.com/aio-libs/aiobotocore/blob/master/aiobotocore/args.py#L97

that line must be replaced with:
`response_parser = aiobotocore.parsers.create_parser(protocol)`.  Note it's currently `botocore.` not `aiobotocore.`

jakob-keller: Mea culpa, you were absolutely right!

thehesiod (approved):

jakob-keller:
@thehesiod: Thanks for the thorough review, as always!

thehesiod:
> @thehesiod: Thanks for the thorough review, as always!

my pleasure, thanks for all the hard work!

PR #1350, opened 2025-05-01
Bump `uv` dev dependency

jakob-keller:
### Description of Change
Bump `uv` dev dependency

### Assumptions
None

  * [ ] Why needed


.pre-commit-config.yaml
```diff
@@ -22,7 +22,7 @@ repos:
     args: [--fix]
   - id: ruff-format
 - repo: https://github.com/astral-sh/uv-pre-commit
-  rev: 0.6.17
+  rev: 0.7.2
```
thehesiod, line 25: if we don't specify the version the default behavior is to pick the version from the pyproject.toml file IIRC

pyproject.toml
```diff
@@ -87,7 +87,7 @@ default-groups = [
     "botocore-dev",
     "dev",
 ]
-required-version = ">=0.6.15"
+required-version = ">=0.7.0"
```
thehesiod, line 90: safer would be like `~=0.7` or if more paranoid `~=0.7.0`

.pre-commit-config.yaml
jakob-keller: I believe this is referencing / pinning a git tag in that repo. We should keep that convention to maintain reproducibility. Full reproducibility would require pinning a git commit instead of a tag, but that would be overkill IMO.

pyproject.toml
jakob-keller: Paranoid? Count me in :-)

thehesiod (approved):

PR #1351, opened 2025-05-02
Remove Python 3.8 support

stj:
Python 3.8 has reached EOL.

### Description of Change

Drop Python 3.8 from CI and official supported versions.

### Context on why 

While working on adding support for [user agent 2.1 support ](https://github.com/boto/botocore/pull/3389/), existing tests started to fail with  

> <ContextVar name='_context' at 0x....> at 0x...> was created in a different Context

My belief is that this comes from how pytest is managing contextvars and tasks. Locall testing has shown that [these changes](https://github.com/pytest-dev/pytest-asyncio/commits/main/?since=2024-12-11&until=2024-12-12) fix this. Though, the `pytest-asyncio` release that has this change dropped support for py3.8 .

### Assumptions

Not sure what the [python3.8](https://github.com/aio-libs/aiobotocore/tree/master/tests/python3.8) folder under `tests` is for. I assume tests under it requires at least Python 3.8. Maybe this PR can move tests under it and remove the directory. 

  * [ ] Why needed


pyproject.toml
```diff
@@ -76,7 +75,7 @@ dev = [
     "packaging >= 24.1, < 26", # Used in test_version.py
     "pip >= 24.3.1, < 26", # Used in test_version.py
     "pre-commit >= 3.5.0, < 5",
-    "pytest-asyncio >= 0.23.8, < 0.27",
+    "pytest-asyncio >= 0.25.0, < 0.27",
```
stj, line 78: These version bumps aren't needed to drop support for 3.8.
I can remove them if it is desired to keep the change minimal.

jakob-keller:
We tend to follow upstream, i.e. drop support for Python 3.8 when we bump to botocore 1.38.0. Unless @thehesiod disagrees, I would suggest we defer work on this PR until then.

stj:
> We tend to follow upstream, i.e. drop support for Python 3.8 when we bump to botocore 1.38.0. Unless @thehesiod disagrees, I would suggest we defer work on this PR until then.

I get it and was expecting pushback. That's why I wrote some reasoning in the PR description. Basically while [working to support 1.37.4](https://github.com/stj/aiobotocore/tree/update-botocore-1.37.4) I run into issues that I narrowed down to a behaviour in pytest-asyncio. It is fixed in a more recent version, but that version dropped support for 3.8.

stj:
> Would it be possible to mark those tests with skip or xfail for Python 3.8?

Feels odd because a large amount of tests (~60) will not run under 3.8 anymore. 
It requires different dependencies per Python version to work.
Something like 
```
"pytest-asyncio >= 0.25.0, < 0.27;python_version<'3.9'"
"pytest-asyncio >= 0.26.0, < 0.27;python_version>='3.9'"
```
No clue if that is supported in `pyproject.toml` and the tooling that reads from it.
I can do some research on it.

stj:
I just noticed that the commit I referenced to pytest-asyncio is one of many commits around improving contextvars support in pytest-asyncio.
https://github.com/pytest-dev/pytest-asyncio/commits/main/?since=2024-12-11&until=2024-12-12

aiobotocore/httpsession.py
```diff
@@ -3,7 +3,7 @@
 import io
 import os
 import socket
-from typing import Dict, Optional
+from typing import Optional
```
thehesiod, line 6: technically optional is not needed as well, can do `| None`

pyproject.toml
thehesiod: if you want I'd just do it the latest, like, `~= 0.26.0`

thehesiod: basically we lock these so people don't have to figure out when something breaks when it's simply re-locked.  But if someone is actively looking at it feel free to update to latest.

aiobotocore/httpsession.py
thehesiod: ahh, `| None` was added in 3.10 so can't do that yet

thehesiod (approved):
I think looks good, with the fact botocore is finally getting rid of 3.8 I think we're good to go.

pyproject.toml
```diff
@@ -57,14 +56,14 @@ botocore-dev = [
     # "wheel==0.43.0",
     # "behave==1.2.5",
     # "jsonschema==4.21.1",
-    "coverage==7.2.7",
+    "coverage==7.5.4",
```
jakob-keller, line 59: Could you leave this in sync with upstream, please?

pyproject.toml
```diff
@@ -57,14 +56,14 @@ botocore-dev = [
...
     # "setuptools==71.1.0;python_version>='3.12'",
     # "packaging==24.1;python_version>='3.12'", # Requirement for setuptools>=71
 
     # Pytest specific deps
-    "pytest==8.1.1",
+    "pytest==8.3.5",
```
jakob-keller, line 64: I believe we should keep this in sync with upstream, too.

pyproject.toml
```diff
@@ -57,14 +56,14 @@ botocore-dev = [
...
     "pytest-cov==5.0.0",
-    "pytest-xdist==3.5.0",
```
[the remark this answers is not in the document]
thehesiod: lets put a cmt in this block so we remember to keep it in sync...either that or add a test that ensures it

jakob-keller (approved):
Thanks for the PR! I'm still not familiar with the test suite and can't fully assess changes there, but the rest LGTM!

jakob-keller:
Looks like some of the commits aren't properly verified by GitHub.

stj:
> I'm still not familiar with the test suite and can't fully assess changes there, but the rest LGTM!

Changes to the test is only moving code out of the python3.8 directory. I simply put the tests at the end of an existing equally named file. If it helps to review, I can create a different PR for these changes.

jakob-keller:
> If it helps to review, I can create a different PR for these changes.

Thanks, but that's not necessary.

jakob-keller:
> Looks like some of the commits aren't properly verified by GitHub.

@stj: The PR is blocked due to unverified commits. Could you fix that so we can merge your work? Thanks!

stj:
Sorry, I haven't had a chance to get to this. I'll squash and force push once I am on a machine that has my commit signature configured.

PR #1085, opened 2024-02-02
Replace aiohttp with httpx

jakkdl:
First step of #749 as described in https://github.com/aio-libs/aiobotocore/issues/749#issuecomment-1698255533

I was tasked with implementing this, but it's been a bit of a struggle not being very familiar with aiohttp, httpx or aiobotocore - and there being ~zero in-line types. But I think I've fixed enough of the major problems that it's probably useful to share my progress.

There's a bunch of random types added. I can split those off into a separate PR or remove if requested. Likewise for `from __future__ import annotations`.

TODO:

* [ ] exceptions
  * retryable_exceptions: mostly just need to go through all httpx exceptions and decide which ones are fine
  * replace all aiohttp exceptions in httpsession.send
* [ ] proxy support
  * this was previously handled per-request, but AFAICT you can only configure proxies per-client in httpx. So need to move the logic for it, and cannot use `botocore.httpsession.ProxyConfiguration.proxy_[url,headers]_for(request.url)`
  * raising of ProxyConnectionError is very ugly atm, and probably not "correct"?
  * BOTO_EXPERIMENTAL__ADD_PROXY_HOST_HEADER
    * seems not possible to do when configuring proxies per-client?
* [ ] wrap io.IOBase data in a non-sync-iterable async iterable
* [ ] url's were previously wrapped with `yarl.URL(url, encoding=True)`. httpx does not support yarl, but does encoding still need to be done? I have not managed to figure out what this previously accomplished.
* [ ] The following connector_args I have yet to fix:
  * `use_dns_cache`: did not find any mentions of dns caches on a quick skim of httpx docs
  * `force_close`: same. Can maybe find out more by digging into docs on what this option does in aiohttp.
  * `resolver`: this is an `aiohttp.abc.AbstractResolver` which is obviously a no-go.
    * raise error for code passing this
    * figure out equivalent functionality for httpx
* [ ] I have added change info to [CHANGES.rst](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.rst)
* [ ] `test_patches` previously cared about aiohttp. That can probably be retired?
* [ ] replace aiohttp with httpx in `tests.mock_server.AIOServer`?

It would also be very good if tests were added that checked the above functionality. I do have 5 tests failing currently (at least locally), 2 are for retries, the other three I haven't fully figured out yet.

### Checklist when updating botocore and/or aiohttp versions

TODO: Fix the above template
[...]

PR #1357, opened 2025-05-14
dont use matrix for conditional variables in CI

jakkdl:
### Description of Change
Currently CI uses `include` in the matrix to define variables, which is non-standard usage and makes adding additional entries counter-intuitive and makes the config harder to read.

See https://github.com/aio-libs/aiobotocore/pull/1085#issuecomment-2697049084

This PR simply moves the logic to where it's actually used

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -41,20 +41,15 @@ jobs:
         os:
         - ubuntu-24.04
         - ubuntu-24.04-arm
-        include:
-        - experimental: false
-        - upload-coverage: false
-        - upload-coverage: true
-          python-version: 3.11
-          os: ubuntu-24.04
       fail-fast: false
     uses: ./.github/workflows/reusable-test.yml
     with:
       python-version: ${{ matrix.python-version }}
       os: ${{ matrix.os }}
-      continue-on-error: ${{ matrix.experimental }}
+      continue-on-error: false
       enable-cache: ${{ github.ref_type == 'tag' && 'false' || 'auto' }}
-      upload-coverage: ${{ matrix.upload-coverage }}
+      upload-coverage:
+        ${{ matrix.python_version == 3.11 && matrix.os == 'ubuntu-24.04' }}
     secrets:
       # yamllint disable-line rule:line-length
       codecov-token: ${{ matrix.upload-coverage && secrets.CODECOV_TOKEN || '' }}


.github/workflows/ci-cd.yml
```diff
@@ -41,20 +41,15 @@ jobs:
...
-        - upload-coverage: true
-          python-version: 3.11
-          os: ubuntu-24.04
       fail-fast: false
     uses: ./.github/workflows/reusable-test.yml
     with:
       python-version: ${{ matrix.python-version }}
       os: ${{ matrix.os }}
-      continue-on-error: ${{ matrix.experimental }}
+      continue-on-error: false
```
jakob-keller, line 49: `experimental` is currently not used. We might as well remove the `continue-on-error` line.

jakob-keller (commented):
I don't have a strong opinion on whether this PR has merit. I think @webknjaz is the expert who should do the final review.

.github/workflows/ci-cd.yml
jakkdl: I could see keeping the line if you foresee wanting to re-enable it for certain setups in the future. e.g. if you want to start testing 3.14 betas/rc but not block CI if they break

jakob-keller: That's exactly what I had been thinking, but we may just add that line again, if that need arises. I'm fine either way.

webknjaz: Personally, I implemented inferring this from the Python version string. Like if you have something like `3.14-dev` or `~3.14.0-0`, my things know that these are unstable. And for the explicit flag I call it `xfail` to reference known pytest terms..
I'd keep this in. Perhaps, if you end up trying out my reusable workflow for tox, it has this and I wouldn't be removing the flag, only to re-add it later. It's not as important on the calling side but it could be nice for consistency.

.github/workflows/ci-cd.yml
```diff
@@ -41,20 +41,15 @@ jobs:
         os:
         - ubuntu-24.04
         - ubuntu-24.04-arm
```
webknjaz, line 43: Could be useful to be more explicit:
```suggestion
        - ubuntu-24.04-arm
        experimental:
        - false
```

.github/workflows/ci-cd.yml
```diff
@@ -41,20 +41,15 @@ jobs:
...
     uses: ./.github/workflows/reusable-test.yml
     with:
       python-version: ${{ matrix.python-version }}
       os: ${{ matrix.os }}
-      continue-on-error: ${{ matrix.experimental }}
+      continue-on-error: false
       enable-cache: ${{ github.ref_type == 'tag' && 'false' || 'auto' }}
-      upload-coverage: ${{ matrix.upload-coverage }}
+      upload-coverage:
+        ${{ matrix.python-version == 3.11 && matrix.os == 'ubuntu-24.04' }}
```
webknjaz, line 52: Why have it conditional, even?

.github/workflows/ci-cd.yml
jakkdl: I don't think this makes sense as a change - there's no world in which we want to parametrize an entry to run with both experimental as `true` and `false`, so it does not belong in the matrix.

.github/workflows/ci-cd.yml
jakkdl: This was previously used for a specific combo of OS+python with 3.8+arm so `-dev` would not have been sufficient for that, but it's a fine solution in general. (although we've fucked up something in trio https://github.com/python-trio/trio/pull/3264#discussion_r2088628435)

.github/workflows/ci-cd.yml
jakob-keller: I tried to follow [GitHub's documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#expanding-or-adding-matrix-configurations) and it obviously works. Feel free to structure the matrix differently, if you want to apply "best practices". I have no strong opinion either way.

.github/workflows/ci-cd.yml
jakkdl: and why `3.11` in particular :p 

proper solution would indeed be flags+uploads on every matrix entry, but for this PR I just wanted to pull out this change from #1085 so it's less of a pain to keep that PR up to date.

.github/workflows/ci-cd.yml
webknjaz: Yep, this way is what's suggested in GitHub's own docs.

jakkdl:
uh, don't think these fails are related to this PR. Looks like github being weird

.github/workflows/ci-cd.yml
webknjaz: Let's recover this together with my other change.
```suggestion
      continue-on-error: ${{ matrix.experimental }}
```

.github/workflows/ci-cd.yml
```diff
@@ -41,20 +41,17 @@ jobs:
...
       python-version: ${{ matrix.python-version }}
       os: ${{ matrix.os }}
       continue-on-error: ${{ matrix.experimental }}
       enable-cache: ${{ github.ref_type == 'tag' && 'false' || 'auto' }}
-      upload-coverage: ${{ matrix.upload-coverage }}
+      upload-coverage:
+        ${{ matrix.python-version == 3.11 && matrix.os == 'ubuntu-24.04' }}
     secrets:
       # yamllint disable-line rule:line-length
       codecov-token: ${{ matrix.upload-coverage && secrets.CODECOV_TOKEN || '' }}
```
jakob-keller, line 57: Sorry, but we need to change that line also, if we want to get rid of `upload-coverage`.

jakob-keller (changes requested):

.github/workflows/ci-cd.yml
jakob-keller: We could fast-track #1367 to resolve this issue.

webknjaz: Why can't we just merge this one first?

webknjaz: Let's patch it instead, then.

webknjaz, line 57:
```suggestion
       codecov-token: ${{ secrets.CODECOV_TOKEN }}
```

webknjaz (commented):
No need for this to be conditional. The secret will evaluate to an empty string in untrusted contexts.

PR #1362, opened 2025-05-26
Bumped botocore and boto3 to 1.38.23

terricain:
### Description of Change

botocore and boto3 now match versions, so future dep updates should be simpler.

Fixed test_put_object_sha256 as the test now works with the current version of Moto.

Added `context.py` which provides some Async equivalent functions for contextvar wrapping decorator functions.

It seems a 2.23.0 tag does not exist, so I added to that in the changelog.

https://github.com/boto/botocore/compare/1.37.2...1.38.23

### Assumptions

The contextvar bits are fairly well sprinkled over botocore, mainly methods we do not patch or override. It seems like this minimal implementation is enough to satisfy all of that. Before I fixed it, it was very unhappy with "variable is None, attribute exception", so I'm fairly confident this will suffice for now.

  * [x] Why needed

* [x] I have added URL to diff: https://github.com/boto/botocore/compare/[CURRENT_BOTO_VERSION_TAG]...[NEW_BOTO_VERSION_TAG]

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,9 +1,15 @@
 Changes
 -------
 
-2.23.0 (2025-05-02)
+2.23.0 (2025-05-27)
 ^^^^^^^^^^^^^^^^^^^
 * drop support for Python 3.8 (EOL)
+* bump botocore dependency specification
+* patch ``AioClientCreator.create_client()``
+* patch ``StreamingBody.readinto`` and ``StreamingChecksumBody.readinto`` response classes
+* patch ``_apply_request_trailer_checksum``, ``_get_provider_params`` and ``AIOWaiter.wait`` to add context feature information
+* patch ``AioSession._create_client`` to propagate context information. Created ``context.py`` to provide equivalent async context manager and async decorator functions
+* fixed test ``test_put_object_sha256`` as ``moto`` supports ``ChecksumSHA256``
 
 2.22.0 (2025-04-29)
 ^^^^^^^^^^^^^^^^^^^
--- a/aiobotocore/client.py
+++ b/aiobotocore/client.py
@@ -18,6 +18,7 @@ from botocore.waiter import xform_name
 
 from . import waiter
 from .args import AioClientArgsCreator
+from .context import with_current_context
 from .credentials import AioRefreshableCredentials
 from .discovery import AioEndpointDiscoveryHandler, AioEndpointDiscoveryManager
 from .httpchecksum import apply_request_checksum
@@ -104,6 +105,7 @@ class AioClientCreator(ClientCreator):
         )
         self._register_s3express_events(client=service_client)
         self._register_s3_control_events(client=service_client)
+        self._register_importexport_events(client=service_client)
         self._register_endpoint_discovery(
             service_client, endpoint_url, client_config
         )
@@ -327,6 +329,7 @@ class AioBaseClient(BaseClient):
         """Closes underlying endpoint connections."""
         await self._endpoint.close()
 
+    @with_current_context()
     async def _make_api_call(self, operation_name, api_params):
         operation_model = self._service_model.operation_model(operation_name)
         service_name = self._service_model.service_name
--- a/aiobotocore/context.py
+++ b/aiobotocore/context.py
@@ -1,0 +1,42 @@
+from contextlib import asynccontextmanager
+from copy import deepcopy
+from functools import wraps
+from inspect import iscoroutinefunction
+
+from botocore.context import (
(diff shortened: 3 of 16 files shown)


aiobotocore/client.py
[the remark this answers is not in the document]
terricain: Good spot! Yeah added this.

aiobotocore/context.py
```diff
@@ -0,0 +1,75 @@
...
+    ClientContext,
+    get_context,
+    reset_context,
+    set_context,
+)
+
+
+@asynccontextmanager
+async def start_as_current_context(ctx=None):
+    """
```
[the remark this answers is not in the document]
terricain: Removing them is easier 😄

terricain:
Will add a test for StreamingChecksumBody for completeness sake, should be good to go after that

jakob-keller (approved):

terricain:
Added some tests for the streaming checksum body. Which makes sense as we're having to reimplement `readinto`, so this'll give us some more assurances.

> With regards to unit tests, my understanding is limited and I only glanced at those.

Tbh they're mostly just pytest inspired tests from botocore itself.

50:50 codecov will still be unhappy, iirc it wants coverage on the pytest fixtures 😄 🤷🏻‍♀️

aiobotocore/context.py
```diff
@@ -0,0 +1,42 @@
...
+def with_current_context(hook=None):
+    def decorator(func):
+        @wraps(func)
+        async def wrapper(*args, **kwargs):
+            async with start_as_current_context():
+                if hook:
+                    if iscoroutinefunction(hook):
+                        await hook()
+                    else:
+                        hook()
```
thehesiod, line 37: fyi if you'd like: https://github.com/aio-libs/aiobotocore/blob/master/aiobotocore/_helpers.py#L4

aiobotocore/response.py
```diff
@@ -71,6 +71,28 @@ async def read(self, amt=None):
             self._verify_content_length()
         return chunk
 
+    async def readinto(self, b: bytearray):
+        """Read bytes into a pre-allocated, writable bytes-like object b, and return the number of bytes read."""
+        try:
+            # AioHTTP ClientResponse does not support .readinto()
+            chunk = await self.__wrapped__.content.read(len(b))
```
thehesiod, line 78: @jakkdl looks like you'll need to add this to your pr

CHANGES.rst
```diff
@@ -1,9 +1,15 @@
...
-2.23.0 (2025-05-02)
+2.23.0 (2025-05-27)
 ^^^^^^^^^^^^^^^^^^^
 * drop support for Python 3.8 (EOL)
+* bump botocore dependency specification
+* patch ``AioClientCreator.create_client()``
+* patch ``StreamingBody.readinto`` and ``StreamingChecksumBody.readinto`` response classes
+* patch ``_apply_request_trailer_checksum``, ``_get_provider_params`` and ``AIOWaiter.wait`` to add context feature information
+* patch ``AioSession._create_client`` to propagate context information. Created ``context.py`` to provide equivalent async context manager and async decorator functions
+* fixed test ``test_put_object_sha256`` as ``moto`` supports ``ChecksumSHA256``
```
thehesiod, line 12: not sure we need to be so chatty for internals

thehesiod:
going to merge but may drop some of the chattiness in changes later.  I think we just need to let people know things that can affect them.  Thanks Teri!

aiobotocore/context.py
terricain: Ha that would have been ideal 😄

aiobotocore/waiter.py
```diff
@@ -81,6 +83,7 @@ async def __call__(self, **kwargs):
 
 
 class AIOWaiter(Waiter):
+    @with_current_context(partial(register_feature_id, 'WAITER'))
```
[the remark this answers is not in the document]
terricain: We do have tests which test the waiters so its not impacting anything, but fixing it would be "more correct" 😄

aiobotocore/context.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def with_current_context(hook=None):
+    def decorator(func):
+        @wraps(func)
+        async def wrapper(*args, **kwargs):
+            async with start_as_current_context():
+                if hook:
+                    if iscoroutinefunction(hook):
+                        await hook()
```
stj, line 35: I like that the hook can be a coroutine. Is that something worth putting in a public documentation?

PR #1364, opened 2025-05-27
Add Python 3.14 to test matrix

jakob-keller:
### Description of Change
Add Python 3.14 to test matrix

### Assumptions
CPython 3.14 has recently entered beta phase:

> We strongly encourage maintainers of third-party Python projects to test
> with 3.14 during the beta phase and report issues found to the Python bug
> tracker as soon as possible.

  * [ ] Why needed


jakob-keller:
Python 3.14 is marked as experimental in CI for the time being. That way we could merge this PR now and continue monitoring without breaking CI. Any thoughts @webknjaz, @thehesiod?

webknjaz:
> Personally, I find the existing matrix definition intuitive. But

This is the only project where I saw it being set up like this. So it's best to just follow what's familiar to the majority of people. Even if I spend time understanding it, this will confuse many more people. I'll add a commit on top of #1357 to follow the pattern of other aio-libs projects, and then you can rebase this one + adapt the include entries to enumerate all the factors explicitly.

.github/workflows/ci-cd.yml
```diff
@@ -38,11 +38,14 @@ jobs:
         - 3.11
         - 3.12
         - 3.13
+        - 3.14
```
webknjaz, line 41: If you use
```suggestion
        - ~3.14.0-0
```
or
```suggestion
        - 3.14-dev
```
then you could check for “starts with a tilde or ends with a dash dev” in the reusable workflow and have the experimental flag built-in. For any additional experimental states (OS/arch-dependent), you could use `experimental` explicitly (and stop using `allow-prereleases` in there too).

webknjaz (approved):

PR #1389, opened 2025-07-29
Bump `botocore` dependency specification

claytonparnell:
### Description of Change
Bumping botocore spec to support one more minor version (1.39.9) as this contains functionality needed by my team. Updated boto3/awscli versions as well. Only found one change needed to aiobotocore codebase, but this is my first time committing to this repo so please confirm 🙂 
https://github.com/boto/botocore/compare/1.39.8...1.39.9

### Assumptions
Upstream contains several changes that require adjustments to the aiobotocore codebase.

  * [ ] Why needed

* [x] I have added URL to diff: https://github.com/boto/botocore/compare/1.39.8...1.39.9


pyproject.toml
```diff
@@ -31,7 +31,7 @@ dynamic = ["version", "readme"]
 dependencies = [
     "aiohttp >= 3.9.2, < 4.0.0",
     "aioitertools >= 0.5.1, < 1.0.0",
-    "botocore >= 1.39.7, < 1.39.9", # NOTE: When updating, always keep `project.optional-dependencies` aligned
+    "botocore >= 1.39.7, < 1.39.10", # NOTE: When updating, always keep `project.optional-dependencies` aligned
```
jakob-keller: Could you raise the minimum supported version to 1.39.9? Your modifications are incompatible with earlier releases.

The same goes for `awscli` and `boto3` further down below.

claytonparnell:
Thanks @jakob-keller for running the checks - I see only py3.14 tests failed, which seem to be consistent with the [previous version bump](https://github.com/aio-libs/aiobotocore/pull/1387).

jakob-keller (changes requested):
Thank you for proposing this PR!

Please have a look at the following issues and comments:

* I believe, we need to patch `create_nested_client()` in `utils.py`: Otherwise the plugin context will not be properly applied to the client. A solution looks non-trivial to me.
* Following botocore's lead, could you add the Python 3.14 PyPI classifier to `pyproject.toml`?

The Python 3.14 tests are expected to fail. That can be ignored. We have a known incompatibility with the pydantic dev dependency which should be resolved any day now.

claytonparnell:
>I believe, we need to patch create_nested_client() in utils.py: Otherwise the plugin context will not be properly applied to the client. A solution looks non-trivial to me.

@jakob-keller - thank you for your comments! I've taken a stab at it, overriding `create_nested_client` in utils.py as well as `_get_client_creator` in credentials.py. Added unit tests as well. Can you pls take a look when you have time? Thank you! 🙏

aiobotocore/utils.py
```diff
@@ -747,3 +747,27 @@ async def _get_response(self, full_url, headers, timeout):
...
+    from botocore.utils import (
+        PluginContext,
+        reset_plugin_context,
+        set_plugin_context,
+    )
+
+    ctx = PluginContext(plugins="DISABLED")
+    token = set_plugin_context(ctx)
+    try:
+        return session.create_client(service_name, **kwargs)
```
jakob-keller: I don't think this will work for us: `session.create_client()` returns a coroutine. We need to ensure that code is executed before `reset_plugin_context()` is invoked.

We could turn `create_nested_client()` into an asynchronous function and await `session.create_client()`.

aiobotocore/utils.py
```diff
@@ -747,3 +747,27 @@ async def _get_response(self, full_url, headers, timeout):
...
+    """Create a nested client with plugin context disabled.
+
+    If a client is created from within a plugin based on the environment variable,
+    an infinite loop could arise. Any clients created from within another client
+    must use this method to prevent infinite loops.
+
+    This is the async version of botocore.utils.create_nested_client that works
+    with aiobotocore's async session.
+    """
+    from botocore.utils import (
```
jakob-keller: Why use nested import instead of regular import at the top of the module?

claytonparnell: Good callout; idk why I defaulted to lazy loading, moved the imports to the top of module to match rest of file's format

aiobotocore/utils.py
claytonparnell: Reworked as async function, please take a look when you get a chance. Thank you!

aiobotocore/credentials.py
```diff
@@ -1101,3 +1101,20 @@ async def load(self):
             method=self.METHOD,
             refresh_using=sso_fetcher.fetch_credentials,
         )
+
+
+def _get_client_creator(session, region_name):
```
jakob-keller, line 1106: Please include the patched function in `test_patches.py` to protect against future regressions.

aiobotocore/utils.py
```diff
@@ -747,3 +750,32 @@ async def _get_response(self, full_url, headers, timeout):
                 f"container metadata: {e}"
             )
             raise MetadataRetrievalError(error_msg=error_msg)
+
+
+@contextlib.asynccontextmanager
+async def create_nested_client(session, service_name, **kwargs):
```
jakob-keller, line 756: Please include the patched function in `test_patches.py` to protect against future regressions.

jakob-keller, line 756: `create_nested_client()` is also called in upstream `tokens.py` line 260. Looks like we need to patch that occurrence as well.

claytonparnell:
Thank you for your patience and comments @jakob-keller 🙂 I've addressed the comments, working now on figuring out the commit verification.

.github/workflows/reusable-build.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         persist-credentials: false
     - name: Install uv
       # yamllint disable-line rule:line-length
-      uses: astral-sh/setup-uv@7edac99f961f18b581bbd960d59d049f04c0002f  # v6.4.1
+      uses: astral-sh/setup-uv@e92bafb6253dcd438e0484186d7669ea7a8ca1cc  # v6.4.3
```
jakob-keller: Should this be included in the PR?

.github/workflows/reusable-test.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         submodules: true
     - name: Install uv
       # yamllint disable-line rule:line-length
-      uses: astral-sh/setup-uv@7edac99f961f18b581bbd960d59d049f04c0002f  # v6.4.1
```
jakob-keller: Should this be included in the PR?

.github/workflows/reusable-build.yml
claytonparnell: Sorry - guess this automatically happened at some point. can remove now

claytonparnell: Removed the commit https://github.com/claytonparnell/aiobotocore/commit/ea8a695a678bf44a409245026dc6558e43ffd961

claytonparnell:
taking a look at test failures

tests/test_patches.py
```diff
@@ -1159,6 +1167,12 @@ def test_protocol_parsers():
                 '824d41775dbb8a05184f6e9c7b2ea7202b72f2a9',
             },
         ),
+        (
+            SSOTokenProvider._client,
```
jakob-keller: I guess the hash is unstable due to the applied decorator. Unless there's a `__wrapped__` attribute or similar that I am unaware of, we might be forced to leave out that test case.

claytonparnell: That's what it seems like; the hash seems to change per environment. If you're good with removing it for now that would unblock, or I can spend some time working on modifying the test to support the decorated function

jakob-keller: No, I guess we're good.

aiobotocore/tokens.py
```diff
@@ -152,6 +157,14 @@ async def _refresher(self):
             token_dict["accessToken"], expiration=expiration
         )
 
+    @CachedProperty
+    def _client(self):
+        config = Config(
```
jakob-keller: I am pretty sure that this should be an instance of `aiobotocore.config.AioConfig`. We must have overlooked that previously.

tests/test_plugin_context.py
```diff
@@ -0,0 +1,426 @@
...
+                mock_token = Mock()
+                mock_set.return_value = mock_token
+
+                # This might raise an exception for invalid service, but context should still be managed
+                client_context = create_nested_client(
+                    session, 'invalid-service', region_name='us-east-1'
+                )
+                try:
+                    async with client_context:
+                        pass
```
jakob-keller: You could add the following comment to avoid the coverage miss:

```# pragma: no cover```

File: aiobotocore/tokens.py (hunk 9 above)
Comment (jakob-keller) [reply aiobotocore/tokens.py]: Well, on second thought: That is probably irrelevant here.

File: aiobotocore/tokens.py [hunk 11]
```diff
@@ -152,6 +157,14 @@ async def _refresher(self):
             token_dict["accessToken"], expiration=expiration
         )
 
+    @CachedProperty
```
jakob-keller: I wonder if this is or needs to be reentry safe? I'm afraid the cached async context manager might only be consumed once. Any thoughts?

tests/test_plugin_context.py
jakob-keller: The comment needs to be appended to the `pass` statement to be effective.

pyproject.toml
```diff
@@ -31,7 +32,7 @@ dynamic = ["version", "readme"]
 dependencies = [
     "aiohttp >= 3.9.2, < 4.0.0",
     "aioitertools >= 0.5.1, < 1.0.0",
-    "botocore >= 1.39.7, < 1.39.9", # NOTE: When updating, always keep `project.optional-dependencies` aligned
+    "botocore >= 1.39.9, < 1.39.10", # NOTE: When updating, always keep `project.optional-dependencies` aligned
```
jakob-keller: If you like, you may safely bump the upper bound to `< 1.39.12` since those two additional botocore releases do not contain any relevant changes. If you do, don't forget to bump the optional dependencies as well and run `uv lock` before committing.

aiobotocore/tokens.py
claytonparnell: this is a good point.. I guess in botocore the client itself is cached, but here create_nested_client is context manager. Thinking using `@property` instead of `@CachedProperty` would solve reentry issue, so we get a fresh context manager each time

jakob-keller: That's what I was thinking. Just hoping that there won't be a significant performance penalty from removing the cache. But as far as I can tell, `_client` is (currently) used solely on token refresh, so hopefully not that often.

pyproject.toml
jakob-keller: Sorry, my bad: Please drop that commit. botocore 1.39.10 contains updates to data files that require changes to test cases.

claytonparnell: if it's just the SNS topic, I can remove SNS:Receive from the test case policy and it passes locally; not sure if there's other code changes that would be needed for such a change?

jakob-keller: No, I guess that's fine then. Thanks!

claytonparnell: would I need to update the lowerbound in this case? I think all the functionality works with older versions in the bounds, but the tests would only pass with latest version

jakob-keller: The updated `uv.lock` bumps a bunch of other dependencies. We tend to do that separately (i.e. #1390). Could you reverse those?

FWIW: I use this to selectively update the lock file:
```uv sync --upgrade-package botocore --upgrade-package boto3 --upgrade-package awscli```

Comment (jakob-keller) [reply pyproject.toml]: > would I need to update the lowerbound in this case? I think all the functionality works with older versions in the bounds, but the tests would only pass with latest version

That's fine.

**Comment by claytonparnell** [comment]
Done, thanks so much for all the support on this Jakob 😄

**Review by jakob-keller — APPROVED** [review]

**Comment by jakob-keller** [comment]
> Done, thanks so much for all the support on this Jakob 😄

Don't mention it! You picked a non-trivial version bump for your first contribution and kept at it. Well done!

I asked @thehesiod for a second opinion. If he does not spot any issues within the next days, I'll cut a release.

**Comment by thehesiod** [comment]
awesome work guys! will take a look later

File: CHANGES.rst [hunk 13]
```diff
@@ -1,6 +1,10 @@
 Changes
 -------
 
+2.24.0 (2025-07-31)
```
thehesiod, line 4: should we try to normalize against botocore?  perhaps 2.39.11 ?  Only bad part of this is if we need patch releases between theirs

aiobotocore/tokens.py
```diff
@@ -152,6 +156,14 @@ async def _refresher(self):
             token_dict["accessToken"], expiration=expiration
         )
 
+    @property
```
thehesiod, line 159: should we use `@CachedProperty` like they do?

aiobotocore/utils.py
```diff
@@ -747,3 +750,32 @@ async def _get_response(self, full_url, headers, timeout):
...
+            response = await client.assume_role(...)
+    """
+    # Set plugin context to disabled
+    ctx = PluginContext(plugins="DISABLED")
+    token = set_plugin_context(ctx)
+
+    try:
+        # Create the client context and enter it
+        client_context = session.create_client(service_name, **kwargs)
+        async with client_context as client:
```
thehesiod: i wish botocore would adopt this workflow, it would be so much nicer

aiobotocore/utils.py
```diff
@@ -747,3 +750,32 @@ async def _get_response(self, full_url, headers, timeout):
...
+            yield client
+    finally:
+        # Always reset plugin context
+        reset_plugin_context(token)
```
thehesiod: I think this is not quite right, in botocore after the client is created it resets the plugin context, if so this means right after `async with client_context as client:` reset should be called

thehesiod:
ok finished my cmts

aiobotocore/tokens.py
claytonparnell: I had previously used `@CachedProperty`; but we thought this would not have the desired behavior (comments [here](https://github.com/aio-libs/aiobotocore/pull/1389#discussion_r2246277237)). 
In botocore, the `@CachedProperty` is the client itself. For aiobotocore we would return a context manager rather than a client, using `@property` ensures a fresh context manager every time and ensures it is reentry safe

aiobotocore/utils.py
claytonparnell: Makes sense; will update

CHANGES.rst
claytonparnell: up to you 😄 can add this or can address it separately

claytonparnell:
@jakob-keller @thehesiod - addressed comments, can you take a look? Thank you!

CHANGES.rst
jakob-keller: I'd argue that our release cycles are too disconnected to bother about it. In any case, it would be beyond the scope of this PR.

aiobotocore/utils.py
```diff
@@ -747,3 +750,34 @@ async def _get_response(self, full_url, headers, timeout):
...
+    # Set plugin context to disabled
+    ctx = PluginContext(plugins="DISABLED")
+    token = set_plugin_context(ctx)
+
+    try:
+        # Create the client context
+        client_context = session.create_client(service_name, **kwargs)
+    finally:
+        # Reset plugin context immediately after client creation, matching botocore behavior
+        reset_plugin_context(token)
```
jakob-keller: I think this would be the correct way to do it:

``` python
ctx = PluginContext(plugins="DISABLED")
token = set_plugin_context(ctx)

try:
    async with session.create_client(service_name, **kwargs) as client:
        reset_plugin_context(token)
        token = None

        yield client
finally:
    if token:
        reset_plugin_context(token)
```

claytonparnell: Makes sense, I've updated to match this

thehesiod:
will do asap, wife coming back home soon and will have more time

aiobotocore/tokens.py
thehesiod: we can add it later too

thehesiod (approved):

claytonparnell:
@thehesiod @jakob-keller thanks so much for the help reviewing and merging this! Is there anything needed to kick off a pypi release?

jakob-keller:
> @thehesiod @jakob-keller thanks so much for the help reviewing and merging this! Is there anything needed to kick off a pypi release?

Thanks for the heads up. It should be published within the hour!

PR #1395, opened 2025-08-13
Fix endpoint circular import issue

thehesiod:
### Description of Change
Resolved an issue with circular imports in the endpoint code by reorganizing the imports to avoid dependency loops.

  * [x ] Detailed description of new feature
  * [x] Why needed

* [ x] I have read and followed [CONTRIBUTING.rst](https://github.com/aio-libs/aiobotocore/blob/master/CONTRIBUTING.rst#how-to-upgrade-botocore)


.pre-commit-config.yaml
```diff
@@ -18,7 +18,7 @@ repos:
     args: [--fix]
   - id: ruff-format
 - repo: https://github.com/astral-sh/uv-pre-commit
-  rev: 0.7.21
+  rev: 0.8.8
```
jakob-keller: Would you mind merging my pending dev deps PR first? It includes that bump.

jakob-keller (changes requested):
LGTM, but the uv bump should be done separately IMO.

thehesiod:
I've always considered the idea of always segregating different types of changes into different PRs due to later bisection one of those ideas that makes sense when you have two large bodies of distinct changes, however as someone who has 25 years of experience I find one needs to always consider each circumstance individually (this happens at my work as well).

IOW at the end of the day no one should just mindlessly revert commits, one should always analyze what was changed.  So the question is, what are the chances of a PR breaking in the future, what is the effort involved undoing a commit that has an issue, and how much effort is it in creating new PR processes for the "unrelated" changes. 

In this PR I think it's pretty clear, there is minimal chance this can break something, the "unrelated" changes have no effect in the issue, and they're are very easy to segregate in the future.

TLDR we should not be blindly following "best practices", everything needs to be done intelligently, minimizing current and future efforts.  Take the whole picture.

anyways my rant for now.  I think generally people like easy procedures, but tend not think about the fact that life is more complicated, and there is no solution that is best in every circumstance.

anyways will revert for now until/if I convert you to my "dark" side :)

jakob-keller (approved):

jakob-keller:
Sorry 🙂

thehesiod:
@jakob-keller we're between a rock and hard place with 3.14:
```
 `pydantic-core` (v2.33.2) was included because `aiobotocore:dev` depends on `moto[server]` (v5.1.10) which depends on `cfn-lint` (v1.38.3) which depends on `aws-sam-translator` (v1.99.0) which depends on `pydantic` (v2.11.7) which depends on `pydantic-core`
```
thoughts, swap to experimental?

thehesiod:
actually it is, hmm, why is this blocked

thehesiod:
there's really no point in building 3.14 until it's stable, it just gives noise in the PRs, I'm going to disable until it's stable

jakob-keller:
> actually it is, hmm, why is this blocked

I don’t think there was anything wrong with regards to Python 3.14 tests. Codecov has been flaky for a few days. Maybe that’s the issue? Trying again usually helps.

thehesiod:
codecov said it had passed

thehesiod:
ok @jakob-keller looks like I need another approval

jakob-keller:
To me it looks like CI is now broken, whereas it worked fine before.

thehesiod:
omg this system is so broken lol, before it said everything was fine except it needed approval.  Looks like something got cancelled, let me try re-building

jakob-keller:
It should also work without the changes done to CI. Would you mind trying again? Otherwise, just merge the PR without my approval.

thehesiod:
I'm going to keep as-is, I really don't like the idea of builds "failing" but presumably not affecting the mergability, it adds mental load to something that should be dead simple

Issue #1375, opened 2025-06-20
Replace aiobotocore with Native Async Python Bindings to the AWS SDK for C++


thehesiod:
(note had AI write a bunch of this *g*)

aiobotocore enables asyncio support by layering async I/O on top of botocore, which is a synchronous, low-level AWS SDK designed for boto3. It does this by patching internal components and substituting key behaviors with async-compatible implementations using aiohttp or httpx.

While this has extended the usefulness of botocore in the async world, it comes with significant downsides:
	•	Fragility: aiobotocore depends on internal botocore APIs, which change frequently and cause frequent breakages.
	•	Incomplete Async Behavior: Many operations still run synchronously or with blocking I/O, limiting concurrency.
	•	Maintenance Overhead: Keeping up with botocore releases and protocol changes is a continuous and complex burden.
	•	Performance Constraints: Python-level async adds overhead compared to true native async implementations.

A New Direction

Rather than continuing to patch around botocore, this proposal suggests building a native async Python client that wraps the [AWS SDK for C++](https://github.com/aws/aws-sdk-cpp), offering:
	•	✅ True async I/O and thread safety backed by native code
	•	✅ Higher performance for high-throughput services (e.g., S3, DynamoDB streams)
	•	✅ Long-term stability from using officially supported AWS SDKs
	•	✅ Elimination of fragile monkey-patching and compatibility layers

Why This Is Feasible Now

Previously, this approach was dismissed as too complex due to the cost of:
	•	Writing Pybind11 or CFFI wrappers for many AWS services
	•	Managing cross-platform builds of the AWS SDK for C++
	•	Designing clean, Pythonic async interfaces on top of C++ code

Today, with the help of AI-assisted tooling, these barriers are lower than ever:
	•	AI can generate and maintain binding code, docstrings, and type hints
	•	Glue code and examples can be iterated on rapidly with LLM assistance

Goals
	•	Build a proof-of-concept async Python wrapper for the AWS C++ SDK (e.g. S3 or STS).
	•	Benchmark it against aiobotocore in terms of latency, throughput, and memory usage.
	•	Maintain idiomatic asyncio interfaces (e.g., async for, async with, await client.get_object(...)).
	•	Explore possible adoption paths for production workloads.

thehesiod:
@claude could you do deep research on this subject, I think before the major sticking point was that we'd have to release many native bindings, and the size of the library would increase drastically

dacevedo12:
Hey, a few notes:
[...]

PR #1399, opened 2025-08-22
Switched async test runner from pytest-asyncio to AnyIO

agronholm:
This is to allow tests to be run against Trio in addition to asyncio in the future.

### Description of Change

This migrates the test suite to use AnyIO for running async tests. This allows async tests to be run against Trio in a follow-up PR.

* [X] I have added change info to [CHANGES.rst](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.rst)
  * [ ] Why needed


agronholm:
@thehesiod Given https://github.com/aio-libs/aiobotocore/issues/1375, is this something you would still want? It's not clear to me how those proposed changes would mesh with different Python async frameworks.

tests/test_sqs.py
```diff
@@ -2,6 +2,8 @@
 
 import pytest
 
+pytestmark = pytest.mark.anyio
```
thehesiod: is this necessary?  Can't this be done globally?

thehesiod:
so my worry is going to be support costs, we have very few maintainers of aiobotocore.  How big a lift will it be maintaining support for trio?

thehesiod:
> @thehesiod Given #1375, is this something you would still want? It's not clear to me how those proposed changes would mesh with different Python async frameworks.

well, depends on if that task is even possible :) That's a pie in the sky wish for when someone gets a lot of free time. Hard to say right now.

tests/test_sqs.py
agronholm: No, it can't. The AnyIO pytest plugin was specifically designed to require explicit opt-in so as to repeat pytest-asyncio's mistakes with its "auto" mode which tramples over other async plug-ins just by getting loaded.

agronholm:
> so my worry is going to be support costs, we have very few maintainers of aiobotocore. How big a lift will it be maintaining support for trio?

Not a lot, in fact the run-time part of the code shouldn't even see any references to Trio. And given how #1085 was already merged, the necessary minimum piece (beginnings of httpx back-end support) is already in place, I don't see this adding any further maintenance overhead. Of course, given that aiohttp is just an implementation detail, you could consider ditching it entirely in favor of httpx which does support both asyncio and Trio. This would greatly reduce the maintenance overhead. I'm happy to do this work if you give me the green light.

agronholm:
A significant change I would have to make is to get rid of `_RefCountedSession`, as it will potentially cause `__aenter__()` and `__aexit__()` to be called in different tasks which wreaks havoc on any task group running in the HTTP client (which is essential on HTTP/2). This would be replaced with a permanent HTTP session to be started (and stopped) along with the aioboto session. Another problem is that `_RefCountedSession` is tightly coupled with `aiohttp` so it can't use the newly-introduced HttpxSession.

agronholm:
@thehesiod Would you be okay to make the HTTP back-end selection automatic? Aiohttp on asyncio and httpx on other event loop implementations (Trio in practice). That would make things a lot simpler.

thehesiod:
@agronholm omg that RefCountedSession was missed on the httpx PR, that needs to be fixed.  Actually similarly broken in botocore if you have a custom session type.  I think we can get rid of it and assert if session is not passed in, and make sure session is passed over everywhere it's instantiated

tests/test_sqs.py
thehesiod: how about instead we do this in conftest.py?

thehesiod:
> > so my worry is going to be support costs, we have very few maintainers of aiobotocore. How big a lift will it be maintaining support for trio?
> 
> Not a lot, in fact the run-time part of the code shouldn't even see any references to Trio. And given how #1085 was already merged, the necessary minimum piece (beginnings of httpx back-end support) is already in place, I don't see this adding any further maintenance overhead. Of course, given that aiohttp is just an implementation detail, you could consider ditching it entirely in favor of httpx which does support both asyncio and Trio. This would greatly reduce the maintenance overhead. I'm happy to do this work if you give me the green light.

I don't think we want to ditch aiohttp yet, if you want to put a PR together that adds trio support and will take a look

agronholm:
> @agronholm omg that RefCountedSession was missed on the httpx PR, that needs to be fixed. Actually similarly broken in botocore if you have a custom session type. I think we can get rid of it and assert if session is not passed in, and make sure session is passed over everywhere it's instantiated

Yes, I've been looking for the proper place to actually establish that session and it has so far eluded me. It looks like are some weird shenanigans (`ClientCreatorContext` et al) going on to shoehorn async support into classes that inherit from `botocore` and override methods with async ones (thus violating the LSP btw).

> I don't think we want to ditch aiohttp yet, if you want to put a PR together that adds trio support and will take a look

Yes, I've already been working on one that uses `httpx` and `anyio` primitives instead of asyncio's when available. This at least avoids making `anyio` a new hard dependency.

So what about defaulting to httpx sessions outside of asyncio? That would keep the semantics intact for asyncio users.

tests/test_sqs.py
agronholm: I suppose I could try with the `pytest_generate_tests` hook. I'll let you know how that worked out.

thehesiod:
@agronholm Given that the botocore interfaces are dynamic, we don't want to generate/duplicate documentation, and a variety of other reasons following LSP is not at the top of my worries.  I actually came to this project after the initial design was already cemented.  Definitely up for suggestions to simplify/enhance though.

I'm not sure we want to change the default based on the run loop you have running.  What's the issue in having the user select based on a new AioConfig parameter for example?

agronholm:
> I'm not sure we want to change the default based on the run loop you have running. What's the issue in having the user select based on a new AioConfig parameter for example?

Convenience. It just won't work at all on Trio unless you add this specific, mandatory configuration option. Users will be left asking, rightfully, why they have to add that boilerplate code to use `aiobotocore`.

agronholm:
It looks like `test_release_versions` is failing, just as it did locally. But what am I supposed to write as the version number in the changelog?

jakob-keller:
> It looks like `test_release_versions` is failing, just as it did locally. But what am I supposed to write as the version number in the changelog?

The new version number needs to also be included in `__init__.py`, AFAIK.

thehesiod:
> > I'm not sure we want to change the default based on the run loop you have running. What's the issue in having the user select based on a new AioConfig parameter for example?
> 
> Convenience. It just won't work at all on Trio unless you add this specific, mandatory configuration option. Users will be left asking, rightfully, why they have to add that boilerplate code to use `aiobotocore`.

ah I see, it's because aiohttp doesn't work in trio right?  ok that sounds fine then

thehesiod:
I think we need a new issue with the proposal and reasonings then we can all chime in.  Educate us as to why you think this change is needed, the benefits, why you don't think other proposals are as good, etc :)

tests/test_sqs.py
thehesiod: ya, these look kind of like sprinkled yuk (internally we have another name, lol)

agronholm:
> > > I'm not sure we want to change the default based on the run loop you have running. What's the issue in having the user select based on a new AioConfig parameter for example?
> > 
> > 
> > Convenience. It just won't work at all on Trio unless you add this specific, mandatory configuration option. Users will be left asking, rightfully, why they have to add that boilerplate code to use `aiobotocore`.
> 
> ah I see, it's because aiohttp doesn't work in trio right? ok that sounds fine then

Indeed, that was the point of this effort.

I see a few tests failed on 3.9 and 3.10. I'll fix those failures tomorrow. And I'll create that issue you requested too.

jakkdl:
> I think we need a new issue with the proposal and reasonings then we can all chime in. Educate us as to why you think this change is needed, the benefits, why you don't think other proposals are as good, etc :)

This PR is very much a part of https://github.com/aio-libs/aiobotocore/issues/749 and was the planned next step after #1085 (which as mentioned was done because aiohttp doesn't support trio). There really isn't any viable alternative other than running pytest-trio separately, but the anyio pytest plugin is explicitly made for this purpose and doing that separation manually would be a huge pain. So I'm not sure if there's much else to consider in a separate issue.

Though the PR on its own doesn't make much sense without the planned followup PR where enabling the test suite to also run trio + fixing the code such that it properly works on trio, so there is a case for not merging it, or merging it into some other branch, until that is also done.

thehesiod:
if this makes the tests more robust for future PRs I'm fine with merging.  I just hope we can avoid the pytest.mark.anyio sprinkling to make it smaller.  Looks like chatgpt suggested a couple ways around this:
```python
def pytest_configure(config):
    # Apply `anyio` mark to all tests by default
    config.addinivalue_line("markers", "anyio: run test under anyio")
    pytestmark = pytest.mark.anyio
```

or
```python
@pytest.fixture(autouse=True)
def _anyio_marker(pytestconfig):
    # ensures anyio plugin kicks in for all tests
    pytest.mark.anyio
```

I'm guessing you can configure it with a pytest command-line arg as well if you want

agronholm:
> if this makes the tests more robust for future PRs I'm fine with merging. I just hope we can avoid the pytest.mark.anyio sprinkling to make it smaller. Looks like chatgpt suggested a couple ways around this:
> 
> ```python
> def pytest_configure(config):
>     # Apply `anyio` mark to all tests by default
>     config.addinivalue_line("markers", "anyio: run test under anyio")
>     pytestmark = pytest.mark.anyio
> ```
> 
> or
> 
> ```python
> @pytest.fixture(autouse=True)
> def _anyio_marker(pytestconfig):
>     # ensures anyio plugin kicks in for all tests
>     pytest.mark.anyio
> ```
> 
> I'm guessing you can configure it with a pytest command-line arg as well if you want

As I suspected, ChatGPT is just making up random stuff. Just in case, I tried these and neither approach did anything. Just accessing `pytest.mark.anyio` is a no-op. Nothing I tried with `pytest_collection_modifyitems` or `pytest_pycollect_makeitem` worked either. If this is possible to do, then my skill with pytest isn't enough.

thehesiod:
lol very possible with chatgpt . I just got sinus infection but will try asap as well

pyproject.toml
```diff
@@ -100,7 +100,6 @@ readme = { file = ["README.rst", "CHANGES.rst"] }
 include = ["aiobotocore*"]
 
 [tool.pytest.ini_options]
-asyncio_mode = "auto"
```
thehesiod, line 102: replace with `anyio_backend = ["asyncio"]` and you don't need to add `pytestmark = pytest.mark.anyio` to every test file.  Later when trio support is added you can add trio and it will then test with both

tests/test_sqs.py
thehesiod: ok, figured out how to do it, see comment I recently added

pyproject.toml
agronholm: No, this doesn't work. There would have to be an option like that configured in the AnyIO pytest plugin, and there isn't, which I know because I **wrote** said plugin. An option for a similar effect could be added to AnyIO, but right now there isn't any.

agronholm:
So: https://github.com/agronholm/anyio/pull/971
Is this a showstopper issue for this PR?

**EDIT**: The PR has now been merged to `master`.

pyproject.toml
thehesiod: but I did this and it worked all the tests passed, how is that?

thehesiod: I'm going to try again because now I'm confused

thehesiod: ahh, I think they were all skipped ;)

thehesiod: yep, resolving this comment, sorry

thehesiod:
> So: [agronholm/anyio#971](https://github.com/agronholm/anyio/pull/971) Is this a showstopper issue for this PR?

If it's possible to get a new release with the enhance functionality, and it makes sense, without a huge delay I'd prefer that, otherwise it's more overhead for maintainers of aiobotocore when editing/adding new test files.  The skipped tests thing is bad because like in pycharm it looks like they all succeeded.  thoughts?

agronholm:
> > So: [agronholm/anyio#971](https://github.com/agronholm/anyio/pull/971) Is this a showstopper issue for this PR?
> 
> If it's possible to get a new release with the enhance functionality, and it makes sense, without a huge delay I'd prefer that, otherwise it's more overhead for maintainers of aiobotocore when editing/adding new test files.

Editing test modules does not require any extra work. Adding a new test module requires adding this one line to make async tests work, and even that will go away once the linked PR makes its way into a release, so I'm not sure what the fuss is about TBH.

> The skipped tests thing is bad because like in pycharm it looks like they all succeeded. thoughts?

You could turn warnings into errors (selectively even) to get around this.

thehesiod:
@jakob-keller what's your thoughts, wait until that enhancement is added to avoid the sprinkling across all the files, or leave as is and cleanup later?

jakob-keller:
> @jakob-keller what's your thoughts, wait until that enhancement is added to avoid the sprinkling across all the files, or leave as is and cleanup later?

Thanks for asking. My main concern is that I don't really understand the motivation behind the PR or the larger plan with regards to supporting AnyIO, trio, ... As of now, our README states: 

> Async client for amazon services using [botocore](https://github.com/boto/botocore) and [aiohttp](https://github.com/aio-libs/aiohttp)/[asyncio](https://docs.python.org/3/library/asyncio.html).

I interpret that as a call to strive for **timely feature parity with upstream** above all else. Considering community requests, I am probably not alone with that opinion.

Limiting ourselves to aiohttp has not prevented us from becoming one of the most frequently downloaded Python packages.

We recently added experimental httpx support through special casing in several modules, but I would not know when or how to use it and I believe we lack documentation.

Will upcoming changes also be similarly sprinkled in the codebase? That gives me headaches and would make fulfilling this project's main goal more difficult, in particular considering the small number of active contributors.

Having said that, I very much prefer to keep the codebase as maintainable as possible and don't see a need to rush this PR. From my perspective, it does not add any value at this point and might as well be picked up again when/if there are any activities with regards to supporting AnyIO/trio/....

agronholm:
> Having said that, I very much prefer to keep the codebase as maintainable as possible and don't see a need to rush this PR. From my perspective, it does not add any value at this point and might as well be picked up again when/if there are any activities with regards to supporting AnyIO/trio/....

I have a changeset locally where I'm pretty far along in making the library work with async event loops other than asyncio. Testing with AnyIO is a pre-requisite for that, as pytest-asyncio, true to its name, only works with asyncio.

If you do not agree with our goals of moving libraries to use [structured concurrency](https://en.wikipedia.org/wiki/Structured_concurrency) and expanding async event loop support beyond asyncio, I would like to know ASAP so I won't needlessly devote any more time to my efforts. But the PR to add httpx support was done precisely for this purpose, so why did you merge that then?

thehesiod:
From what I read of Trio, expanding support (albeit most likely informal), much like httpx did, of it to aiobotocore is a lofty goal and requires usage of anyio.  I think @jakob-keller and I are in agreement that the PR as is requires a lot of changes particularly to the test framework which will be alleviated by the new drop of your module.

lets revisit once that lands in this PR.  I think it will make it much more palatable in that regard.

As to the usefulness, it sounds like this should then go together will perhaps running the httpx tests w/ and w/o trio?  how does this sound?  that would then mean aiobotocore is compatible (at least informally) w/ trio and the httpx backend.

agronholm:
> From what I read of Trio, expanding support (albeit most likely informal), much like httpx did, of it to aiobotocore is a lofty goal and requires usage of anyio

Not strictly true, but it opens the door to future event loop back-ends supported by AnyIO. And if you feel adventurous enough, you could even switch to AnyIO's APIs which eliminates a lot of duplication. If you're worried about the extra run-time dependency, you should know that my local changes eliminate the `aioitertools` dependency (which cannot be used as it too only works with asyncio).

> lets revisit once that lands in this PR. I think it will make it much more palatable in that regard.

Given that it's currently just a development dependency, I could just add the current master version to the `dev` group.

> as to the usefulness, it sounds like this should then go together will perhaps running the httpx tests w/ and w/o trio? how does this sound? that would then mean aiobotocore is compatible (at least informally) w/ trio and the httpx backend.

Yes, my local changes already parametrize the tests (or try to) in such a way that they're run against both Trio and asyncio (aiohttp + asyncio, httpx + asyncio, httpx + trio). That's where I realized that httpx was in fact not used by most tests. Automatically selecting `aiohttp` on asyncio and `httpx` on Trio would, however, simplify this a lot, as I don't see much value in cherry picking the HTTP client on asyncio. In my personal opinion, the HTTP client should be an implementation detail.

agronholm:
> > lets revisit once that lands in this PR. I think it will make it much more palatable in that regard.
> 
> Given that it's currently just a development dependency, I could just add the current master version to the `dev` group.

I just pushed this update which considerably reduces the surface area of this PR.

thehesiod:
now this is something I can get behind :)

thehesiod (dismissed):

thehesiod:
@agronholm looks like you need to force push with signed commits

pyproject.toml
```diff
@@ -72,14 +72,14 @@ botocore-dev = [
     # "colorama>0.3.0", # Windows requirement
 ]
 dev = [
+    "anyio @ https://github.com/agronholm/anyio.git",
```
jakob-keller: @thehesiod: Are we sure, we want to introduce this unpinned GIT dependency? This could be considered a security issue to anyone working as a developer of the project. At least pin it to a commit hash, please. And we should double check that source to be on the safe side before merging the PR.

jakob-keller (dismissed):
Do we need any documentation changes as part of this?

pyproject.toml
agronholm: Isn't that why you use `uv`?

agronholm: Ref: https://github.com/aio-libs/aiobotocore/pull/1399/files#diff-84321598744d84dbee2318e634c74c9aae39a1c253f1c4bd17ebf9ef2f807b11R249

jakob-keller: Apologies, you are correct: The uv lock file includes the pin I was expecting.

Still, we should double check that source to be on the safe side before merging the PR. Have you done that @thehesiod?

agronholm: > Apologies, you are correct: The uv lock file includes the pin I was expecting.
> 
> Still, we should double check that source to be on the safe side before merging the PR. Have you done that @thehesiod?

While this is obviously not a task for me, I am curious as to what you'd be looking for there.

jakob-keller: The diff relative to upstream would be what I‘d be interested in.

agronholm: Not sure what you're talking about. What upstream?

jakob-keller: anyio

agronholm: My repo is the authoritative repo for AnyIO. I am its author and maintainer.

jakob-keller: In that case: apologies again! I wasn’t aware. I’m really not qualified to comment on this PR.

agronholm:
> @agronholm looks like you need to force push with signed commits

This doesn't look like a technical requirement, nor is it mentioned in the contributions doc. Can you clarify? Won't the commits be auto-squashed on merge anyway?

agronholm:
> > @agronholm looks like you need to force push with signed commits
> 
> This doesn't look like a technical requirement, nor is it mentioned in the contributions doc. Can you clarify? Won't the commits be auto-squashed on merge anyway?

Never mind, I saw the text in the "Merging is blocked" section.

agronholm:
I did an interactive rebase and signed all the commits with `git commit --amend -S --no-edit`. Why is it still complaining about the signatures?

agronholm:
It seems that no matter what I try, git ignores my `-S` switch and proceeds to commit w/o a signature. I am losing hope.

agronholm:
I've googled myself blue about this issue, even asked an AI which gave me contradictory instructions (because it can't think), and even `strace` tells me that git is not even looking at the key I provided in the config.

agronholm:
It turns out that if you haven't specified an "allowed signers" file, it will just report the commit as having no signature at all. So my commits were in fact signed. Now I need to just understand why Github doesn't agree with the signatures.

agronholm:
Phew, I needed to add my SSH key again as a signing key. That did it!

agronholm:
@jakob-keller Just a heads up, your change request is still blocking the merge.

jakob-keller:
> @jakob-keller Just a heads up, your change request is still blocking the merge.

I can never figure this out: How do I clear that stale review, short of approving?

jakob-keller:
> > @jakob-keller Just a heads up, your change request is still blocking the merge.
> 
> I can never figure this out: How do I clear that stale review, short of approving?

@thehesiod: Could you dismiss my review, please? You should be able to follow those steps: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review

agronholm:
> > > @jakob-keller Just a heads up, your change request is still blocking the merge.
> > 
> > 
> > I can never figure this out: How do I clear that stale review, short of approving?
> 
> @thehesiod: Could you dismiss my review, please? You should be able to follow those steps: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review

Does this not work? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review

jakob-keller:
> > > > @jakob-keller Just a heads up, your change request is still blocking the merge.
> > > 
> > > 
> > > I can never figure this out: How do I clear that stale review, short of approving?
> > 
> > 
> > @thehesiod: Could you dismiss my review, please? You should be able to follow those steps: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review
> 
> Does this not work? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review

I does not show in my UI. I guess that is related to permissions. I've had this issue before.

agronholm:
On another note, what's your Python version support policy? How long do you plan to still support 3.8? Or 3.9 whose EOL is coming next month?

jakob-keller:
> On another note, what's your Python version support policy?

ASAIK, we attempt to follow upstream (botocore).

> How long do you plan to still support 3.8?

Python 3.8 is no longer supported: https://github.com/aio-libs/aiobotocore/blob/d4eeb849df6cf53253cea7f6c5b396a4051f6781/pyproject.toml#L8

> Or 3.9 whose EOL is coming next month?

Most likely around the time upstream drops support, unless there is a strong reason to deviate. Python EOL is probably not a big concern to us.

agronholm:
Ah, nice. I must've been looking at something else when I saw `python_version >= '3.8'`. I think I'll send another PR which tidies up `pyproject.toml` a bit.

thehesiod:
done

thehesiod:
I'll re review in a few hours after kid drop off

pyproject.toml
thehesiod: we should not rely on github repos, can you instead swap to the latest release version?

thehesiod:
just 1 cmt

pyproject.toml
agronholm: The latest version does not have the `anyio_mode="auto"` feature yet. Otherwise I would've pointed it to the latest version. The commit hash is pinned in `uv.lock` though.

agronholm: So what's it going to be? Do you want me to:

1. Pin the dev dependency to a specific hash in `pyproject.toml`?
2. Wait until the next release of AnyIO?
3. ???

thehesiod: how long until a release is published on pypi?

agronholm: Might be a while since v4.10.0 was just released recently.

agronholm: What I tried to say is that I don't like making frequent releases.

thehesiod: ok let me know when you've swapped to an official release, then can approve

agronholm: Alright, I've released v4.11.0 and switched this PR to use that.

tests/test_dynamodb.py
```diff
@@ -113,17 +113,19 @@ async def test_waiter_table_exists_failure(dynamodb_client):
...
-    await waiter.wait(
-        TableName=table_name, WaiterConfig=dict(Delay=1, MaxAttempts=5)
-    )
+        waiter = dynamodb_client.get_waiter('table_exists')
+        await waiter.wait(
+            TableName=table_name, WaiterConfig=dict(Delay=1, MaxAttempts=5)
+        )
 
-    assert task.done()
+        assert done_event.is_set()
```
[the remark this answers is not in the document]
agronholm: This looks to be invalid. The `_create_table()` task has to finish before `waiter.wait()` completes.

thehesiod:
kewl, copilot working!

thehesiod (approved):

thehesiod:
thanks for the patience, been sick these last several weeks

agronholm:
Thanks!

agronholm:
@thehesiod this needs another go at merging.

CHANGES.rst
```diff
@@ -1,8 +1,11 @@
 Changes
 -------
 
+2.24.4 (TBD)
+^^^^^^^^^^^^
+* switch async test runner from pytest-asyncio to AnyIO
+
 2.24.3 (2025-10-06)
-^^^^^^^^^^^^^^^^^^^
```
thehesiod, line 5: looks like you accidentally removed this line

agronholm: Damnit. Fixed.

PR #1391, opened 2025-08-08
Bump `botocore` dependency specification

jakob-keller:
### Description of Change
This PR intends to improve general compatibility of `aiobotocore` within the Python ecosystem by bumping the dependency specification of `botocore`, as well as `boto3` and `awscli`.

### Assumptions
Upstream contains several changes that require adjustments to the aiobotocore codebase.

### Checklist for All Submissions
* [x] I have added change info to [CHANGES.rst](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.rst)
* [ ] If this is resolving an issue (needed so future developers can determine if change is still necessary and under what conditions) (can be provided via link to issue with these details):
  * [ ] Detailed description of issue
  * [ ] Alternative methods considered (if any)
  * [ ] How issue is being resolved
  * [ ] How issue can be reproduced
* [ ] If this is providing a new feature  (can be provided via link to issue with these details):
  * [ ] Detailed description of new feature
  * [ ] Why needed
  * [ ] Alternatives methods considered (if any)

### Checklist when updating botocore and/or aiohttp versions

* [x] I have read and followed [CONTRIBUTING.rst](https://github.com/aio-libs/aiobotocore/blob/master/CONTRIBUTING.rst#how-to-upgrade-botocore)
* [x] I have updated test_patches.py where/if appropriate (also check if no changes necessary)
* [x] I have ensured that the awscli/boto3 versions match the updated botocore version
* [x] I have added URL to diff: https://github.com/boto/botocore/compare/1.39.11...1.39.12

PR #1405, opened 2025-09-07
Fix compatibility with botocore v1.40.19+

carlsmedstad:
The botocore method `RequestSigner.check_and_register_feature_id()` was removed in https://github.com/boto/botocore/pull/3547.

From what I can tell, calling it should no longer be necessary during signing since credentials are registered on retrieval.

### DIFF
--- a/aiobotocore/signers.py
+++ b/aiobotocore/signers.py
@@ -158,10 +158,6 @@ class AioRequestSigner(RequestSigner):
             return auth
 
         credentials = request_credentials or self._credentials
-        if credentials and (
-            cred_method := getattr(credentials, 'method', None)
-        ):
-            self.check_and_register_feature_id(cred_method)
         if getattr(cls, "REQUIRES_IDENTITY_CACHE", None) is True:
             cache = kwargs["identity_cache"]
             key = kwargs["cache_key"]


jakob-keller:
Thanks for submitting this PR. Please take a look at https://github.com/aio-libs/aiobotocore/blob/master/CONTRIBUTING.rst#how-to-upgrade-botocore and check out #1391 as an example of what is required to bump the requirement specification of botocore for this project. If you decide to go ahead with this PR, I am happy to provide guidance and review as soon as the PR is complete. Also, we have a template that should be used when submitting PRs.

carlsmedstad:
Hey - Sorry for skipping the template & thanks for the quick and nice response.
I realize that introducing compatibility is a larger undertaking than my PR, which I'm afraid I don't have the bandwidth for now. I stumbled on the incompatibility when packaging the latest botocore+aiobotocore for Arch Linux and thought I'd share it and the suggested solution, but an issue might have been more appropriate.
Feel free to close this PR, cheers!

jakob-keller:
No worries, we will pick this up as part of a future release for sure.

jakob-keller:
In case, you're interested: Here's a full PR that includes the removal of `RequestSigner.check_and_register_feature_id()`: #1409

PR #1428, opened 2025-10-28
Bump botocore dependency

claytonparnell:
### Description of Change

### Assumptions

  * [ ] Why needed

* [ ] I have added URL to diff: https://github.com/boto/botocore/compare/1.40.49...1.40.61

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -340,7 +340,7 @@ wheels = [
 
 [[package]]
 name = "awscli"
-version = "1.42.60"
+version = "1.42.61"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "botocore" },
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -31,7 +31,7 @@ dynamic = ["version", "readme"]
 dependencies = [
     "aiohttp >= 3.9.2, < 4.0.0",
     "aioitertools >= 0.5.1, < 1.0.0",
-    "botocore >= 1.40.46, < 1.40.50", # NOTE: When updating, always keep `project.optional-dependencies` aligned
+    "botocore >= 1.40.46, < 1.40.61", # NOTE: When updating, always keep `project.optional-dependencies` aligned
```
jakob-keller: I believe 1.40.61 was recently shipped and does not include any breaking changes for us. Could we relax the dependency specification further, i.e. to `< 1.40.62`?

CHANGES.rst
```diff
@@ -1,6 +1,10 @@
 Changes
 -------
 
+2.25.1 (2025-10-28)
+^^^^^^^^^^^^^^^^^^^
+* bump botocore dependency specification
```
jakob-keller: Would you mind saying `relax botocore dependency specification`?

pyproject.toml
jakob-keller: Now we only need to update the lock file one more time...

claytonparnell: right right

jakob-keller:
Thanks for the PR!

claytonparnell:
No problem!

Should probably be separate from this PR, but could try adding `@pytest.mark.flaky(reruns=3)` ([ref](https://pytest-rerunfailures.readthedocs.io/latest/mark.html)) to `tests/test_sns.py::test_publish_to_http`, since it seems to be intermittent faliures? But haven't looked into why the actual failure is coming.

claytonparnell:
Thanks for the quick review/merge!

PR #1433, opened 2025-11-10
Relax botocore dependency specification

claytonparnell:
### Description of Change

### Assumptions

  * [ ] Why needed

* [ ] I have added URL to diff: https://github.com/boto/botocore/compare/1.40.61...1.40.70

### DIFF
--- a/tests/test_patches.py
+++ b/tests/test_patches.py
@@ -300,6 +300,7 @@ def test_protocol_parsers():
         (
             Config,
             {
+                'da2d8b797568604449695cdcf803e344bdf887eb',
                 'a3abd6cb69883934f48403acfaa5ef598c548982',
             },
         ),


tests/test_patches.py
```diff
@@ -300,7 +300,7 @@ def test_protocol_parsers():
         (
             Config,
             {
-                'da2d8b797568604449695cdcf803e344bdf887eb',
+                'a3abd6cb69883934f48403acfaa5ef598c548982',
```
jakob-keller, line 304: Since the upstream change only affected the docstring, we should keep the previous hash and add the new one to the set.

claytonparnell: Makes sense, done

jakob-keller (approved):

PR #1440, opened 2025-11-28
Remove `boto3` packaging extra

### LINKED ISSUE #1252: Automate check for consistent upstream dependencies
By the way, you should be able to write a simple check built into the pre-commit framework, verifying that these values are the same. Here's an example of embedding a few lines of python there: https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/648f815/.pre-commit-config.yaml#L45-L74.

_Originally posted by @webknjaz in https://github.com/aio-libs/aiobotocore/pull/1246#discussion_r1900126947_

### LINKED ISSUE #1424: Remove `awscli` and `boto3` optional dependencies
We currently support and advertise installing aiobotocore along with supported versions of awscli and boto3 using optional dependencies: 

`pip install -U 'aiobotocore[awscli,boto3]'`

I view this primarily as a historic legacy and propose to remove those optional dependencies on the following grounds:

* modern pip and equivalent tools (uv, poetry, ?) should be fully capable of resolving the dependency graph and installing compatible sets of dependencies by merely running `pip install -U aiobotocore awscli boto3` or equivalent. I am unaware of any need for those optional dependencies.
* awscli is pinning to 1.x releases of the AWS CLI. AWS CLI 2.x has been around for years. AWS CLI 1.x might not be relevant anymore for the majority of users.
* The awscli extra is not beening tested as part of the development lifecycle, AFAIK.
* Keeping those extra dependency pins aligned with the pin on botocore is a recurring, manual task that complicates the process of bumping the supported botocore version range.

Any thoughts from users and maintainers?

If we reach consensus on this proposal, I would be happy to prepare a PR to include the change in the upcoming 3.0.0 release.

jakob-keller:
### Description of Change
Remove `boto3` packaging extra

### Assumptions
The discussion around #1424 has not yielded substantial concerns with regards to removing the boto3 packaging extra. This PR makes the proposed change concrete and may act as another opportunity for critical discussion or feedback.

* [x] If this is resolving an issue (needed so future developers can determine if change is still necessary and under what conditions) (can be provided via link to issue with these details): closes #1252, closes #1424
  * [ ] Why needed

* [ ] I have ensured that the boto3 version matches the updated botocore version

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -20,5 +20,4 @@
 
 * [ ] I have read and followed [CONTRIBUTING.rst](https://github.com/aio-libs/aiobotocore/blob/master/CONTRIBUTING.rst#how-to-upgrade-botocore)
 * [ ] I have updated test_patches.py where/if appropriate (also check if no changes necessary)
-* [ ] I have ensured that the boto3 version matches the updated botocore version
 * [ ] I have added URL to diff: https://github.com/boto/botocore/compare/[CURRENT_BOTO_VERSION_TAG]...[NEW_BOTO_VERSION_TAG]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ Changes
 ^^^^^^^^^^^^^^^^^^^
 * BREAKING: forbid creating loose ``ClientSession`` when ``AioBaseClient`` exits context
 * BREAKING: remove `awscli` packaging extra. Instead of ``pip install aiobotocore[awscli]``, use ``pip install aiobotocore awscli`` or similar to install compatible versions of `aiobotocore`, `botocore` and `awscli`.
+* BREAKING: remove `boto3` packaging extra. Instead of ``pip install aiobotocore[boto3]``, use ``pip install aiobotocore boto3`` or similar to install compatible versions of `aiobotocore`, `botocore` and `boto3`.
 * relax botocore dependency specification
 
 2.26.0 (2025-11-27)
--- a/README.rst
+++ b/README.rst
@@ -216,13 +216,3 @@ Requirements
 .. _Pylance: https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance
 .. _pyright: https://github.com/microsoft/pyright
 .. _mypy: http://mypy-lang.org/
-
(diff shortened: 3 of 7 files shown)


jakob-keller:
Reviewers: I intend to keep this PR open for a week or so to give everyone the opportunity to chime in.

jakob-keller:
@terricain: aioboto3 is currently relying on this packaging extra. Does your project or its users still have a need for it that I’m unaware of? I assume the situation is similar to what’s being [discussed over at s3fs](https://github.com/fsspec/s3fs/issues/994).

terricain:
Yeah `aioboto3` makes use of this. I suppose this could be removed but I'll then have to pin it myself, as like `aiobotocore` had some version constraints on `botocore` as it patches methods, `aioboto3` would need to do the same. 😄 it'll mean a bit more work for me but if that simplifies this here, its fine.

jakob-keller:
> Yeah `aioboto3` makes use of this. I suppose this could be removed but I'll then have to pin it myself, as like `aiobotocore` had some version constraints on `botocore` as it patches methods, `aioboto3` would need to do the same. 😄 it'll mean a bit more work for me but if that simplifies this here, its fine.

I understand. IMO, if aioboto3 directly relies on a specific botocore version, it should make that dependency explicit. Note that this dependency specification may well be different from aiobotocore's. On the user side, modern dependency resolvers should then be able to resolve a consistent set of compatible package versions taking the user's environment into account.

jakob-keller:
OK, now it's been a full week and my understanding is that, despite outreach to [aioboto3](https://github.com/aio-libs/aiobotocore/pull/1440#issuecomment-3589787292) and [s3fs](https://github.com/fsspec/s3fs/issues/994) as well as [performing various tests](https://github.com/aio-libs/aiobotocore/issues/1424), no issues with this PR could be substantiated.

In case anything unexpected breaks for users after the release of aiobotocore 3.0.0, there is always the [possibility of reintroducing the boto3 packaging extra](https://github.com/aio-libs/aiobotocore/issues/1424#issuecomment-3600140839).

Please submit an approving review so I can merge the PR.

Thanks to everyone who weighed in!

webknjaz (approved):

Issue #1450, opened 2025-12-08
Allow passing socket_factory in connector_args


tomasfarias:
## Problem

* aiohttp supports a `socket_factory` parameter to configure socket settings: https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.TCPConnector.
  * This was added in aiohttp 3.12.
* Trying to pass `socket_factory` via `AioConfig` as a key in `connector_args` fails the validation check in `AioConfig._validate_connector_args`.
  * Using latest aiobotocore.

This is similar to #1258, which introduced the `ttl_dns_cache` argument.

## Alternatives

One can get around this by subclassing `AioConfig` to ignore the check for `socket_factory`, but it would be better if this was just accepted by aiobotocore. 

## Development

Happy to send a PR if this is considered a valid issue.

Thank you!

PR #1451, opened 2025-12-09
Add support for executor configuration in AioConfig for file loading

### LINKED ISSUE #1199: Detected blocking call inside the event loop
**Describe the bug**
Home Assitant recently added the ability to detect I/O function that block event loop:

```
2024-06-12 09:40:41.237 WARNING (MainThread) [homeassistant.util.loop] Detected blocking call to listdir inside the event loop by integration 'aws'
at homeassistant/components/aws/notify.py, line 38: return await session.get_available_regions(service) (offender: /usr/local/lib/python3.12/site-pa
ckages/botocore/loaders.py, line 311: api_versions = os.listdir(full_dirname)), please create a bug report at https://github.com/home-assistant/core
/issues?q=is%3Aopen+is%3Aissue+label%3A%22integration%3A+aws%22
```
After trying a first fix in #1120, I was redirected to botocore.
I opened a issue there: https://github.com/boto/botocore/issues/3222, but was closed as botocore doesn't use event loop at all.
After some more discussions in the original PR, a new idea was raised and I applied it to a new fix in #1196.
Lately I was asked to open a issue to discuss the approach.

**Checklist**
- [ ] I have tried the same code in botocore to ensure this is an aiobotocore specific issue
- [ ] I have tried similar code in aiohttp to ensure this is is an aiobotocore specific issue

**pip freeze results**
[pip_freeze.txt](https://github.com/user-attachments/files/16872437/pip_freeze.txt)

**Environment:**
 - Python Version:  3.12.4
 - OS name and version: alpine-release 3.20.1

**Additional context**

thehesiod:
### Description of Change
Added the ability to configure a specific executor in `AioConfig` for handling file load events. This enhancement enables greater flexibility and control for users managing asynchronous tasks.

### Assumptions
No additional assumptions were made.

  * [x] Why needed


thehesiod:
python 3.9 is EOL, removing support to fix tests

bdraco:
Sure looks like it will solve the issue. Nice work!

@chemelli74 and @zweckj will need to confirm since they have the test cases.

jakob-keller:
Thank you for taking the initiative!

I would like to clarify the proposed design, before providing an in-depth review.

In my understanding, this PR:
1. Introduces `load_executor` in `AioConfig`
2. Provides `optionally_run_in_executor()` to run a (blocking) function in an executor, if provided
3. Uses 1. and 2. to avoid blocking calls in `AioClientCreator.create_client()`

Pros:
* May be used by HA community to resolve their reported issues
* Only affects client creation - no impact on making API calls

Cons:
1. No comprehensive solution to blocking loaders: Does not remove all blocking codepaths from `create_client()` (e.g. `_register_retries()` is potentially blocking). Leaves majority of affected codepaths untouched (e.g. waiters, paginators, ...)
2. Does not leverage instance caches already in place in `botocore.loaders`: executor invocations could be avoided for any subsequent cache hits
3. Relies on substantial modification of `create_client()` - will not be simple to extend to further blocking codepaths without incurring undue drift from upstream and violating DRY principle
4. Requires explicit configuration - likely will not benefit the majority of users

**Alternative proposal:**

UPDATE: #1452 is provided as proof-of-concept of the following, alternative design proposal.

I was thinking of a cache-aware wrapper for blocking loader functions, such as:

```python
# to be placed in aiobotocore/loaders.py
async def call_cached(loader, func, *args, **kwargs):
    # key building logic taken from botocore.loaders
    key = (func.__name__,) + args
    for pair in sorted(kwargs.items()):
        key += pair

    if key in loader._cache:
        return loader._cache[key]

    return await asyncio.to_thread(func, loader, *args, **kwargs)
```

This could be used in light-weight patches wherever we decide to provide non-blocking behavior (everywhere!?), e.g.:

```python
class AioClientCreator(ClientCreator):
    # ...
    async def _load_service_model(self, service_name, api_version=None):
        json_model = await call_cached(
            self._loader,
            self._loader.load_service_model,
            service_name,
            'service-2',
            api_version=api_version,
        )
        service_model = ServiceModel(json_model, service_name=service_name)
        return service_model
```

I believe this would provide more potential to deliver full non-blocking loaders. Also note that no client configuration would be needed, if we agree that a cache-aware implementation would eliminate the need for user opt-in and use the recommended, high-level `asyncio.to_thread()` API.

Nitpick 1: I welcome introducing type hints such as those enabled by `_ConnectorArgsType` and `_HttpSessionTypes`. Could we separate that work out into a separate PR since it appears to be unrelated to unblocking loaders? This could also be related to #1450.

Nitpick 2: [botocore provides support for Python 3.9 until 2026-04-29](https://github.com/boto/botocore?tab=readme-ov-file#notices). Are we sure we want to drop support before then? If so, it would simplify this PR if we do this in a separate PR. I'd be happy to prepare that, just let me know.

jakob-keller:
#1452 is provided as proof-of-concept of the alternative design proposal

thehesiod:
ya you're right, we should support 3.9 until then, will remove that part, not as bad given we can use the type_extensions

thehesiod:
Responses for Cons
1. I think for now we only want to solve for things that potentially block due to loading large number of file operations as those are not async.  I don't think we need to solve for places where small files may be loaded as that would be a much larger change.  waiters and paginators don't load large number of files to my understanding (if any).
2. this does not bypass caches already in place in botocore.  It calls the existing caching code, just in another thread.  The only thing is that the cache will be managed from a separate thread now.
3. I'm not sure there are other, and if so very few, places where large number of files are loaded.  If you think about it it's before the client is created that a large number of files need to be read.  I don't understand how wrapping two function calls in a wrapper is a large change to a function we already override.  It's rather minimal in fact
4. to my understanding for most users this is not an issue as they do an upfront load of the client and have it stick around for awhile, so it seems to make sense to me that you'd need to opt-in for better behavior.  I think long term we need to just scrap relying on botocore as a foundation so we can better manage issues like these without dramatically increasing our footprint.  Another option I suppose is pre-warming all these caches ourselves in dedicated threadpool that we throw away.  Actually thinking about this I think that's what we want to do, is manage these caches ourselves via a temporarily thread pool we throw away, then it's ok for everyone.  I'll investigate this a bit in my PR if you want to investigate this in yours too.

jakob-keller:
> 1. I think for now we only want to solve for things that potentially block due to loading large number of file operations as those are not async.  I don't think we need to solve for places where small files may be loaded as that would be a much larger change.  waiters and paginators don't load large number of files to my understanding (if any).

I don't know what those operations are and was working under the assumption that we need to wrap all file IO in a thread, at least eventually. All the more reason to clarify the goal. @chemelli74 and @zweckj: Could the HA community weigh in on what exactly constitutes unacceptable blocking from their point of view?
Waiters and paginators do perform additional file IO on first use and AFTER client creation. Those are just examples and there might be more blocking uses of `Loader` than that. 

> 2. this does not bypass caches already in place in botocore.  It calls the existing caching code, just in another thread.  The only thing is that the cache will be managed from a separate thread now.

Correct, but being aware of the cache could be used to avoid paying the overhead of running code in a separate thread, as demonstrated by #1452. This is not a big issue in `create_client()`, but could be useful for further codepaths that need to be non-blocking. I am thinking of a massively concurrent application (number of asynchronous jobs making API calls >> number of threads available to executor). As long as this behavior is opt-in we might be good, but it would be better if a design allowed for out-of-the-box non-blocking behavior.

> 4. to my understanding for most users this is not an issue as they do an upfront load of the client and have it stick around for awhile, so it seems to make sense to me that you'd need to opt-in for better behavior.  I think long term we need to just scrap relying on botocore as a foundation so we can better manage issues like these without dramatically increasing our footprint.  Another option I suppose is pre-warming all these caches ourselves in dedicated threadpool that we throw away.  Actually thinking about this I think that's what we want to do, is manage these caches ourselves via a temporarily thread pool we throw away, then it's ok for everyone.  I'll investigate this a bit in my PR if you want to investigate this in yours too.

aiobotocore / botocore already supports passing in a custom loader. As I had originally proposed, HA could "simply" initialize and pre-warm a loader (in a thread to prevent blocking) and then have their clients use that without having to worrying about blocking. If that is not feasible, we could provide a utility function to initialize and pre-warm a loader for such a purpose.

thehesiod:
the loader case is interesting, since you have context could you post a little pseudo code of how that could work?

jakob-keller:
> the loader case is interesting, since you have context could you post a little pseudo code of how that could work?

```python
import asyncio

import aiobotocore.session

from botocore.loaders import create_loader


def pre_warm_loader(
    loader,
    /,
    service_name,
    api_version=None,
):
    # from session.py
    loader.load_data_with_path('endpoints')
    loader.load_data('sdk-default-configuration')
    loader.load_service_model(
        service_name, 'waiters-2', api_version
    )
    loader.load_service_model(
        service_name, 'paginators-1', api_version
    )
    loader.load_service_model(
        service_name, type_name='service-2', api_version=api_version
    )
    loader.list_available_services(
        type_name='service-2'
    )

    # from client.py
    loader.load_data('partitions')
    loader.load_service_model(
        service_name, 'service-2', api_version=api_version
    )
    loader.load_service_model(
        service_name, 'endpoint-rule-set-1', api_version=api_version
    )
    loader.load_data('_retry')

    # from docs/service.py
    loader.load_service_model(
        service_name, 'examples-1', api_version
    )


async def main():
    loader = create_loader()
    await asyncio.to_thread(pre_warm_loader, loader, "s3")

    session = aiobotocore.session.get_session()
    session.register_component("data_loader", loader)

    async with session.create_client("s3") as client:  # should not block
        await client.list_buckets()
```

jakob-keller:
It should be feasible to perform such pre-warming steps optionally in `session.create_client()`:

```async with session.create_client("s3", pre_load=True) as client:```

The default would initially be `pre_load=False` to preserve current behavior.

`AioConfig` could be used to influence behavior if the `pre_load` argument is not passed to `create_client()` and could also be used to specify a custom executor to be used, if that's a requirement.

If that turns out to be useful and performant, `pre_load=True` could become the default behavior in aiobotocore 4.0.0.

**Comment by thehesiod** [comment]
@jakob-keller sorry for delays, really want to get back to this, will try to care out some time to think about this some more

**Comment by chemelli74** [comment]
Did a quick test, and get:

```
  File "/workspaces/core/homeassistant/components/aws/__init__.py", line 8, in <module>
    from aiobotocore.config import AioConfig
  File "/home/vscode/.local/ha-venv/lib/python3.13/site-packages/aiobotocore/config.py", line 12, in <module>
    from .endpoint import DEFAULT_HTTP_SESSION_CLS
  File "/home/vscode/.local/ha-venv/lib/python3.13/site-packages/aiobotocore/endpoint.py", line 17, in <module>
    from aiobotocore.httpchecksum import handle_checksum_body
  File "/home/vscode/.local/ha-venv/lib/python3.13/site-packages/aiobotocore/httpchecksum.py", line 3, in <module>
    from botocore.httpchecksum import (
    ...<9 lines>...
    )
ImportError: cannot import name '_register_checksum_algorithm_feature_id' from 'botocore.httpchecksum' (/home/vscode/.local/ha-venv/lib/python3.13/site-packages/botocore/httpchecksum.py)
```

even if `def _register_checksum_algorithm_feature_id(algorithm):` is there in **botocore/httpchecksum.py**.
Using  v1.42.1 according to pip show

**Comment by thehesiod** [comment]
@chemelli74 this is most likely your botocore does not match the aiobotocore version requirement

**Comment by thehesiod** [comment]
ok did some research with AI and it seems to think along my lines that 1452 puts its fingers on too much internals when we have a solution that should work at a high level for the problem at hand and is opt-in so it can provide valuable feedback. Also the slowdown would be minimal given they'll be passing in their own executor which could be pre-spawned.  Also I did some research in that there's no docs stating that thread workers are GC'd, and, it keeps around 5 threads already once you touch the default work pool (as I previously saw).  It will re-use existing threads if they exist but that's just the definition of a pool.  Also, I think we want to give the option for clients to restrict how many threads are spawned across multiple sessions and clients if they so please so I think this is would be a great enhancement.  I've gone ahead and reverted the 3.9 changes so I think this PR should be good to go.  thoughts?

**Comment by jakob-keller** [comment]
> Nitpick 1: I welcome introducing type hints such as those enabled by `_ConnectorArgsType` and `_HttpSessionTypes`. Could we separate that work out into a separate PR since it appears to be unrelated to unblocking loaders? This could also be related to #1450.

I took the liberty and added type annotations to `AioConfig` as part of #1454

**Comment by chemelli74** [comment]
> @chemelli74 this is most likely your botocore does not match the aiobotocore version requirement

As I wrote is `v1.42.1` according to pip show and as far as I can tell zhould be fine.
Isn't it ?

File: uv.lock
Comment (jakob-keller) [inline uv.lock:1]: These may be dropped from the PR.

**Comment by thehesiod** [comment]
@chemelli74 yes but you didn't say what version of aiobotocore, works fine for me:
```bash
$ python3
Python 3.11.11 (main, Jan 14 2025, 23:36:41) [Clang 19.1.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from aiobotocore.config import AioConfig
>>> import aiobotocore
>>> aiobotocore.__version__
'3.1.0'
>>> import botocore
>>> botocore.__version__
'1.42.19'
>>> 
```

thehesiod:
@jakob-keller ok updated and merged in your PR

aiobotocore/config.py
jakob-keller, line 1: Looks like there are a few merge related issues in here.

aiobotocore/config.py
```diff
@@ -34,6 +35,7 @@ class _ConnectorArgs(TypedDict):
     ssl_context: NotRequired[ssl.SSLContext]
     resolver: NotRequired[AbstractResolver]
     socket_factory: NotRequired[Optional[SocketFactoryType]]
+    resolver: NotRequired[AbstractResolver]
```
jakob-keller: Duplicate of line 36.

aiobotocore/config.py
```diff
@@ -44,17 +46,24 @@ def __init__(
         self,
         connector_args: Optional[_ConnectorArgs] = None,
         http_session_cls: type[_HttpSessionType] = DEFAULT_HTTP_SESSION_CLS,
+        load_executor: Optional[ThreadPoolExecutor] = None,
         **kwargs,
     ):
         super().__init__(**kwargs)
 
+        self._validate_connector_args(connector_args, http_session_cls)
```
jakob-keller, line 55: Are changes from #1454 reverted on purpose? Why?

aiobotocore/config.py
```diff
@@ -44,17 +46,24 @@ def __init__(
...
+                report='load_executor value must be an instance of an Executor.'
+            )
+
+        self.load_executor = load_executor
+
         self.connector_args: _ConnectorArgs = (
             copy.copy(connector_args) if connector_args else {}
         )
         self.http_session_cls: type[_HttpSessionType] = http_session_cls
-        self._validate_connector_args(
```
jakob-keller, line 55: Are changes from #1454 reverted on purpose? Why?

aiobotocore/config.py
```diff
@@ -72,6 +81,9 @@ def _validate_connector_args(
         connector_args: _ConnectorArgs,
         http_session_cls: type[_HttpSessionType],
     ) -> None:
+        if connector_args is None:
```
jakob-keller, line 87: Are changes from #1454 reverted on purpose? Why?

aiobotocore/config.py
```diff
@@ -44,17 +46,24 @@ def __init__(
...
         connector_args: Optional[_ConnectorArgs] = None,
         http_session_cls: type[_HttpSessionType] = DEFAULT_HTTP_SESSION_CLS,
+        load_executor: Optional[ThreadPoolExecutor] = None,
         **kwargs,
     ):
         super().__init__(**kwargs)
 
+        self._validate_connector_args(connector_args, http_session_cls)
+
+        if load_executor and not isinstance(load_executor, ThreadPoolExecutor):
```
jakob-keller: Maybe move to end of method?

aiobotocore/_helpers.py
```diff
@@ -14,3 +17,15 @@ async def async_any(items):
             return True
 
     return False
+
+
+async def optionally_run_in_executor(
+    loop: AbstractEventLoop,
+    executor: Optional[ThreadPoolExecutor],
```
jakob-keller, line 26: [run_in_executor()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) accepts instances of [concurrent.futures.Executor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor) and uses the default executor if `executor` is `None`. It might reduce friction, if we follow the standard library's approach.

Then again, the function would be reduced to line 29 and could be dropped entirely.

aiobotocore/config.py
```diff
@@ -44,17 +46,24 @@ def __init__(
         self,
         connector_args: Optional[_ConnectorArgs] = None,
         http_session_cls: type[_HttpSessionType] = DEFAULT_HTTP_SESSION_CLS,
+        load_executor: Optional[ThreadPoolExecutor] = None,
```
jakob-keller: `load_executor: Union[bool, concurrent.futures.Executor] = False,`

Would be more flexible:
* `False` (default): keep current behavior
* `True`: run in default executor, i.e. using `run_in_executor(None, ...)` or `asyncio.to_thread(...)`
* instance of `Executor`: use `run_in_executor(executor, ...)`

aiobotocore/client.py
jakob-keller, line 1: How about remaining blocking file I/O? Is the scope of this PR limited to `_load_service_model()` and `_load_service_endpoints_ruleset()` on purpose?

jakob-keller:
> It should be feasible to perform such pre-warming steps optionally in `session.create_client()`:
> 
> `async with session.create_client("s3", pre_load=True) as client:`
> 
> The default would initially be `pre_load=False` to preserve current behavior.
> 
> `AioConfig` could be used to influence behavior if the `pre_load` argument is not passed to `create_client()` and could also be used to specify a custom executor to be used, if that's a requirement.
> 
> If that turns out to be useful and performant, `pre_load=True` could become the default behavior in aiobotocore 4.0.0.

#1462 sketches this alternative design. I think it is cleaner and more comprehensive. What are your thoughts?

aiobotocore/client.py
thehesiod: These are the main pain points to my understanding, we can always expand upon this later

aiobotocore/config.py
thehesiod: I want to validate before assigning any variables

aiobotocore/config.py
thehesiod: there's nothing reverted, just the order has changed, validate before assignment

aiobotocore/config.py
thehesiod: validate before assignment

aiobotocore/config.py
thehesiod: fixed thanks

aiobotocore/config.py
thehesiod: i guess technically yes put back the null check

aiobotocore/config.py
thehesiod: added support for default executor via sentinal

thehesiod:
> > It should be feasible to perform such pre-warming steps optionally in `session.create_client()`:
> > `async with session.create_client("s3", pre_load=True) as client:`
> > The default would initially be `pre_load=False` to preserve current behavior.
> > `AioConfig` could be used to influence behavior if the `pre_load` argument is not passed to `create_client()` and could also be used to specify a custom executor to be used, if that's a requirement.
> > If that turns out to be useful and performant, `pre_load=True` could become the default behavior in aiobotocore 4.0.0.
> 
> #1462 sketches this alternative design. I think it is cleaner and more comprehensive. What are your thoughts?

we should get feedback from community, @bdraco @chemelli74 .  No reason both can't exist too I think.  The warm-up one seems like something people can do on their own tho

aiobotocore/config.py
jakob-keller: `if connector_args:` would be slightly more efficient

jakob-keller:
> The warm-up one seems like something people can do on their own tho

I agree, but they seemed to have struggled with it, probably since it's not at all obvious how proper cache warming would need to be performed. My PR is aiming to provide a smooth experience for users who wish to do so.

chemelli74:
> @chemelli74 yes but you didn't say what version of aiobotocore, works fine for me:

Obviously this PR code so:
 `pip install "git+https://github.com/aio-libs/aiobotocore.git@refs/pull/1451/head#aiobotocore==3.1.0"`

thehesiod:
> > @chemelli74 yes but you didn't say what version of aiobotocore, works fine for me:
> 
> Obviously this PR code so: `pip install "git+https://github.com/aio-libs/aiobotocore.git@refs/pull/1451/head#aiobotocore==3.1.0"`

still works for me with that version of botocore:
```python
>>> from aiobotocore.config import AioConfig
>>> import aiobotocore
>>> import botocore
>>> aiobotocore.__version__
'3.1.0'
>>> botocore.__version__
'1.42.1'
```
you should do a runtime version check, may have a pathing issue.  And not obvious, you could always apply this patch to older versions

thehesiod:
lets wait until we hear back from @bdraco @chemelli74

thehesiod:
I'm going to close this for now, went with your pr.  Can revive if people are interested

PR #1452, opened 2025-12-10
Async loading

jakob-keller:
### Description of Change

### Assumptions

  * [ ] Why needed

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,7 +3,7 @@ Changes
 
 3.1.0 (2025-12-10)
 ^^^^^^^^^^^^^^^^^^^
-* patch ``AioClientCreator.create_client_class()`` and ``AioClientCreator._load_service_model()``
+* patch ``AioClientCreator.create_client_class()``, ``AioClientCreator._load_service_model()`` and ``AioClientCreator._load_service_endpoints_ruleset()``
 
 3.0.0 (2025-12-09)
 ^^^^^^^^^^^^^^^^^^^
--- a/aiobotocore/client.py
+++ b/aiobotocore/client.py
@@ -54,8 +54,10 @@ class AioClientCreator(ClientCreator):
             service_name, api_version
         )
         try:
-            endpoints_ruleset_data = self._load_service_endpoints_ruleset(
-                service_name, api_version
+            endpoints_ruleset_data = (
+                await self._load_service_endpoints_ruleset(
+                    service_name, api_version
+                )
             )
             partition_data = self._loader.load_data('partitions')
         except UnknownServiceError:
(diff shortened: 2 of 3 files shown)


jakob-keller:
In this proof-of-concept, `instance_cached()` is used to unblock loading of service model and endpoint ruleset data.

[Loading partition data](https://github.com/aio-libs/aiobotocore/blob/f898d218eb79c097993cab3542688d59b84fac2b/aiobotocore/client.py#L56) still poses a challenge, because `Loader.load_data()` is not using the instance cache directly, but indirectly via `.load_data_with_path()`:

https://github.com/boto/botocore/blob/89cfb9f40ba4db6249e62337a02d858564ff7524/botocore/loaders.py#L437-L473

If applied to `load_data`, the proposed wrapper would encounter false positive cache misses and call `to_thread` needlessly.

If this is considered an issue (I do), it would require special-casing and/or subclassing `botocore.Loader` to resolve.

Any input would be much appreciated.

aiobotocore/loaders.py
```diff
@@ -0,0 +1,33 @@
...
+    key = (bound_method.__name__,) + args
+    for pair in sorted(kwargs.items()):
+        key += pair
+
+    # check loader instance cache
+    if key in loader._cache:
+        return loader._cache[key]
+
+    # run potentially blocking method in thread
+    return await asyncio.to_thread(bound_method, *args, **kwargs)
```
thehesiod, line 34: I'm not a fan of running in the default thread pool, as soon as you do that it creates Ncpu * multiplier threads, which makes dumps of the thread stack very messy.  I think it should be opt-in via AioConfig like I did in my PR

thehesiod:
just some notes, in this version you need to override several more methods which leads to a higher maintenance burden.

aiobotocore/loaders.py
jakob-keller: In current CPython, that is no longer the case: the default executor is initialized as a ThreadPoolExecutor (the only type permitted). Threads are reused and the pool only grows if all worker threads are busy and their total count is less than the number of *logical* CPUs (usable by the current process) + 4. No multiple is applied anymore.

In case that is problematic or you are stuck with older Python releases, `loop.set_default_executor()` may be used to manually set the default executor, i.e. one with a custom max_workers value. This would be the pythonic way to go and affects any call to `asyncio.to_thread()` or `loop.run_in_executor(executor=None)`.

jakob-keller:
> just some notes, in this version you need to override several more methods which leads to a higher maintenance burden.

That is correct, which is one of the reasons I am not sure this is the way to go. The PR is only meant as a proof of concept for now.

I still find this approach easier to reason about (and thus maintain), because the `instance_cached()` wrapper is applied at the inferface between `Loader` and the remaining codebase. It is obvious that Loader's (blocking) codepaths are being wrapped. Not by coincident does botocore use the instance cache to avoid reloading data from disk. Those are exactly the calls we need to unblock - that's why I try to piggyback on the instance cache!

aiobotocore/loaders.py
thehesiod: I'm thinking perhaps instead we keep our own threadpool executor in the session, thoughts?

jakob-keller: Could you provide a use case where this would be worth the effort? I don't see one.

thehesiod: ya you're right modern python don't need it, just being overly cautious :)

PR #1474, opened 2026-02-14
Relax `botocore` dependency specification

jakob-keller:
### Description of Change
This PR intends to improve general compatibility of `aiobotocore` within the Python ecosystem by relaxing the dependency specification of `botocore`.

### Assumptions
Upstream contains no changes that require adjustments to the aiobotocore codebase. 

  * [ ] Why needed

* [x] I have added URL to diff: https://github.com/boto/botocore/compare/1.42.42...1.42.49


webknjaz (approved):
In `pip-tools`, we're talking about having an extra with an upper bound for the dep we're actually testing. In general, upper boundaries in mandatory metadata is considered harmful (with my PyPA hat on) and such bumps create a lot of noise too causing too much burden (with my serial maintainer hat). For this reason, I tend to configure dep bumping automations (dependabot, pre-commit.ci etc) to be as unfrequent as possible, short of security fixes, of course.

Having to approve this many seemingly pointless bumps every day just steals time from more urgent/useful maintenance needs. I'm approving this but I'll probably change the branch protection to allow self-approvals to try to reduce this burden. Though the correct fix is not having the upper bounds in this place of the metadata.

jakob-keller:
> Having to approve this many seemingly pointless bumps every day just steals time from more urgent/useful maintenance needs. I'm approving this but I'll probably change the branch protection to allow self-approvals to try to reduce this burden.

Makes a lot of sense to me.

> Though the correct fix is not having the upper bounds in this place of the metadata.

aiobotocore by design relies on patching of _private_ `botocore` code. As a result, each `botocore` patch release is potentially breaking for us and our users. IMO, this justifies the tight upper bound on `botocore`, allowing relaxation only after successfully reviewing and testing each subsequent botocore release.

webknjaz:
> aiobotocore by design relies on patching of private botocore code. As a result, each botocore patch release is potentially breaking for us and our users.

`pip-tools` has exactly the same use case. And so we want to let the end-users opt in through an extra but not create busywork. What they need for reproducibility is lock files and testing when they are regenerated.

jakob-keller:
> `pip-tools` has exactly the same use case. And so we want to let the end-users opt in through an extra but not create busywork. What they need for reproducibility is lock files and testing when they are regenerated.

Opened #1475 to pick up your suggestion and continue discussion.

PR #1523, opened 2026-03-20
Fix sync prompt template vars, add MCP commit fallback

thehesiod:
## Summary

Fixes from analyzing the latest sync run ($0.63, 40 turns):

### Fix template variable substitution
`$LATEST_BOTOCORE`, `$CURRENT_UPPER`, etc. appeared literally
in the prompt — env vars were on the `claude-code-action` step
but the `read-prompt` step used `cat` without access to them.
Fix: move env vars to read-prompt step and use `envsubst`.
Saves ~3 turns per run.

### Fix envsubst erasing unintended variables
Bare `envsubst` replaces *all* `$VAR` patterns — including
`$VERSION` used in git commands within the prompt. Added an
explicit variable list so only the 6 intended template vars
are substituted.

### Add MCP commit fallback instructions
`mcp__github_file_ops__commit_files` keeps defaulting to main
despite explicit branch param in prompt. Added fallback
instructions to use GitHub Git Data API directly (`gh api`
for blobs/trees/commits), which also produces signed commits.
Saves ~5 turns when MCP fails.

### Pre-commit fixes
- Split `envsubst` command across lines to stay under
  yamllint's 120-char limit
- Removed invalid `workflows: write` from job permissions
  (not a valid `GITHUB_TOKEN` scope — modifying workflow
  files requires a PAT with the `workflow` scope)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,6 @@ dynamic = ["version", "readme"]
 dependencies = [
     "aiohttp >= 3.12.0, < 4.0.0",
     "aioitertools >= 0.5.1, < 1.0.0",
-    "boto3 >= 1.42.62, < 1.42.71",
     "botocore >= 1.42.62, < 1.42.71",
     "python-dateutil >= 2.1, < 3.0.0",
     "jmespath >= 0.7.1, < 2.0.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ dynamic = ["version", "readme"]
 dependencies = [
     "aiohttp >= 3.12.0, < 4.0.0",
     "aioitertools >= 0.5.1, < 1.0.0",
+    "boto3 >= 1.42.62, < 1.42.71",
     "botocore >= 1.42.62, < 1.42.71",
     "python-dateutil >= 2.1, < 3.0.0",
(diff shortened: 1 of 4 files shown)


thehesiod:
@claude can you address your feedback

thehesiod:
@claude I think we need to update the prompt used in the Botocore Sync workflow so that when you update botocore you update boto3 to the matching range.  Lets add that change to this PR

thehesiod:
reminder to self to allow claude to update .github/workflows/, which requires the workflows permission

pyproject.toml
```diff
@@ -31,6 +31,7 @@ dynamic = ["version", "readme"]
 dependencies = [
     "aiohttp >= 3.12.0, < 4.0.0",
     "aioitertools >= 0.5.1, < 1.0.0",
+    "boto3 >= 1.42.62, < 1.42.71",
```
jakob-keller: We do NOT depend on `boto3`! If necessary, please add to development dependencies.

thehesiod: ya, this was a mistake, working on a new machine it's very "green", cleaning up

thehesiod: resolving as this file was reverted

thehesiod:
merging as this is just claude workflow optimizations

PR #1538, opened 2026-04-07
Add dev environment to Claude CI workflow

thehesiod:
## Summary
- Add `setup-uv` + `uv sync` steps to `claude.yml` so the Claude bot has Python, pytest, and all dev deps pre-installed. Previously it wasted ~3 minutes pip-installing packages one at a time and still failed to run tests.
- Add `.gitattributes` with `eol=lf` to fix yamllint failures on Windows checkouts
- Add Environment and Honesty sections to the CI prompt template — tells Claude to use `uv run`, not `pip install`, and to never claim tests pass without actually running them
- Add Environment section to `CLAUDE.md` with `uv` and botocore source reading guidance

Context: https://github.com/aio-libs/aiobotocore/actions/runs/24069519757 — Claude spent half its run fumbling with pip installs, then falsely claimed tests passed in the PR description.

## Test plan
- [ ] Trigger `@claude` on an issue and verify it can run `uv run pytest` successfully
- [ ] Verify yamllint passes on both Windows and Linux


LICENSE
thehesiod, line 1: this is just whitespace diff due to .gitattributes

thehesiod: fyi @jakob-keller since I want to smash this for the ai fixes, don't be alarmed :)

PR #1462, opened 2026-01-02
Support warm-up of loader caches

jakob-keller:
### Description of Change
Support warm-up of loader caches

### Assumptions
*Replace this text with any assumptions made (if any)*

### Checklist for All Submissions
* [x] I have added change info to [CHANGES.rst](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.rst)
* [x] If this is resolving an issue (needed so future developers can determine if change is still necessary and under what conditions) (can be provided via link to issue with these details): closes #1199
  * [ ] Detailed description of issue
  * [ ] Alternative methods considered (if any)
  * [ ] How issue is being resolved
  * [ ] How issue can be reproduced
* [ ] If this is providing a new feature  (can be provided via link to issue with these details):
  * [ ] Detailed description of new feature
  * [ ] Why needed
  * [ ] Alternatives methods considered (if any)

### Checklist when updating botocore and/or aiohttp versions

* [ ] I have read and followed [CONTRIBUTING.rst](https://github.com/aio-libs/aiobotocore/blob/master/CONTRIBUTING.rst#how-to-upgrade-botocore)
* [ ] I have updated test_patches.py where/if appropriate (also check if no changes necessary)
* [ ] I have added URL to diff: https://github.com/boto/botocore/compare/[CURRENT_BOTO_VERSION_TAG]...[NEW_BOTO_VERSION_TAG]

PR #1539, opened 2026-04-07
Define explicit StreamingBody API, add httpx read(amt) support

### LINKED ISSUE #1365: Specify API for `StreamingBody`
[`botocore.StreamingBody`](https://github.com/boto/botocore/blob/665cd36b3ee67e6a9775ccb3bb22ea3a4128d44c/botocore/response.py#L38) is a wrapper class for [urllib3.response.HTTPResponse](https://urllib3.readthedocs.io/en/stable/reference/urllib3.response.html#urllib3.response.HTTPResponse), which explicitly defines an API that exposes some of the underlying methods and adds additional ones on top. It does not otherwise forward attribute calls to the underlying `HTTPResponse` afaict.

[`aiobotocore.StreamingBody`](https://github.com/aio-libs/aiobotocore/blob/7b81bd9ecc46cbf9318e0eefdb8ae630b641cdc1/aiobotocore/response.py#L19) uses [`wrapt.ObjectProxy`](https://wrapt.readthedocs.io/en/latest/wrappers.html) to wrap [`aiohttp.ClientResponse`](https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponse), which means it forwards all underlying methods, and additionally specifies most of the methods that `botocore.StreamingBody` does. This means it has a larger API, but one that is tightly tied to `aiohttp.ClientResponse`.

The soon-to-be-added [`aiobotocore.HttpxStreamingbody`](https://github.com/aio-libs/aiobotocore/blob/6daaa79983a468003b7429fd7d92a9c1a63e8954/aiobotocore/response.py#L137) is in a tough position. It currently uses `wrapt.ObjectProxy` to wrap [`httpx.Response`](https://www.python-httpx.org/async/#streaming-responses), but `httpx.Response` differs against both of the former in that it requires you to [specify the chunk size up-front](https://github.com/encode/httpx/blob/6c7af967734bafd011164f2a1653abc87905a62b/httpx/_models.py#L982), so unless we write a complicated helper method we cannot expose a `read` method that takes a parameter to specify the number of bytes to read.

There's also differences in e.g. `aiohttp.ClientResponse` exposing a sync `close()` method, which `httpx.Response` doesn't.

It's all kind of a mess at the moment, and if we want the aiobotocore API of `[Httpx]StreamingBody` to be library-agnostic between aiohttp & httpx we'll likely need to stop using `wrap.ObjectProxy` and explicitly define the API.
We then have to decide whether to remove functionality, write complicated functions to translate between them, petition httpx to implement more functionality, make it possible to directly access the underlying stream when users *do* want to opt into library-specific functionality, and/or raise helpful error messages when something breaks because the backend changed.

See https://github.com/aio-libs/aiobotocore/pull/1085#discussion_r2076634558 and https://github.com/aio-libs/aiobotocore/pull/1085#discussion_r2076627743 for previous discussion.

thehesiod:
## Summary

Restructure aiobotocore's `StreamingBody` classes to subclass `botocore.response.StreamingBody`, following the repo's established `Aio*` override pattern. Addresses @jakob-keller's review feedback that the original PR diverged from this convention.

### Class hierarchy

- `botocore.response.StreamingBody`
  - `AioStreamingBody` — aiohttp backend, async I/O overrides
    - `AioHttpxStreamingBody` — httpx backend, inherits all iteration/validation logic, only overrides the I/O layer (buffered `read(amt)`, `readinto`, async `close`)

`AioStreamingChecksumBody` and `AioHttpxStreamingChecksumBody` mix in `_ChecksumMixin` over the corresponding base.

### What this delivers

- **Drops `wrapt.ObjectProxy`** so the API no longer leaks backend internals; users access the raw stream via `.raw_stream`.
- **Adds full API parity for httpx**: `read(amt)`, `readinto()`, `readlines()`, `__aiter__`/`__anext__`, `iter_lines()`, `iter_chunks()`, `tell()`, `readable()`, `close()`, content-length validation. Previously `read(amt)` raised `ValueError` and most other methods didn't exist.
- **Fixes `HttpxStreamingChecksumBody.readinto`** which called `content.read()` on an `httpx.Response` (would have failed at runtime).
- **Eliminates duplication** by having `AioHttpxStreamingBody` inherit `__aiter__`, `__anext__`, `iter_chunks`, `iter_lines`, `readlines`, `tell`, `_verify_content_length` from `AioStreamingBody`. Only the buffer machinery and httpx I/O layer live in the subclass.
- **Blocks the inherited sync API** (`__iter__`, `__next__`, `__enter__`, `__exit__`, `set_socket_timeout`) with explicit errors so users don't accidentally get coroutines from `for chunk in body` or call urllib3-shaped APIs on an aiohttp/httpx stream.
- **Pre-Aio names preserved as module aliases**: `StreamingBody = AioStreamingBody`, `HttpxStreamingBody = AioHttpxStreamingBody`, same for the checksum classes. External callers and the existing `isinstance` checks throughout the codebase continue to work unchanged.

### Behavior change

`AioStreamingBody.__aenter__` now returns `self` (previously returned the raw aiohttp `ClientResponse`). This matches the httpx side and is the standard pattern. Use `body.raw_stream` for the underlying response object. Noted in CHANGES.rst.

## Test plan

- [x] 22 existing + 2 new aiohttp `StreamingBody` unit tests pass
- [x] 19 httpx `StreamingBody` unit tests pass
- [x] Checksum body tests (`test_httpchecksum.py`, `test_response.py` in `botocore_tests/`) pass
- [x] Patch hash tests (`test_patches.py`) pass — botocore class hash unchanged
- [x] Full unit test suite minus 3 pre-existing failures (`test_configprovider::test_defaults_mode`, `test_eventstreams::test_kinesis_stream_json_parser`) — unrelated to this PR
- [ ] CI moto-based integration tests

Closes #1365


jakob-keller:
I'm all for dropping `wrapt.ObjectProxy`, but beyond that this still confuses me a lot. We discussed potential designs in #1365. Did we ever reach consensus? Why do we deviate from our usual pattern of subclassing upstream with asynchronous methods where required?

thehesiod:
@jakob-keller great point, let me point the agent at the context, this was a simple v1 agent draft

thehesiod:
@claude can you review the pr cmts and related issue to see if there's a better design that follows our repo design patterns

thehesiod:
I'll have my local claude do the refactor to save some money once i get some sleep ;)

thehesiod:
@jakob-keller restructured the design to follow the repo's `Aio*` subclass pattern (35e68c3). The hierarchy now:

```
botocore.response.StreamingBody
  └── AioStreamingBody                   (aiohttp)
        └── AioHttpxStreamingBody        (httpx, inherits all iter/tell/validate logic)
```

`AioHttpxStreamingBody` now inherits `__aiter__`/`__anext__`/`iter_chunks`/`iter_lines`/`readlines`/`tell`/`_verify_content_length` from `AioStreamingBody` — only the buffered httpx I/O layer (`read`, `readinto`, `close`) and the `_buffer`/`_stream_iter` machinery live in the subclass. Net result: ~120 lines deleted, ~89 added.

Old class names (`StreamingBody`, `HttpxStreamingBody`, both checksum variants) remain as module-level aliases so `isinstance` checks and external imports keep working.

One deliberate behavior change worth flagging: `AioStreamingBody.__aenter__` now returns `self` instead of the raw aiohttp `ClientResponse` — matches the httpx side and the surrounding `Aio*` convention. Noted in CHANGES.rst. Use `body.raw_stream` if you need the underlying response.

Re jakkdl's earlier objection in #1365 about subclassing being low-value: the win turned out to be sharing the iteration/validation layer across both backends (which we couldn't do without a common base), not method-signature reuse. Most methods do go sync→async, but that just means the base provides a stable type identity (`isinstance(body, botocore.StreamingBody)` works) while async overrides shadow each sync method. Sync-API entry points (`__iter__`, `__enter__`, `set_socket_timeout`) are now explicit `TypeError`/`NotImplementedError` rather than being silently inherited.

jakob-keller:
@claude: Resolve the conflicts

thehesiod:
ok, due to claude action bump, (at)claude was broken, merging from main

thehesiod:
ok @jakob-keller i think its ready for re-review

aiobotocore/response.py
```diff
@@ -130,53 +111,144 @@ async def iter_lines(self, chunk_size=_DEFAULT_CHUNK_SIZE, keepends=False):
...
+        """
+        self._ensure_stream()
+
+        if amt is None:
+            chunks = [self._buffer]
+            self._buffer = b''
+            try:
+                async for chunk in self._stream_iter:
+                    chunks.append(chunk)
+            except StopAsyncIteration:
```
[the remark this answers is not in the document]
thehesiod: Fixed in 5422d75 — removed the `try/except StopAsyncIteration: pass` entirely. `async for` already handles `StopAsyncIteration` from the iterator's `__anext__` internally, so the catch was dead code. (The parallel manual-`__anext__` loop in `_fill_buffer` line 185-190 still needs and keeps its catch, since manual `__anext__` calls do propagate.)

thehesiod:
@jakob-keller ready for re-review:

- Merged `main` (which now includes #1462's loader warmup at 3.7.0). Bumped this PR to 3.8.0 since 3.7.0 is taken.
- Removed dead `try/except StopAsyncIteration: pass` in `AioStreamingBody.read()` — `async for` handles it internally, the bot's flag was correct.
- Pre-commit clean. `test_response.py`, `test_httpchecksum.py`, `test_response.py` (botocore_tests/unit), `test_patches.py` all pass locally (320 passed, 320 skipped — httpx skips on Windows).

The Aio* hierarchy from 35e68c3 is unchanged: `AioStreamingBody(botocore.response.StreamingBody)` → `AioHttpxStreamingBody(AioStreamingBody)`, with `_ChecksumMixin` for the checksum variants. Old names retained as module aliases.

aiobotocore/__init__.py
jakob-keller, line 1: Please undo: version bumps are now performed at the time of release.

CHANGES.rst
jakob-keller, line 1: Please undo: `CHANGES.rst` is now updated at the time of release.

aiobotocore/response.py
```diff
@@ -16,93 +15,79 @@ class AioReadTimeoutError(ReadTimeoutError, asyncio.TimeoutError):
     pass
 
 
-class StreamingBody(wrapt.ObjectProxy):
-    """Wrapper class for an http response body.
+class AioStreamingBody(botocore.response.StreamingBody):
```
jakob-keller: This class implements an asynchronous context manager. We should make that explicit by adding [contextlib.AbstractAsyncContextManager](https://docs.python.org/3/library/contextlib.html#contextlib.AbstractAsyncContextManager) as a parent class.

aiobotocore/response.py
```diff
@@ -130,53 +111,141 @@ async def iter_lines(self, chunk_size=_DEFAULT_CHUNK_SIZE, keepends=False):
...
+    def __enter__(self):
+        raise TypeError(f"{type(self).__name__} is async; use 'async with'")
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        raise TypeError(f"{type(self).__name__} is async; use 'async with'")
+
+    def set_socket_timeout(self, timeout):
+        raise NotImplementedError(
+            "set_socket_timeout is urllib3-specific; configure timeouts on "
+            "the aiobotocore Config or HTTP session instead."
```
jakob-keller: The message of the error raised by `set_socket_timeout()` is aiohttp-specific, but inherited by `AioHttpxStreamingBody`.

aiobotocore/response.py
jakob-keller: Ideally, we would have an abstract base class `AioStreamingBodyBase(botocore.response.StreamingBody)` that includes the generic patches. Then we could provide concrete subclasses `AioStreamingBody(AioStreamingBodyBase)` with aiohttp-specific tweaks and `HttpxStreamingBody(AioStreamingBodyBase)` with https-specifics.

jakob-keller (changes requested):

aiobotocore/__init__.py
thehesiod: Done in 85ecd02 — reverted `aiobotocore/__init__.py` to match main (back to 3.7.0).

CHANGES.rst
thehesiod: Done in 85ecd02 — reverted `CHANGES.rst` to match main; the 3.8.0 entry is removed.

aiobotocore/response.py
thehesiod: Done in e21d6e6 — `contextlib.AbstractAsyncContextManager` is now an explicit parent of the new shared base `AioStreamingBodyBase`, so both backends inherit it. Verified metaclass-compatible (`io.IOBase`, botocore `StreamingBody`, and `AbstractAsyncContextManager` are all `ABCMeta`).

aiobotocore/response.py
thehesiod: Done in e21d6e6 — moved `set_socket_timeout` to the shared base and reworded the message to be backend-agnostic: "set_socket_timeout is not supported for async streaming bodies; configure timeouts on the aiobotocore Config or HTTP session instead."

aiobotocore/response.py
thehesiod: Done in e21d6e6 — restructured as suggested:

- `AioStreamingBodyBase(botocore.response.StreamingBody, contextlib.AbstractAsyncContextManager)` holds the generic patches (iteration, line splitting, content-length validation, async context manager, blocked sync surface, `set_socket_timeout`).
- `AioStreamingBody(AioStreamingBodyBase)` — aiohttp-specific I/O.
- `AioHttpxStreamingBody(AioStreamingBodyBase)` — httpx-specific I/O (no longer subclasses the aiohttp class).

One decision worth flagging: I kept the back-compat alias `StreamingBody = AioStreamingBody` (the aiohttp concrete class) rather than pointing it at the base, because the existing tests both construct `response.StreamingBody(...)` expecting an aiohttp body and use `isinstance(body, StreamingBody)` to distinguish the aiohttp backend (sync `close()`) from httpx (async `aclose()`). A nice side effect: this fixes a latent bug — previously httpx bodies were instances of `StreamingBody` and hit the sync `close()` branch, calling the `async def close` without awaiting. The base is importable as `AioStreamingBodyBase` for "is this any streaming body" checks.

I didn't mark the base's I/O methods `@abstractmethod` to keep the diff minimal — happy to make it a strict ABC if you'd prefer.

CHANGES.rst
jakob-keller: Looks like it has removed and/or rephrased previous entries?

amohr-northstar: Good catch — that one's on me. I reverted the file using main's *current* content, but the PR diff is computed against the merge base and this branch was ~10 commits behind main. So it looked like the PR was removing/rephrasing the older `3.7.0 (2026-05-08)` / `3.6.1` entries, when it was really just pulling in main's consolidated `3.7.0 (2026-05-09)` section.

Fixed in 2e6fec3: `CHANGES.rst` is restored to its merge-base content, so this PR doesn't diff it at all. I've also merged main in (4810184), so the file now tracks main via the merge rather than via this PR.

`git diff origin/main...HEAD -- CHANGES.rst aiobotocore/__init__.py` is now empty — the PR only touches `CLAUDE.md`, `endpoint.py`, `httpchecksum.py`, `response.py`, and `tests/test_response.py`.

thehesiod:
going to merge after ci/cd since it just needed a changes.rst revert and that file is clean here

aiobotocore/response.py
```diff
@@ -16,167 +16,250 @@ class AioReadTimeoutError(ReadTimeoutError, asyncio.TimeoutError):
...
+        super().__init__(raw_stream, content_length)
+        self._buffer = b''
+        self._stream_iter = None
+        self._stream_exhausted = False
 
-        This is achieved by reading chunk of bytes (of size chunk_size) at a
-        time from the raw stream, and then yielding lines from there.
+    def _ensure_stream(self):
+        if self._stream_iter is None:
+            self._stream_iter = self._raw_stream.aiter_bytes().__aiter__()
```
thehesiod: [P1] Iterate raw bytes before validating Content-Length

httpx.Response.aiter_bytes() yields decoded content (gzip/deflate/etc.), while Content-Length and AWS checksum headers describe the wire/object bytes. This branch now feeds decoded bytes into botocore's content-length/checksum validation. A gzip response with 41 wire bytes and 1100 decoded bytes raises IncompleteReadError (1100 read, expected 41) on await body.read(); the aiohttp backend explicitly uses auto_decompress=False. Please use aiter_raw() here, or otherwise validate raw byte counts, so content-encoded streaming outputs such as S3 objects remain readable.

aiobotocore/response.py
```diff
@@ -16,167 +16,250 @@ class AioReadTimeoutError(ReadTimeoutError, asyncio.TimeoutError):
...
+        if amt is None:
+            chunks = [self._buffer]
+            self._buffer = b''
+            async for chunk in self._stream_iter:
+                chunks.append(chunk)
+            self._stream_exhausted = True
+            result = b''.join(chunks)
+        elif amt == 0:
+            return b''
+        else:
```
thehesiod, line 221: [P2] Treat negative sizes as read-all

The aiohttp implementation passes any negative amt through to StreamReader.read(), and Python/botocore file-like callers routinely use read(-1) for read-all. Here -1 reaches the else branch: _fill_buffer(-1) does nothing, self._buffer[:-1] is empty initially, and the stream is not advanced. A direct six-byte reproduction returns b'' with tell() still 0. Please normalize amt < 0 to None (or route it through the read-all path) and add HTTPX coverage.

thehesiod (commented):
Found two HTTPX streaming regressions inline: decoded response bytes are validated against raw Content-Length/checksums, and read(-1) silently returns no data. Focused response/checksum/hash tests otherwise pass on both backend selections. GitHub does not allow requesting changes on my own PR, so recording this as a comment review.

aiobotocore/response.py
thehesiod: Confirmed and fixed in fe73d3f. Reproduced exactly as described — a gzip body with 44 wire bytes and 1300 decoded bytes:

```
wire(compressed)=44  decoded=1300  Content-Length header=44
IncompleteReadError: 1300 read, but total bytes expected is 44.
```

Switched `_ensure_stream()` to `aiter_raw()`, which matches the aiohttp backend's `auto_decompress=False` (httpsession.py:197) so both backends now yield wire bytes and validate against wire counts. This also covers the checksum path, since `_ChecksumMixin` would otherwise have hashed decoded bytes.

Worth noting this was a regression introduced by this PR: the old `HttpxStreamingBody` was a `wrapt.ObjectProxy` whose `read()` was just `await self.__wrapped__.aread()`, with no content-length validation at all — inheriting `_verify_content_length` is what exposed it.

Added `test_httpx_content_encoded_body_validates_wire_bytes`, which fails with the original `IncompleteReadError` without the fix. It builds the response with `stream=httpx.ByteStream(...)` rather than `content=`, since `content=` marks the stream consumed and `aiter_raw()` then raises `StreamConsumed` (not a production concern — `httpxsession.py` sends with `stream=True` and only builds a streaming body while the stream is unread).

aiobotocore/response.py
thehesiod: Confirmed and fixed in fe73d3f. Reproduced exactly as described — `read(-1)` on a six-byte body returned `b''` with `tell()` still 0:

```
httpx read(-1) -> b''  tell=0     # expected b'abcdef'
```

`_fill_buffer(-1)` no-ops because `0 < -1` is false, and `buffer[:-1]` is empty, so the stream never advanced. Routed `amt < 0` to the read-all path so it matches aiohttp, which passes negative sizes straight to `StreamReader.read()`.

I kept the content-length validation condition as `amt is None or (not result and amt > 0)`, so `read(-1)` reads all *without* triggering validation — same as the aiohttp backend and botocore's `StreamingBody.read`, rather than inventing a third behavior.

Added `read(-1)` coverage for both backends (`test_httpx_read_negative_reads_all`, `test_streaming_body_read_negative_reads_all`); the httpx one fails without the fix.

PR #1534, opened 2026-04-06
Bump `botocore` dependency specification

jakob-keller:
### Description of Change
This PR intends to improve general compatibility of `aiobotocore` within the Python ecosystem by bumping the dependency specification of `botocore`.

### Assumptions
Upstream contains several changes that require adjustments to the aiobotocore codebase. 

### Checklist for All Submissions
* [x] I have added change info to [CHANGES.rst](https://github.com/aio-libs/aiobotocore/blob/master/CHANGES.rst)
* [x] If this is resolving an issue (needed so future developers can determine if change is still necessary and under what conditions) (can be provided via link to issue with these details): closes #1533
  * [ ] Detailed description of issue
  * [ ] Alternative methods considered (if any)
  * [ ] How issue is being resolved
  * [ ] How issue can be reproduced
* [ ] If this is providing a new feature  (can be provided via link to issue with these details):
  * [ ] Detailed description of new feature
  * [ ] Why needed
  * [ ] Alternatives methods considered (if any)

### Checklist when updating botocore and/or aiohttp versions

* [x] I have read and followed [CONTRIBUTING.rst](https://github.com/aio-libs/aiobotocore/blob/master/CONTRIBUTING.rst#how-to-upgrade-botocore)
* [x] I have updated test_patches.py where/if appropriate (also check if no changes necessary)
* [x] I have added URL to diff: https://github.com/boto/botocore/compare/1.42.70...1.42.84

PR #1567, opened 2026-04-19
evals: switch default model to Sonnet + harden parse-error path

thehesiod:
## Summary

Two follow-ups from analysis of the first post-merge eval run ([runs/24625423913](https://github.com/aio-libs/aiobotocore/actions/runs/24625423913)): 8/8 PASS, but with 3/24 individual-run flakes including one parse-error on the 259-line diff.

### 1. Default model: Opus 4.7 → Sonnet 4.6

No principled reason for Opus on the evals when `claude-code-action` in `claude.yml` and `botocore-sync.yml` uses its built-in Sonnet default. Structured classification with a clear system prompt doesn't need Opus-level reasoning.

- ~5× cheaper per call
- Faster responses
- First-run comparison on Opus showed 8/8 on both eval suites; Sonnet expected to match

### 2. Parse-error hardening

Last run's single parse-error (PR #1534, 259-line diff) happened because the model's response didn't contain a regex-matchable `CLASSIFICATION:` line. Two likely causes: markdown-bold wrapping (`**CLASSIFICATION**:`) or verdict placed after explanation prose. Fixing both sides:

- **Regex loosened** in `check_async_need.py` and `check_override_drift.py` to tolerate markdown asterisks and leading whitespace:
  ```python
  VERDICT_RE = re.compile(
      r"^\s*\**\s*CLASSIFICATION\s*\**\s*:\s*(\S+)",
      re.MULTILINE,
  )
  ```
- **User-message format tightened**: replaced the soft "starting with a line `CLASSIFICATION: <verdict>`" instruction with strict two-step output rules:
  > 1. The VERY FIRST line of your response must be `CLASSIFICATION: <verdict>`…
  > 2. Any supporting reasoning goes AFTER the classification line.

Both changes together should eliminate the parse-error class entirely — if the model ignores the format rule, the looser regex still catches the markdown-wrapped fallback.

### 3. Tests updated

`test_invoke_and_parse_verdict_regex_shape` now covers three plausible model emissions: plain, markdown-bold, and leading-whitespace.

## Expected on next run

- Same 8/8 and 2/2 majority-vote pass
- Individual-run parse-error rate → 0
- Job duration drop observable (Sonnet latency < Opus)

## Test plan

- [x] `uv run pytest tests/evals/` — 20/20 pass
- [x] Pre-commit passes
- [ ] Manual `gh workflow run evals.yml --ref amohr/evals-sonnet-and-parse-hardening` to confirm expectations

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
     {
       "name": "aiobotocore-bot",
       "source": "./plugins/aiobotocore-bot",
-      "description": "Slash commands for the aiobotocore CI automation: /aiobotocore-bot:review-pr and /aiobotocore-bot:analyze-pr-feedback."
+      "description": "Skills for the aiobotocore CI automation: review-pr, analyze-pr-feedback, check-async-need, check-override-drift, open-pr, bump-version, pyright-delta, complete-run."
     }
   ]
 }
--- a/.github/botocore-sync-prompt.md
+++ b/.github/botocore-sync-prompt.md
@@ -184,9 +184,9 @@ Run the classifier:
 /aiobotocore-bot:check-async-need --from=$LAST_SUPPORTED --to=$LATEST_BOTOCORE
 ```
 
-The command diffs the two botocore versions, finds new/changed functions in overridden files, and returns one of:
+The skill diffs the two botocore versions, finds new/changed functions in overridden files, and returns one of:
 
-- `no-port` → no async-need signals found. Go to Step 4 (no-port path). Quote the command's summary line in the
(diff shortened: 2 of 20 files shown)


thehesiod:
@jakob-keller FYI what I'm working on with the "evals", tests based on a baseline ensuring historical classification runs properly categorize "bumps" as non-port/minor (bumping end version), to port (requiring code changes and bump of start version).  This helps show that future automatically created PRs most likely will have value

thehesiod:
also some internal housekeeping of helpers (skills) for the ai to do its job

PR #1573, opened 2026-04-19
port-tests: allow creating new mirror files for tests exercising registered overrides

thehesiod:
## Summary

Observed on [#1571](https://github.com/aio-libs/aiobotocore/pull/1571): the port-tests skill
skipped `test_auth_scheme_preference` because no mirror file existed at
`tests/botocore_tests/unit/test_endpoint_provider.py`. But the test exercises
`EndpointRulesetResolver` behavior that `AioEndpointRulesetResolver` inherits — valuable
coverage for aiobotocore's subclass. A human `@claude` override was needed to unblock,
producing commit [`35a5ebd`](https://github.com/aio-libs/aiobotocore/commit/35a5ebdd7630a5b951cf6cf1bd0cd128cd0d80da).

Relax Step 1's filter: a botocore test file is in scope if EITHER it has an existing
aiobotocore mirror OR its new/changed tests reference a symbol in the `overrides` /
`aio_classes` registries. In the second case the skill creates a new mirror file,
bringing over only the test + its direct dependencies (constants, helpers).

Fully out-of-scope cases remain untouched: files with no mirror AND no tracked-override
reference stay skipped — that's a genuine human judgment call, not a mechanical port.

## Minor knock-on changes

- Step 4's revert path: if the mirror file was newly created, `rm <path>` instead of
  `git checkout -- <path>` (no prior content to restore).
- `allowed-tools` frontmatter gains `Bash(rm:*)` for the new revert path.
- Step 5 skipped-file report reason updated to reflect the registry check.
- Backfill mode is unchanged — still fills gaps in existing mirrors only.

## Test plan

- [x] Pre-commit passes
- [ ] Next port-required sync run — confirm the skill creates a new mirror when appropriate
      instead of deferring to a human override

### DIFF
--- a/plugins/aiobotocore-bot/skills/port-tests/SKILL.md
+++ b/plugins/aiobotocore-bot/skills/port-tests/SKILL.md
@@ -1,7 +1,7 @@
 ---
 description: Use when porting botocore tests to aiobotocore (during a port-required sync, or when backfilling historical test coverage). Converts sync test files under `botocore/tests/...` to their async counterparts under `tests/botocore_tests/...`, validates each port with `pytest -x`, commits on pass, reverts on fail. Handles both new/changed tests from a botocore diff AND backfill of relevant-but-not-yet-ported tests.
 argument-hint: "[--from=<version>] [--to=<version>] [--backfill] [--paths=<file1,file2,...>] [--dry-run]"
-allowed-tools: Bash(git -C /tmp/botocore:*) Bash(git checkout:*) Bash(rm:*) Bash(uv run pytest:*) Bash(uv run python:*) Bash(ls:*) Bash(find:*) Bash(cat:*) Bash(diff:*) Bash(cp:*) mcp__github_file_ops__commit_files
+allowed-tools: Bash(git -C /tmp/botocore:*) Bash(git checkout:*) Bash(rm tests/botocore_tests/:*) Bash(uv run pytest:*) Bash(uv run python:*) Bash(ls:*) Bash(find:*) Bash(cat:*) Bash(diff:*) Bash(cp:*) mcp__github_file_ops__commit_files
 ---
 
 Port botocore tests to aiobotocore. Two driving scenarios:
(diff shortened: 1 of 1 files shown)


thehesiod:
While reviewing this in the context of [#1574](https://github.com/aio-libs/aiobotocore/pull/1574) (cost/clarity follow-ups from #1571), I noticed one place where the case-2 logic could be tightened — flagging here so the call is yours whether to fold in or defer:

**Step 1 case 2 (\"new/changed tests reference a symbol in the `overrides` registry\")** is described declaratively but the SKILL doesn't tell the agent *how* to check. Today the agent has to pattern-match symbol names against `scripts/registries.py` contents inline. That's fine for an Opus run; once #1574 lands and Sonnet runs the no-port path (and any port-tests work it dispatches), the implicit grep step is more likely to miss subtle cases (e.g. `mock.patch('botocore.foo.X')` strings vs bare class refs).

A concrete grep example in case 2 would help. Something like:

```bash
# Load tracked overrides
overrides=$(uv run python plugins/aiobotocore-bot/scripts/registries.py | jq -r '.overrides[]')

# For each candidate test file, check whether its new/changed lines reference any tracked override
git -C $BOTOCORE_CLONE diff $FROM..$TO -- $candidate \
  | grep -E '^\+' \
  | grep -Ff <(echo \"\$overrides\")
```

Match on:
- bare function names (e.g. `_apply_request_trailer_checksum`)
- dotted method names (e.g. `ClientArgsCreator.get_client_args`) — both literal references AND `mock.patch('botocore.args.ClientArgsCreator.get_client_args')` strings
- bare class names — either as constructor calls (`ClientCreator(...)`) or in patch strings

Edge case worth calling out: `aio_classes` matches need to consider that a test referencing the bare botocore class (e.g. `EndpointRulesetResolver(...)`) is exercising what `AioEndpointRulesetResolver` inherits — that's exactly the case [#1571 commit `35a5ebd`](https://github.com/aio-libs/aiobotocore/commit/35a5ebdd7630a5b951cf6cf1bd0cd128cd0d80da) fixed manually and what this PR is generalizing.

Otherwise the diff looks solid — `Bash(rm tests/botocore_tests/:*)` is correctly scoped (Claude Code's permission DSL prefix-matches), the 3-tier scope rule is internally consistent with Step 2's `overrides` filter, and the dual revert path (`git checkout --` for existing mirrors / `rm` for new ones) handles both cases cleanly. No blocking findings.

PR #1579, opened 2026-04-30
Relax `botocore` dependency specification

jakob-keller:
### Description of Change

This PR intends to improve general compatibility of `aiobotocore` within the Python ecosystem by relaxing the dependency specification of `botocore`.

Following the release of botocore 1.43.0, this PR removes support for end-of-life Python 3.9.

### Assumptions

Upstream contains no changes that require adjustments to the aiobotocore codebase.

### Checklist for All Submissions

* [x] I have added change info to [CHANGES.rst](https://github.com/aio-libs/aiobotocore/blob/main/CHANGES.rst)
* [ ] If this is resolving an issue (needed so future developers can
  determine if change is still necessary and under what conditions)
  (can be provided via link to issue with these details):
  * [ ] Detailed description of issue
  * [ ] Alternative methods considered (if any)
  * [ ] How issue is being resolved
  * [ ] How issue can be reproduced
* [ ] If this is providing a new feature (can be provided via link to issue with these details):
  * [ ] Detailed description of new feature
  * [ ] Why needed
  * [ ] Alternatives methods considered (if any)

### Checklist when updating botocore and/or aiohttp versions

* [x] I have read and followed [CONTRIBUTING.rst](https://github.com/aio-libs/aiobotocore/blob/main/CONTRIBUTING.rst#how-to-upgrade-botocore)
* [x] I have updated test_patches.py where/if appropriate (also check if no changes necessary)
* [x] I have added URL to diff: https://github.com/boto/botocore/compare/1.42.97...1.43.0

PR #1582, opened 2026-05-01
ci(claude): skip dependabot and fork PRs in pull_request trigger

thehesiod:
## Summary

Several independent fixes to the Claude Code workflow, surfaced from the same investigation of recent failures.

**1. Switch `pull_request` → `pull_request_target`.** `claude-code-action`'s anti-tampering validator hard-fails any run where the calling workflow file diverges from the default branch — i.e. every PR that touches `claude.yml` itself (dependabot bumps, maintainer tweaks like this very PR). Recent example: [#1580 run](https://github.com/aio-libs/aiobotocore/actions/runs/25196525529). `pull_request_target` runs in the base repo's context, so the workflow file used IS main's and the validator passes.

Forks are still excluded in the `if:`, which is critical: with `pull_request_target` the job runs with full base secrets (`ANTHROPIC_API_KEY`, write `GITHUB_TOKEN`, etc.), and `uv sync` / the prompt template / plugin marketplace / `usage-summary.py` are all loaded from the PR's checkout. Letting fork code through would be the canonical pull_request_target privilege escalation antipattern. Maintainer-triggered review on fork PRs via `@claude` in `issue_comment` continues to work — it runs from main's workflow with no fork code execution.

`actions/checkout` pins to `github.event.pull_request.head.sha` (not the ref name) so a force-push mid-job can't swap in different code than the `if:` gate evaluated against.

**2. Skip dependabot PRs.** `claude-code-action`'s `allowed_bots: "claude[bot]"` gate hard-fails any run initiated by `dependabot[bot]` with `Workflow initiated by non-human actor: dependabot (type: Bot)`. Recent example: [#1581 run](https://github.com/aio-libs/aiobotocore/actions/runs/25196625746). Nothing useful to review on a dep bump anyway.

**3. Skip fork PRs in auto-review (with explanatory notice).** Auto-review on fork PRs has been broken (OIDC token unavailable) and remains undesirable for the security reasons above. Adds a tiny `notify-fork-pr` job that fires on `pull_request_target: [opened]` for fork PRs only, posting a single comment so external contributors know to ask a maintainer to trigger review with `@claude`. Safe `pull_request_target` use: no checkout, no fork code executed, hardcoded body, only the PR number is read from the event payload. Gated to `opened` to avoid re-commenting on every push.

**4. Stop posting duplicate top-level comments on `@claude` mentions.** `claude.yml` enables `use_sticky_comment: true` + `track_progress: true`, so `claude-code-action` already posts the agent's final assistant message as a top-level PR comment. The prompt's Cleanup section then ran `complete-run --summary=…` *on top*, producing two top-level comments per run for `issue_comment` and `pull_request_review` events. Concrete example on PR #1579: comment [4358756424](https://github.com/aio-libs/aiobotocore/pull/1579#issuecomment-4358756424) from the sticky, comment [4358767006](https://github.com/aio-libs/aiobotocore/pull/1579#issuecomment-4358767006) from `complete-run`, near-identical content.

Updated `.github/claude-review-prompt.md` Cleanup section: pass `--skip-reply` on `issue_comment` and `pull_request_review`, since the sticky comment is the canonical reply on those paths. `pull_request_review_comment` (inline-thread reply) and `issues` (no PR sticky) keep posting because they're on different surfaces.

## Test plan

- [ ] Land this; confirm next dependabot PR shows the claude job as skipped, not failed.
- [ ] Confirm next fork PR shows the claude job as skipped and a comment from the `notify-fork-pr` job appears once.
- [ ] Confirm `@claude please review` from a maintainer on a fork PR still triggers the workflow via `issue_comment` and produces a single top-level review comment (not two).
- [ ] Confirm same-repo non-bot PRs (including ones that modify `claude.yml`) still auto-review.

## Caveat

`pull_request_target` + `claude-code-action`'s `prompt:` parameter is not the documented happy path. Earlier research surfaced [issue #621](https://github.com/anthropics/claude-code-action/issues/621) about review comments not posting under `pull_request_target` in some configurations. If we hit that, revert the trigger change and accept the validation failure on workflow-touching PRs (workaround: maintainer comments `@claude` to trigger review from main's workflow via `issue_comment`).

### DIFF
--- a/.github/claude-review-prompt.md
+++ b/.github/claude-review-prompt.md
@@ -195,19 +195,33 @@ IMPORTANT: Never merge or close pull requests. Never close issues. These actions
 End-of-run cleanup via `/aiobotocore-bot:complete-run`. The skill posts the summary reply to the right target
 (inline thread vs top-level PR comment vs issue comment, based on `--event`) and swaps the 👀 reaction for 👍.
 
+`claude.yml` enables `use_sticky_comment: true` + `track_progress: true`, so claude-code-action automatically
+posts a top-level PR comment containing your final assistant message. For events whose natural reply target
+*is* a top-level PR comment, that sticky comment already serves as the reply — passing `--skip-reply` to
+`complete-run` avoids duplicate top-level comments. For inline-thread replies (`pull_request_review_comment`)
+and issue threads (`issues`), the sticky comment is on a different surface, so `complete-run` still needs to
+post.
+
 Invocations by event:
 
-- `pull_request_review_comment` or `issue_comment`:
-  `/aiobotocore-bot:complete-run --event=$EVENT_NAME --number=$NUMBER --comment-id=$COMMENT_ID --summary="<body>"`
-- `pull_request_review`:
-  `/aiobotocore-bot:complete-run --event=pull_request_review --number=$NUMBER --summary="<body>"`
-- `issues`:
-  `/aiobotocore-bot:complete-run --event=issues --number=$NUMBER --summary="<body>"`
+- `pull_request_review_comment`:
+  `/aiobotocore-bot:complete-run --event=pull_request_review_comment --number=$NUMBER --comment-id=$COMMENT_ID --summary="<body>"`
+  — inline-thread reply, distinct surface from the sticky comment.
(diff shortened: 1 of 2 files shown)


thehesiod:
@claude please review this pr

PR #1588, opened 2026-05-09
fix: credential_process on Windows + Selector event loop (closes #1415)

### LINKED ISSUE #1415: Issue with external Credential Process on Windows using WindowsSelectorEventLoopPolicy.
Because psycopg doesn't support Proactor event loops, on Windows it's common to use Selector event loops for asyncio.  However, Selector event loops don't support async subprocesses.  The result of this is that this minimal code fails on Windows, if you have an `.aws/config` with a `credential_process` set up.

```
import asyncio, aioboto3

asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())

async def test_upload():
    session = aioboto3.Session(region_name='us-east-1')
    async with session.client('s3') as client:
        await client.download_fileobj("my-bucket", "my-key", open("downloaded_test.py", "wb"))

asyncio.run(test_upload())
```

with the below stack.  One fix is simply to detect this condition and not bother using async Popen.  Calling sync Popen instead of `self._popen` works fine (like this:

        p = subprocess.Popen(process_list, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
        stdout, stderr = p.communicate()

```
  File "...\.venv\Lib\site-packages\aiobotocore\session.py", line 197, in _create_client
    credentials = await self.get_credentials()
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\.venv\Lib\site-packages\aiobotocore\session.py", line 100, in get_credentials
    self._credentials = await self._components.get_component(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\.venv\Lib\site-packages\aiobotocore\credentials.py", line 1024, in load_credentials
    creds = await provider.load()
            ^^^^^^^^^^^^^^^^^^^^^
  File "...\.venv\Lib\site-packages\aiobotocore\credentials.py", line 538, in load
    creds_dict = await self._retrieve_credentials_using(credential_process)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\.venv\Lib\site-packages\aiobotocore\credentials.py", line 558, in _retrieve_credentials_using
    p = await self._popen(
        ^^^^^^^^^^^^^^^^^^
  File "...\Python311\Lib\asyncio\subprocess.py", line 223, in create_subprocess_exec
    transport, protocol = await loop.subprocess_exec(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\Python311\Lib\asyncio\base_events.py", line 1708, in subprocess_exec
    transport = await self._make_subprocess_transport(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\Python311\Lib\asyncio\base_events.py", line 503, in _make_subprocess_transport
    raise NotImplementedError
NotImplementedError
```

thehesiod:
## Summary

Apps that mix aiobotocore with libraries that require ``asyncio.WindowsSelectorEventLoopPolicy`` (e.g. ``psycopg``, which doesn't support the Proactor loop) currently fail at credential resolution time if ``~/.aws/config`` has a ``credential_process`` set:

```
File ".../aiobotocore/credentials.py", line 586, in _retrieve_credentials_using
  p = await self._popen(...)
File ".../asyncio/base_events.py", line 1798, in subprocess_exec
  transport = await self._make_subprocess_transport(...)
NotImplementedError
```

The Selector loop on Windows doesn't implement subprocess transports.

Closes #1415.

## Reproduction (verified locally)

`~/.aws/config`:
```
[default]
credential_process = python /path/to/fake_cred_process.py
```

```python
import asyncio, aiobotocore.session
asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())

async def main():
    session = aiobotocore.session.AioSession()
    creds = await session.get_credentials()  # raised NotImplementedError before this PR
    frozen = await creds.get_frozen_credentials()
    print(frozen.access_key)

asyncio.run(main())
```

Before: ``NotImplementedError`` at ``loop.subprocess_exec``. After: prints the access key.

## Fix

Catch ``NotImplementedError`` around the existing ``await self._popen(...)`` call and fall back to ``subprocess.run`` dispatched via ``asyncio.to_thread``. The event loop stays unblocked while the credential process executes; the rest of the flow (JSON parsing, refreshable-credentials wiring) is unchanged.

The fallback is duck-typed on ``NotImplementedError`` rather than ``sys.platform == 'win32'`` — any custom or constrained loop that lacks subprocess transports gets the same behavior.

## Test plan

- [x] New regression test ``test_processprovider_falls_back_to_sync_when_subprocess_unsupported`` in ``tests/botocore_tests/unit/test_credentials.py`` mocks ``_popen`` to raise ``NotImplementedError``, mocks ``subprocess.run`` for the fallback, and verifies credentials are returned end-to-end.
- [x] Existing 9 ``processprovider`` tests continue to pass.
- [x] Pre-commit clean.
- [x] Manual repro on Windows: ``WindowsSelectorEventLoopPolicy`` + ``credential_process`` succeeds with the fix; fails on ``main`` without it.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,13 @@
 Changes
 -------
 
+3.7.1 (2026-05-08)
+^^^^^^^^^^^^^^^^^^
+* fall back to synchronous ``subprocess.run`` (via ``asyncio.to_thread``) for
+  ``credential_process`` when the running event loop does not implement
+  subprocess transports — notably ``asyncio.SelectorEventLoop`` on Windows
+  (closes #1415)
+
 3.7.0 (2026-05-08)
 ^^^^^^^^^^^^^^^^^^
 * add ``AioSession.warm_up_loader_caches()`` and ``warm_up_loader_caches`` option in ``AioConfig`` to pre-populate
--- a/aiobotocore/__init__.py
+++ b/aiobotocore/__init__.py
@@ -1,3 +1,3 @@
 """Async client for AWS services, wrapping botocore with aiohttp/httpx."""
 
-__version__ = '3.7.0'
+__version__ = '3.7.1'
--- a/aiobotocore/credentials.py
+++ b/aiobotocore/credentials.py
@@ -552,6 +552,13 @@ class AioAssumeRoleWithWebIdentityCredentialFetcher(
         return assume_role_kwargs
 
 
+def _run_credential_process_sync(process_list):
+    # Synchronous fallback for event loops that don't implement
+    # subprocess transports. Caller runs this via ``asyncio.to_thread``.
+    p = subprocess.run(process_list, capture_output=True, check=False)
+    return p.stdout, p.stderr, p.returncode
+
+
 class AioProcessProvider(ProcessProvider):
     def __init__(self, *args, popen=asyncio.create_subprocess_exec, **kwargs):
         super().__init__(*args, **kwargs, popen=popen)
@@ -583,11 +590,23 @@ class AioProcessProvider(ProcessProvider):
         # We're not using shell=True, so we need to pass the
         # command and all arguments as a list.
         process_list = compat_shell_split(credential_process)
-        p = await self._popen(
-            *process_list, stdout=subprocess.PIPE, stderr=subprocess.PIPE
-        )
-        stdout, stderr = await p.communicate()
-        if p.returncode != 0:
+        try:
+            p = await self._popen(
+                *process_list, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+            )
+            stdout, stderr = await p.communicate()
+            returncode = p.returncode
+        except NotImplementedError:
+            # Some event loops don't implement subprocess transports —
+            # notably ``asyncio.SelectorEventLoop`` on Windows, which
+            # users select when integrating with libraries that don't
+            # support the Proactor loop (e.g. ``psycopg``). Fall back
+            # to running the credential process synchronously in a
+            # worker thread so the loop stays unblocked. (#1415)
+            stdout, stderr, returncode = await asyncio.to_thread(
+                _run_credential_process_sync, process_list
+            )
+        if returncode != 0:
             raise CredentialRetrievalError(
                 provider=self.METHOD, error_msg=stderr.decode('utf-8')
             )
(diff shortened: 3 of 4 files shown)


aiobotocore/__init__.py
```diff
@@ -1,3 +1,3 @@
 """Async client for AWS services, wrapping botocore with aiohttp/httpx."""
 
-__version__ = '3.7.0'
```
jakob-keller, line 3: This version bump is unnecessary: The released version is still at 3.6.0.

CHANGES.rst
```diff
@@ -1,6 +1,13 @@
 Changes
 -------
 
+3.7.1 (2026-05-08)
```
jakob-keller, line 4: No need for version 3.7.1: The changes can go into unreleased 3.7.0.

thehesiod: thanks, nice catch, I'll merge them on main.  Working on cleaning up our backlog

jakob-keller: Yeah, thanks! I'm a bit busy in real life, but will try to review tomorrow.

aiobotocore/__init__.py
thehesiod: Consolidated into a single 3.7.0 entry in #1593 — folds 3.6.1 + 3.7.0 + 3.7.1 (all unreleased) into one bullet list. Also tightened the draft-release skill in #1592 to detect this state going forward (Step 4a: read latest PyPI version, treat anything past it as the in-progress draft, fold new bullets into it instead of stacking).

CHANGES.rst
thehesiod: Done in #1593 — 3.7.1 + 3.7.0 + 3.6.1 all rolled into one 3.7.0 entry. The minor bump is justified by the warm-up-loader-caches feature; the other two bullets are bug fixes that came in afterwards. Bumped #1539 to 3.8.0 in the meantime.

PR #1589, opened 2026-05-09
ci: smoke-test built wheel and sdist after build (closes #1169)

### LINKED ISSUE #1169: Test distribution packages in CI

thehesiod:
## Summary

Adds a ``test-dist`` matrix job (wheel, sdist) that depends on ``build`` and verifies the produced artifact installs into a fresh Python env and imports cleanly. The job is wired into the ``check`` aggregator so branch protection gates on it.

Closes #1169.

## What this catches that current CI doesn't

| Existing | What it catches |
|-|-|
| ``twine check --strict`` (in ``build``) | metadata validity, README rendering |
| Test matrix (``test``) | functional behavior, but installs from source via ``uv sync`` so an editable install — never actually exercises the built wheel/sdist |

This new job catches the gap:

- **MANIFEST issues** — required modules missing from the sdist
- **pyproject.toml dependency mistakes** — package installs from source via uv but the wheel can't resolve its declared dependencies on a clean env
- **Missing data files** — anything referenced at import time but not packaged
- **Conditional imports** — ``aiobotocore.httpxsession`` requires the ``[httpx]`` extra; smoke test uses it to verify the extra resolves

## What this is NOT

@webknjaz's original framing in #1169 was broader: *"the tests should depend on building the dists"* — i.e., make the entire test matrix install from the dist artifact instead of editable source. That's a meaningful CI restructure (changes coverage paths, requires test files outside the source tree, breaks editable-install dev ergonomics). This PR is the smoke-test-sized increment; happy to file a follow-up for the full restructure if there's interest.

## Implementation notes

- Both wheel and sdist as a matrix (fail-fast: false) so each is exercised independently.
- Uses ``uv venv`` + ``uv pip install`` for env creation rather than ``actions/setup-python`` to match the rest of the repo's tooling (every other job uses ``astral-sh/setup-uv``).
- Pinned to the same SHA-and-comment-version pattern as other actions in this workflow.
- 3-minute timeout (build-and-import is fast).
- Imports the public submodules explicitly (``aiobotocore.session``, ``client``, ``config``, ``endpoint``, ``httpsession``, ``httpxsession``, ``credentials``, ``response``) — so an accidentally missing module surfaces in the smoke test rather than at user import time.

## Test plan

- [x] Pre-commit clean locally (yamllint, GitHub Actions schema validation, line-length).
- [x] Local validation: built wheel + sdist via ``uv build``, installed wheel into fresh ``uv venv``, all listed imports succeed.
- [ ] CI run on this PR will exercise the new job against a real GitHub-built dist artifact.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -93,11 +93,55 @@ jobs:
 
     uses: zizmorcore/workflow/.github/workflows/reusable-zizmor.yml@3bb5e95068d0f44b6d2f3f7e91379bed1d2f96a8
 
+  test-dist:
+    name: Smoke-test ${{ matrix.dist-kind }} dist
+    needs:
+    - build
+    strategy:
+      matrix:
+        dist-kind: [wheel, sdist]
+      fail-fast: false
+    runs-on: ubuntu-24.04
+    timeout-minutes: 3
+    steps:
+    - name: Download distribution 📦
+      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c  # v8.0.1
+      with:
+        name: python-package-distributions
+        path: dist/
+    - name: Install uv
+      uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b  # v8.1.0
+      with:
+        version: '0.11.7'
+        python-version: '3.14'
+    # Install the built artifact into a fresh Python with no project source
+    # checked out — verifies the dist is self-contained (MANIFEST is correct,
+    # all required modules ship, dependencies resolve, ``import aiobotocore``
+    # works on a clean env). twine check --strict (in build) covers metadata
+    # validity; this complements it with a real install + import test.
+    - name: Install ${{ matrix.dist-kind }} and verify import
+      env:
+        DIST_KIND: ${{ matrix.dist-kind }}
+      run: |
+        case "$DIST_KIND" in
+          wheel) artifact=(dist/*.whl) ;;
+          sdist) artifact=(dist/*.tar.gz) ;;
+        esac
+        uv venv smoke-env
+        uv pip install --python smoke-env "${artifact[0]}[httpx]"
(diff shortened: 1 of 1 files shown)


thehesiod (commented):
Re-analysed this against #1169 and found the smoke test can't detect the packaging bug that's
live on `main` right now, so I've opened #1686 with the fuller fix and would suggest closing
this one.

The gap: `main`'s sdist ships 24 test modules and omits 29, including `tests/__init__.py`,
`tests/conftest.py`, `tests/mock_server.py` and all of `tests/botocore_tests/`. `pytest` on an
unpacked sdist stops with five collection errors:

```
tests/test_stubber.py:5: from .mock_server import AIOServer
E   ImportError: attempted relative import with no known parent package
!!!! Interrupted: 5 errors during collection !!!!
```

The `test-dist` job here passes green on exactly that sdist, because `import aiobotocore`
succeeds regardless of whether the test suite shipped. Which is fair — it's what the job was
scoped to check — but it means merging it would read as "#1169 is handled" while downstream
packagers still can't run the suite they're shipping.

There was also a second blocker that only showed up on trying the full thing: with flat layout
plus `pythonpath = ["."]`, `import aiobotocore` resolves to the checkout even when the wheel is
installed in the venv, because pytest inserts the rootdir at the front of `sys.path`. So
`needs: build` + `pip install` alone would have silently kept testing source. #1686 handles
that, and #1685 tracks the `src/` layout move that makes it structural.

#1686 keeps what's good here — both dists exercised, `enable-cache: false`, SHA-pinned actions —
and adds: hatchling so sdist contents are declared in `pyproject.toml`, `check-sdist` in
pre-commit, the trio-style collect-and-diff gate (python-trio/trio#2541), and the actual
dependency inversion @webknjaz asked for, with the matrix running out of the sdist against the
installed wheel.

PR #1599, opened 2026-05-09
fix(release): publish via ci-cd.yml workflow_dispatch

thehesiod:
The first live `auto-release-on-merge.yml` run for v3.7.0 (#1595) failed at the publish step:

```
The claims in this token suggest that the calling workflow is a reusable workflow.
...
Reusable workflows are not currently supported by PyPI's Trusted Publishing
functionality, and are subject to breakage.
...
##[error]Trusted publishing exchange failure:
* invalid-publisher: valid token, but no corresponding publisher
```

PyPI's Trusted Publisher matches the OIDC token's `workflow_ref` claim. The reusable-publish.yml refactor in #1597 made `auto-release-on-merge.yml` the entry point — so OIDC reported `workflow_ref=auto-release-on-merge.yml`, which doesn't match PyPI's existing `ci-cd.yml` config.

Without PyPI config write access (would need org owner), the only fix is to **make `ci-cd.yml` the actual entry point of the publish run**.

## Solution: dispatch ci-cd.yml from auto-release-on-merge.yml

`workflow_dispatch` is one of the few event types that fires from `GITHUB_TOKEN`-driven calls (the rest are blocked by GitHub's loop-prevention rule). So `auto-release-on-merge.yml` calls `gh workflow run ci-cd.yml -f tag=$VERSION`, and the resulting run has `workflow_ref=ci-cd.yml` — matching PyPI's existing config.

```mermaid
flowchart LR
    A[Release vX.Y.Z PR merges] --> B[auto-release-on-merge.yml]
    B --> C[create signed tag + draft Release]
    C --> D["gh workflow run ci-cd.yml -f tag=X.Y.Z"]
    D --> E[ci-cd.yml runs with workflow_ref=ci-cd.yml]
    E --> F[full test matrix + build]
    F --> G[pypi-publish job - PyPI Trusted Publisher matches]
```

## Changes

- **`ci-cd.yml`**: add `workflow_dispatch` trigger with `tag` input. Re-inline `pypi-publish` job (was reusable-publish.yml in #1597 — that's the refactor PyPI's restriction invalidates). Gate fires on `tag-push OR workflow_dispatch with tag input`. Version sourced from `inputs.tag || github.ref_name`.
- **`auto-release-on-merge.yml`**: drop the `build` and `publish` jobs. After `tag-and-release`, `gh workflow run ci-cd.yml --ref $VERSION -f tag=$VERSION`.
- **`reusable-publish.yml`**: deleted. No callers, and PyPI rejects it as a reusable anyway.

## Trade-offs

- Publish becomes asynchronous from `auto-release-on-merge`'s view. Maintainer monitors the dispatched ci-cd.yml run separately. Acceptable — the GitHub Release stays draft until the maintainer publishes it manually after PyPI confirms.
- The dispatched ci-cd.yml run re-runs the full test matrix before publishing. Slower (~5min for matrix) but a real safety win — publish is gated on tests passing on the tag commit.
- Manual `git push origin <tag>` from a maintainer's local still works (tag-push trigger fires, same `pypi-publish` job runs).

## Immediate fix for v3.7.0

This PR doesn't republish 3.7.0. After it merges, choices for v3.7.0:

1. **Manual `twine upload` from local** — download the build artifact from the failed run (https://github.com/aio-libs/aiobotocore/actions/runs/25597481808), `twine upload dist/*` with PyPI API token. Fastest.
2. **Dispatch ci-cd.yml manually** after this PR lands — `gh workflow run ci-cd.yml --ref 3.7.0 -f tag=3.7.0` from your local. Clean, runs full CI gate.

## Test plan

- [x] Pre-commit clean.
- [ ] First post-merge release (next time `draft-release` is invoked + a release PR merges) exercises the dispatch path end-to-end.


thehesiod:
Both addressed in `3d288b1`:

- **actions: write** added to the job's permissions block alongside `contents: write`. Verified: `gh workflow run` does call `POST /repos/.../actions/workflows/.../dispatches` which requires that scope; without it the dispatch would 403 and the new release path would silently fail on the next release. Good catch — would have re-broken the same flow this PR is meant to fix.
- `outputs.version` declaration dropped (no consumer left after the publish job was removed).

Also acknowledged the third observation (`--ref $VERSION` ordering): the tag is created synchronously by `gh release create --target` via the Releases API in the previous step, so by the time `gh workflow run --ref $VERSION` runs, the tag is visible. No race.

PR #1600, opened 2026-05-09
fix(release): dispatch ci-cd.yml against merge commit SHA, not tag

thehesiod:
Two related fixes surfaced after #1599 landed:

## 1. Drop `--draft` so the tag actually gets created

After #1599 merged, manually running `gh workflow run ci-cd.yml --ref 3.7.0 -f tag=3.7.0` to publish v3.7.0 hit:

```
PS> gh workflow run ci-cd.yml --ref 3.7.0 -f tag=3.7.0
could not create workflow dispatch event: HTTP 422: No ref found for: 3.7.0
```

Root cause: `gh release create --draft` does NOT create the tag at the target commit — the tag is only created when the Release is published (un-drafted). So `--ref $VERSION` 422s because the tag literally doesn't exist as a ref yet. The same bug was in `auto-release-on-merge.yml`'s dispatch step.

Earlier exploration considered keeping `--draft` and dispatching with `--ref $MERGE_SHA` instead — but `gh workflow run --ref` only accepts branch or tag names, not commit SHAs. Dead end.

**Fix:** drop `--draft` from `gh release create`. The Releases API creates the signed tag immediately at `--target $MERGE_SHA`, and the dispatch step can `--ref $VERSION` directly.

**Trade-off:** the GitHub Release becomes visible the moment `auto-release-on-merge.yml` finishes — ~5-10 min before PyPI catches up via the dispatched ci-cd.yml run. A watcher hitting the release page during that window sees a release that `pip install` can't find for a few minutes. Acceptable in exchange for:

- No more manual un-draft step (release becomes fully unattended)
- Simpler workflow (one ref, no SHA-vs-tag confusion)
- Tag exists for the dispatch (no 422)

If `--draft` safety becomes important later, the path is to provision a PAT/App token, push the tag with that, and let `ci-cd.yml`'s existing tag-push trigger handle publish — that lets us go back to draft semantics. Tracked as a follow-up; not needed for the current flow.

## 2. Switch `claude.yml` from Opus to Sonnet

`claude.yml` runs Opus for every PR review and `@claude` response. ~30 `claude.yml` runs in the past 4 hours during this development session, ~5-10 actually executing reviews — Opus per-run cost is $1-5, so daily cost climbs fast.

PR-review and `@claude` traffic in this repo is shape-equivalent to the `botocore-sync` classify job, which has been on Sonnet from the start (~$0.10/run, vs ~$1+ for Opus on the same workload). No observable quality gap for those tasks.

Switched the default `--model` to `sonnet`. Future PR can add prompt-content / label-based routing to escalate to Opus on demand.

## Test plan

- [x] Pre-commit clean.
- [ ] Next live `Release v...` PR merge: verify the tag is created immediately (no draft), the dispatch fires successfully against `--ref X.Y.Z`, and `pypi-publish` runs to completion.
- [ ] Next PR push: verify `claude.yml` runs on Sonnet (check the run log; cost-meter should drop noticeably across the next ~10 runs).

### DIFF
--- a/.github/workflows/auto-release-on-merge.yml
+++ b/.github/workflows/auto-release-on-merge.yml
@@ -129,6 +129,7 @@ jobs:
       env:
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         VERSION: ${{ steps.ver.outputs.version }}
+        MERGE_SHA: ${{ github.event.pull_request.merge_commit_sha }}
       # Trigger ``ci-cd.yml`` via ``workflow_dispatch`` so the actual
       # PyPI publish runs from there. PyPI's Trusted Publisher matches
       # against the OIDC token's ``workflow_ref``, which for a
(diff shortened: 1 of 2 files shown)


thehesiod:
@jakob-keller due to a painful infra situation, i've had to update it to not be a draft and instead go ahead and create the release to ensure the shas match.  we still have to manually approve the publish and after that a 15m cooldown, so i think we're still ok.  also of course you have to merge the PR

thehesiod:
also bumped down the pr reviewer to sonnet, because i burned through >$50 in api tokens for this just today

PR #1610, opened 2026-05-20
chore(coverage): configure coverage tooling to handle unreachable code properly

### LINKED ISSUE #1490: [TODO] Let coverage tooling handle unreachable code properly (and remove redundant/oudated bits)
> [!caution]
> This is a popular topic of @webknjaz's rants 🤪

I checked the coverage presented on Codecov and noticed a few common mistakes in the coverage collection/configuration.

Look at https://app.codecov.io/gh/aio-libs/aiobotocore/commit/fd5f917524ce98ce5b48b062f2d78815cebe60d6/tree?search=tests%2F&displayType=list and you'll see that the test folder has a bunch of dead code. This ain't good — this means that some tests you might think run don't. This kind of thing may happen when you copy a test but forget to rename the function and so the first one gets shadowed and never runs. Or some logic is bad in helpers.
Some bits might be expected to never run and that's fine, however, having those line marked as red makes it harder to rely on the coverage data.
And so it's recommended to require 100% coverage on the test folder. This is achievable by using line exclusions and "no cover", "no branch" pragmas that would make coveragepy ignore the lines you don't want to matter completely so that they don't contribute to the coverage measurements at all.
I also noticed some mentions of Python 2 in vendored tests and some possibly outdated branches.

So to fix this, it's important to review the uncovered tests/helpers and add pragmas where reasonable (including justification comments for each). In some cases, you'll find that you might just remove excessive exception handling or helpers that are never in use.
Once you clean that up, configure Codecov to specifically require 100% on the tests and make the check required.

Going further, it's a good idea to apply similar treatment to the runtime code — although, it's better to actually cover the cases with tests and pragma justifications might be more rare. One corner case you might find would be branch coverage dependent on an OS or a Python version. For these, there's coveragepy plugins that provide you with custom pragmas that let you make certain lines as not needing coverage under respective conditions. This could be the second stage, though. At some point you might want to make use of `covdefaults` when the code base is ready.

---

Remember: this ain't about the coverage metric ripped out of the context — this is about setting up the tooling to help you spot problems. This would also make Codecov reports more stable and patch/project percentage will stop being affected by the number of lines in the project changing in PRs (which influences the ratio and contributes to visual confusion when it looks like you've got the entire patch covered but the metric ends up being skewed).

---

ctx: https://nedbatchelder.com/blog/200507/sometimes_the_automation_really_knows_best / https://nedbatchelder.com/blog/200710/flaws_in_coverage_measurement / https://nedbatchelder.com/blog/201106/running_coverage_on_your_tests / https://nedbatchelder.com/blog/201908/dont_omit_tests_from_coverage / https://nedbatchelder.com/blog/202008/you_should_include_your_tests_in_coverage

kings9527:
Configure coverage.py exclude_also patterns for standard unreachable code paths (TYPE_CHECKING, NotImplementedError, AssertionError, Protocol classes, abstract methods, __repr__, if __name__ == '__main__').

Add 'tests/' to coverage source so dead code in test helpers is visible.

Add tests/ flag to Codecov with 100% target to catch dead test code early.

Annotate existing unreachable/functional-test-only paths with explicit pragma comments instead of relying on implicit branch coverage gaps.

Fixes #1490

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -17,7 +17,7 @@ coverage:
     # and shadowed test functions early.
     tests:
       paths:
-        - tests/
+      - tests/
       target: 100%
 
 flags:
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -9,16 +9,16 @@ coverage:
   range: "95..100"
 
   status:
-    project: false
+    project:
+      default:
+        target: auto
+      tests:
+        paths:
+        - tests/
+        target: 100%
     patch:
       default:
         target: auto
-    # Require 100% coverage on tests directory to catch dead code
-    # and shadowed test functions early.
-    tests:
-      paths:
-      - tests/
-      target: 100%
 
 flags:
   library:
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -10,6 +10,15 @@ coverage:
 
   status:
     project: false
+    patch:
+      default:
+        target: auto
+    # Require 100% coverage on tests directory to catch dead code
+    # and shadowed test functions early.
+    tests:
+      paths:
+        - tests/
+      target: 100%
 
 flags:
   library:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -181,7 +181,22 @@ docstring-code-line-length = "dynamic"
 exclude_also = [
     "if TYPE_CHECKING",
     "@pytest.mark.localonly",
+    "pragma: no cover",
+    "pragma: no branch",
+    "raise AssertionError",
+    "raise NotImplementedError",
+    "def __repr__",
+    "if 0:",
+    "if __name__ == .__main__.:",
+    "class .*\\bProtocol\\):",
+    "@(abc\\.)?abstractmethod",
+    "Used in functional tests only",
+    "used in functional tests",
 ]
+
+[tool.coverage.run]
+# Include tests in coverage so dead code in tests is visible.
+source = ["aiobotocore", "tests"]
 
 [tool.rumdl]
 line-length = 120
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -181,8 +181,6 @@ docstring-code-line-length = "dynamic"
 exclude_also = [
     "if TYPE_CHECKING",
     "@pytest.mark.localonly",
-    "pragma: no cover",
-    "pragma: no branch",
     "raise AssertionError",
     "raise NotImplementedError",
     "def __repr__",
@@ -190,8 +188,6 @@ exclude_also = [
     "if __name__ == .__main__.:",
     "class .*\\bProtocol\\):",
     "@(abc\\.)?abstractmethod",
-    "Used in functional tests only",
-    "used in functional tests",
 ]
 
 [tool.coverage.run]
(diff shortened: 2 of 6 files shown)


thehesiod (dismissed):

thehesiod:
@claude review

thehesiod:
@kings9527 what do you think of the review?

kings9527:
Thanks for the review, @thehesiod. The points from @claude are solid—I'll address all three:

1. Remove the fragile comment-text patterns from  (those paths are already covered by  annotations)
2. Drop the redundant  /  entries from 
3. Fix the Codecov  block placement under  instead of as a sibling to /

Will push a follow-up commit shortly.

kings9527:
To clarify the three items from the review above (the backticks got mangled in the first comment):

1. Remove `"Used in functional tests only"` and `"used in functional tests"` from `exclude_also` — these are comment-text patterns, not code patterns, and the relevant lines already have `# pragma: no cover` annotations.
2. Remove `"pragma: no cover"` and `"pragma: no branch"` from `exclude_also` — `coverage.py` handles these natively, so they are redundant there.
3. Move the `tests:` block in `.codecov.yml` under `project:` instead of as a sibling to `project`/`patch`.

kings9527:
Pushed the fixes in :

- : removed all four flagged entries from  — comment-text patterns and redundant pragma lines are gone
- : moved  block under  so Codecov actually picks it up

Ready for another look.

thehesiod:
@claude review

thehesiod:
@kings9527 looks like there are some test failures, otherwise good to go

thehesiod (approved):

thehesiod:
@kings9527 : Commits must have verified signatures.

thehesiod (commented):
Reviewed current head `d601a7c`. I found one blocking inconsistency in the new coverage policy and one secondary scope concern; details are inline. The focused suite passed (1,207 tests), and the GitHub test matrix is green, but `codecov/project/tests` is failing and the PR is currently blocked. The unsigned commits are already covered by the existing maintainer comment.

.codecov.yml
```diff
@@ -9,7 +9,16 @@ coverage:
...
 
   status:
-    project: false
+    project:
+      default:
+        target: auto
+      tests:
+        paths:
+        - tests/
+        target: 100%
```
thehesiod, line 18: **[P1] Do not introduce a coverage gate that already fails on this tree**

At this head, Codecov reports 5,313 of 5,382 covered test statements (98.71%), leaving 69 misses across 17 files. Because this line sets the target to 100%, `codecov/project/tests` is failing and GitHub reports the PR as `BLOCKED`; the new gate cannot pass on the commit that introduces it. Please either finish classifying, covering, or explicitly excluding the remaining test lines, or stage this as a ratchet (for example, a no-regression `project/tests` target plus 100% `patch/tests`) until the tree actually reaches 100%.

.codecov.yml
```diff
@@ -9,7 +9,16 @@ coverage:
   range: "95..100"
 
   status:
-    project: false
+    project:
+      default:
```
thehesiod, line 13: **[P2] Preserve the explicitly disabled default project check**

The base configuration has `project: false`, but this `default` stanza turns the general `codecov/project` status back on with `target: auto` in addition to adding the tests-only target; the live PR now shows that extra status. This is broader than the stated tests-coverage change and can block future PRs on aggregate project coverage. If re-enabling general project coverage is not intentional, keep the default project status disabled while adding the named tests target.

PR #1648, opened 2026-07-05
chore: apply ruff RET/C4/PIE cleanups across aiobotocore/ and tests/

Justin-Ahn:
Mechanical fixes only — no behavior change. Applied via `ruff check --select=RET,C4,PIE --fix --unsafe-fixes --preview` followed by `ruff format`. Categories:

- RET502/503/501: implicit-return handling — explicit `return None` where the function has other return paths, dropped bare `return None` where the fall-through is trivial.
- RET504: unnecessary assignment-then-return collapsed.
- RET505/506: `else:` after `return`/`raise` removed to reduce nesting.
- C408: `dict()` / `list()` / `tuple()` calls with no args rewritten as `{}` / `[]` / `()` literals.
- C414: `list(list(x))` and similar unnecessary double-processing removed.
- C416: `[x for x in y]` → `list(y)` where the comprehension is a no-op.
- C420: `{k: None for k in xs}` → `dict.fromkeys(xs)`.
- PIE790/808: unnecessary `pass` placeholders and `range(0, n)` starts.

### Description of Change

### Assumptions

* [ ] If this is resolving an issue (needed so future developers can
  determine if change is still necessary and under what conditions)
  (can be provided via link to issue with these details):
* [ ] If this is providing a new feature (can be provided via link to issue with these details):
  * [ ] Why needed

* [ ] I have read and followed [CONTRIBUTING.rst](https://github.com/aio-libs/aiobotocore/blob/main/CONTRIBUTING.rst#how-to-upgrade-botocore)
* [ ] I have added URL to diff: <https://github.com/boto/botocore/compare/>[CURRENT_BOTO_VERSION_TAG]...[NEW_BOTO_VERSION_TAG]

### DIFF
--- a/aiobotocore/awsrequest.py
+++ b/aiobotocore/awsrequest.py
@@ -22,8 +22,7 @@ class AioAWSResponse(AWSResponse):
         encoding = botocore.utils.get_encoding_from_headers(self.headers)
         if encoding:
             return (await self.content).decode(encoding)
-        else:
-            return (await self.content).decode('utf-8')
+        return (await self.content).decode('utf-8')
 
     @property
     def text(self):
--- a/aiobotocore/client.py
+++ b/aiobotocore/client.py
@@ -128,8 +128,7 @@ class AioClientCreator(ClientCreator):
             base_classes=bases,
         )
         class_name = get_service_module_name(service_model)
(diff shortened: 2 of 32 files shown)


Justin-Ahn:
apologies -- this PR was a side effect of other changes I've been experimenting with.

PR #1662, opened 2026-07-17
fix(draft-release): commit to the release branch, not main

thehesiod:
### Description of Change

The `Draft Release` workflow has been silently broken since v3.7.1: [run 29548268161](https://github.com/aio-libs/aiobotocore/actions/runs/29548268161/job/87785150787) reported success but opened no PR. `mcp__github_file_ops__commit_files` resolves its target branch **once at MCP-server startup** from `CLAUDE_BRANCH`; in agent mode that falls back to `GITHUB_REF_NAME`, so a `workflow_dispatch` run on `main` had the tool committing to `main` — rejected by branch protection (`Changes must be made through a pull request` / merge queue / `codecov/patch`). The tool has **no per-call branch argument**, so the agent's attempts to pass `branch:` / `ref:` were silently ignored, and it aborted after writing a correct changelog. The action still exited 0, so the run went green.

This sets `CLAUDE_BRANCH: claude/release-<run-id>` on the action step — the supported escape hatch (`src/modes/agent/index.ts`: `claudeBranch || GITHUB_HEAD_REF || GITHUB_REF_NAME || defaultBranch`). The MCP tool still does the commit, signed, as intended. It also adds a guard step that fails the job when no release PR was opened, so this can't fail silently again.

Not caused by the recent `claude-code-action` bump — the June 6 run hit the identical 422 on v1.0.152 and only produced a PR because that agent improvised a `gh api` fallback.

### Corrections to CLAUDE.md

Three documented claims were wrong and had actively misled the bot into aborting:

1. *"the MCP commit tool needs the target branch to already exist"* — no: `getOrCreateBranchRef` creates it off `BASE_BRANCH` on 404. It **reuses** an existing ref, which is why the release branch is now run-scoped: a stale branch from a failed run would otherwise be committed on top of.
2. *"call `commit_files` with `branch:` explicitly set"* — there is no such parameter. The documented workaround could never have worked.
3. *"the MCP tool is the only path that ... produces signed commits"* — it has no signing step. It is a thin wrapper over `POST /git/{blobs,trees,commits}` + `PATCH /git/refs`; GitHub signs any API-created commit with the token's identity. The v3.7.1 draft was made by hand-rolling those same calls and its commit (`cc8814f6`) is `verified: true`. `git push` remains banned — that one genuinely is unsigned.

### Assumptions

The release branch name is informational: `auto-release-on-merge.yml` keys off the PR title (`Release v...`), not the branch, so dropping the version from the branch name in favour of the run id is safe. This is stated in the skill's own step 7.

  * [x] Alternative methods considered (if any) — the `gh api` git-data fallback (works, but hand-rolls what the MCP tool already does); `CLAUDE_BRANCH` is the supported path
  * [x] How issue can be reproduced — dispatch `Draft Release` on `main`; before this change `commit_files` targets `main` and 422s
* [ ] If this is providing a new feature — N/A, bug fix

### Verification

- `uv run pre-commit run` passes on all three files, including workflow-schema validation.
- `$CLAUDE_BRANCH` reaches the agent's Bash: `parse-sdk-options.ts` builds SDK env as `{ ...process.env }`.
- The guard's `gh pr list --head` was tested both ways — resolves #1624 by head branch, returns empty for a nonexistent one.
- Latent bug fixed alongside: the skill never spelled out `gh pr create`, and the runner stays on `main`, so an inferred head resolves to `main` and fails. Now explicitly `--head "$CLAUDE_BRANCH"`.

The end-to-end proof is a real dispatch after merge; `dry_run` deliberately skips the commit path, so it cannot exercise this.

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -136,9 +136,12 @@ jobs:
         # The file-ops MCP server commits to a branch fixed at server
         # startup from CLAUDE_BRANCH; there is no per-call branch arg. In
         # agent mode this defaults to GITHUB_REF_NAME (``main``), whose
-        # branch protection rejects the commit. Run-scoped so a failed
-        # run's leftover branch is never committed on top of.
-        CLAUDE_BRANCH: claude/release-${{ github.run_id }}
+        # branch protection rejects the commit. Scoped per *attempt* so a
+        # failed attempt's leftover branch is never committed on top of —
+        # file-ops reuses an existing ref, and run_id alone is stable
+        # across re-runs.
+        # yamllint disable-line rule:line-length
+        CLAUDE_BRANCH: claude/release-${{ github.run_id }}-${{ github.run_attempt }}
       with:
         anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
         use_commit_signing: true
@@ -175,7 +178,8 @@ jobs:
       if: ${{ !inputs.dry_run }}
       env:
         GH_TOKEN: ${{ github.token }}
-        BRANCH: claude/release-${{ github.run_id }}
+        # yamllint disable-line rule:line-length
+        BRANCH: claude/release-${{ github.run_id }}-${{ github.run_attempt }}
       run: |
         url=$(gh pr list --head "$BRANCH" --base main --state open \
           --json url --jq '.[0].url // empty')
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -132,6 +132,13 @@ jobs:
     # yamllint disable-line rule:line-length
     - uses: anthropics/claude-code-action@536f2c32a39763739000b0e1ac69ca2647d97ce9  # v1.0.170
       id: claude
+      env:
+        # The file-ops MCP server commits to a branch fixed at server
+        # startup from CLAUDE_BRANCH; there is no per-call branch arg. In
+        # agent mode this defaults to GITHUB_REF_NAME (``main``), whose
+        # branch protection rejects the commit. Run-scoped so a failed
+        # run's leftover branch is never committed on top of.
+        CLAUDE_BRANCH: claude/release-${{ github.run_id }}
       with:
         anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
         use_commit_signing: true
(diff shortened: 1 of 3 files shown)


.github/workflows/draft-release.yml
```diff
@@ -132,6 +132,13 @@ jobs:
     # yamllint disable-line rule:line-length
     - uses: anthropics/claude-code-action@536f2c32a39763739000b0e1ac69ca2647d97ce9  # v1.0.170
       id: claude
+      env:
+        # The file-ops MCP server commits to a branch fixed at server
+        # startup from CLAUDE_BRANCH; there is no per-call branch arg. In
+        # agent mode this defaults to GITHUB_REF_NAME (``main``), whose
+        # branch protection rejects the commit. Run-scoped so a failed
+        # run's leftover branch is never committed on top of.
+        CLAUDE_BRANCH: claude/release-${{ github.run_id }}
```
thehesiod: [P2] Include the run attempt in the release branch

`github.run_id` identifies the workflow run, not an individual attempt; [GitHub documents](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context) that it does not change when the workflow is re-run, while `github.run_attempt` increments. A retry from the Actions UI will therefore reuse this same `claude/release-<run_id>` ref. Since file-ops deliberately reuses an existing ref, the retry can commit on top of the failed attempt; if an earlier attempt already opened a PR before failing, the new guard can also find that existing PR and let a later no-op/aborted attempt go green. Please include `${{ github.run_attempt }}` in both `CLAUDE_BRANCH` and the guard’s `BRANCH` (and update the skill/docs examples) so every attempt gets fresh state.

amohr-northstar:
Reviewed the current head (`bb760099`). The central `CLAUDE_BRANCH` fix matches the pinned action implementation, and the changed-file checks are green.

I found one retry-safety issue to address before merge: [`github.run_id` is reused across workflow re-runs](https://github.com/aio-libs/aiobotocore/pull/1662#discussion_r3599971704), so both the release branch and the new PR-existence guard can reuse stale state. Including `github.run_attempt` in the branch name and guard will make each attempt independent.

No other actionable findings.

.github/workflows/draft-release.yml
amohr-northstar: Confirmed and fixed in f93d26b. Both `CLAUDE_BRANCH` and the guard's `BRANCH` are now `claude/release-${{ github.run_id }}-${{ github.run_attempt }}`; `CLAUDE.md` and the skill's step 7 updated to match, with a note on why `run_id` alone is insufficient.

You're right on both counts, and the second-order effect is the part that makes this more than cosmetic: a retry finding the previous attempt's PR would let an aborted run report success — reintroducing exactly the silent-success failure this PR exists to remove. I'd noticed the run_id/run_attempt distinction while writing it and filed it as minor; that was a misjudgement, since it lands squarely in the bug class being fixed.

Issue #1685, opened 2026-08-01
Move to a src/ layout


thehesiod:
## Summary

Move the package from a flat layout (`aiobotocore/` at the repo root) to a `src/` layout
(`src/aiobotocore/`).

This is deliberately **not** bundled with the "test the dists" work, because renaming the
package directory touches every path in the repo and would conflict with anything in flight.
Filing it so the reasoning isn't lost.

## Why

With a flat layout, `aiobotocore/` and `tests/` live in the same directory, so a single
`sys.path` entry controls both. That single entry is over-subscribed:

* Tests must be able to import `tests.mock_server` — `conftest.py` forces the multiprocessing
  `spawn` start method, and the child process unpickles `AIOServer` by its qualified name
  `tests.mock_server.AIOServer`, inheriting the parent's `sys.path`.
* The suite must **not** be able to import `aiobotocore` from the checkout, or CI silently
  tests the source tree instead of the wheel it just installed.

Both are served by the same directory, so the two requirements are in direct conflict.
Measured, running the suite out of an unpacked sdist with the wheel installed:

| rootdir on `sys.path`? | installed wheel wins | moto-server tests |
|-|-|-|
| no (`--import-mode=importlib`, no `pythonpath`) | pass | fail — `ModuleNotFoundError: No module named 'tests'` |
| yes | fail — checkout shadows the wheel | pass |

The current workaround is in `tests/conftest.py`: **append** (never insert) the rootdir to
`sys.path`, so `spawn` children can import `tests` while `site-packages` still precedes the
source tree. It works, but it is an ordering convention. Any future `sys.path.insert(0, ...)`,
or a plugin that prepends the rootdir, silently flips CI back to testing the checkout — and
the failure is invisible, because everything still passes.

A `src/` layout removes the conflict instead of balancing it: the rootdir would hold `tests/`
but no importable `aiobotocore`, so putting it on `sys.path` is simply safe. This is PyPA's
stated rationale for the layout:

> if an import package exists in the current working directory with the same name as an
> installed import package, the variant from the current working directory will be used

— https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

urllib3, cryptography, pip, attrs, black, flask, poetry and trio all use `src/` for this reason.

## What it takes

Mechanical, one commit:

- [ ] `git mv aiobotocore src/aiobotocore`
- [ ] `[tool.hatch.version] path` → `src/aiobotocore/__init__.py`
[...]

PR #1667, opened 2026-07-17
feat: run the httpx backend on trio

graingert:
Closes #749, which asked for aiobotocore to be usable from a project using trio and httpx.
The httpx backend landed since; this makes it run on trio.

aiobotocore's httpx backend only ran on asyncio, even though httpx itself runs on any anyio backend. Reaching for asyncio primitives in shared code paths meant a trio caller hit `RuntimeError: no running event loop` as soon as it built an SSL context, streamed a request body, retried, waited, paginated with `result_key_iters`, used adaptive retries, or resolved credentials via IMDS.

The httpx backend now works under trio. aiohttp is asyncio-only, so it is untouched and keeps using asyncio directly.

## How the two backends are kept apart

Selection is by the session class in use, following the existing subclassing pattern — each `Anyio*` class overrides only the primitive its `Aio*` counterpart gets from asyncio.

**anyio does not become a dependency of aiobotocore.** It is imported only on httpx code paths, where it is already installed as a hard dependency of httpx.

| Area | Change |
| --- | --- |
| `httpxsession.py` | `anyio.to_thread.run_sync`, `anyio.sleep` |
| `_tee.py` (new), `paginate.py`, `client.py` | `AnyioPaginator`/`AnyioPageIterator` with a trio-safe `tee` for `result_key_iters` (`aioitertools.tee` is built on `asyncio.Queue`) |
| `waiter.py`, `endpoint.py` | `AnyioWaiter`, `AnyioEndpoint` |
| `retries/bucket.py`, `retries/adaptive.py` | `AnyioTokenBucket`, `AnyioClientRateLimiter` |
| `utils.py`, `credentials.py`, `session.py` | `AnyioIMDSRegionProvider`, `AnyioInstanceMetadataFetcher`, `AnyioContainerMetadataFetcher`, `AnyioContainerProvider` |

## Proxies

The httpx backend previously raised `NotImplementedError` for any proxy configuration. It now supports proxies on par with the aiohttp backend: a per-scheme mounted httpx transport, proxy TLS (`proxy_ca_bundle` / `proxy_client_cert`). The `AsyncClient` is built lazily on the first send and shared across endpoint hosts, preserving the configured session-wide connection-pool cap. This removes the `skip_httpx` test fixture — the proxy tests now run on both backends.

[...]

PR #1686, opened 2026-08-01
ci: test the built dists instead of the checkout

### LINKED ISSUE #1169: Test distribution packages in CI

thehesiod:
## Summary

Makes CI test the **built dists** rather than the Git checkout, which is what #1169 asks for:

> Don't depend on the tests. If anything, it's the tests that should depend on building the
> dists. This way you'd be able to test said dists instead of building things in those jobs
> again.

Closes #1169.

Pointing CI at the real artifact immediately found a live packaging bug. Fixing that made the
suite's own unreliability the limiting factor, so the PR has a second half: making a red run
mean something.

## The bug this found

The matrix installed the project from source via `uv sync`, so nothing ever exercised the wheel
or the sdist. That hides a packaging bug that is live on `main` today:

| | |
|-|-|
| test modules in `tests/` | 57 |
| test modules in `main`'s sdist | **30** |
| missing from the sdist | **27** — incl. `tests/__init__.py`, `tests/conftest.py`, `tests/mock_server.py`, all of `tests/botocore_tests/` |

`pytest` on an unpacked sdist dies immediately:

```
tests/test_stubber.py:5: from .mock_server import AIOServer
E   ImportError: attempted relative import with no known parent package
!!!! Interrupted: 5 errors during collection !!!!
```

Cause: no `MANIFEST.in`, and `[tool.setuptools.packages.find] include = ["aiobotocore*"]` keeps
tests out of the wheel — so the only test files that shipped were the ones distutils' legacy
default rule globs (`tests/test*.py`), without `conftest.py`, `__init__.py` or subdirectories.
Worst of both: dead weight in the sdist *and* an sdist nobody can test. Debian, Fedora,
conda-forge and Gentoo all build from the sdist and run the suite.

After this PR the sdist carries all 57, and the suite collects and passes out of it.

## Testing the dist

**Packaging.** setuptools → hatchling + `hatch-fancy-pypi-readme`, so sdist contents are
declared in `pyproject.toml` instead of a `MANIFEST.in`. The sdist now ships `tests/`,
`scripts/`, `uv.lock`, `CONTRIBUTING.rst`, `examples/` and `docs/` — enough for
`uv sync --frozen` + `pytest` to work on a bare unpack. `check-sdist` in pre-commit fails if a
git-tracked file is neither included nor listed under `[tool.check-sdist] git-only`.

**Dependency inversion.** `test` now `needs: build`, takes its working tree from the sdist via
`re-actors/checkout-python-sdist` with no `actions/checkout` underneath it, and installs
aiobotocore from the built wheel (`uv pip install --no-index --find-links dist/`) rather than
from source. So the matrix workspace is the sdist and nothing else — a file that failed to ship
has nowhere to hide.

**Test config moved into `pyproject.toml`.** Testing an installed wheel means the checkout must
not shadow it, which the Makefile's `python -m pytest` guaranteed it would. Import mode,
`testpaths`, warning filters, marker strictness and coverage path mapping now live in
`[tool.pytest.ini_options]` / `[tool.coverage.*]`, so a bare `pytest` behaves like CI and the
suite is reproducible from an unpacked sdist. `pythonpath` drops `.` — which put the source tree
ahead of site-packages — in favour of `scripts`.

**Makefile replaced by `[tool.poe.tasks]`.** CI reached the suite through `uv run make mototest`
and pre-commit through `uv run make pre-commit`, so `make` was a hop that only forwarded
arguments — and it could not work from an unpacked sdist. uv has no task runner of its own
(`[project.scripts]` are console entry points that ship in the wheel, not dev tasks), so the
tasks move to `poethepoet`, which keeps them in `pyproject.toml` next to the pytest and coverage
config they belong with. CI and contributors run the same `uv run poe <task>`.

**Dependency groups.** A `test` group split out of `dev`, so CI installs what the suite needs
without repo tooling (`pre-commit`, `anthropic`). `tests/evals` drives `plugins/`, so it stays
git-only and now runs in the `lint` job, the only job with a full checkout.

## Making the result trustworthy

The suite was failing 14 tests locally before any of this; it is now 0, and a full local run
takes ~45s. Each cause was real, not flake:

- **`~/.aws/config` leaked into the suite.** A developer's `retry_mode = adaptive` /
  `max_attempts = 20` produced 10 phantom failures and a 10x slowdown. An autouse fixture now
  points `AWS_CONFIG_FILE` / `AWS_SHARED_CREDENTIALS_FILE` at an empty directory and strips
  `AWS_*`.
- **moto's backends are process-global.** A per-test `ThreadedMotoServer` isolates nothing, so
  hardcoded resource names collided across tests in the same xdist worker. Names come from
  `random_name()` now.
- **Creates are not idempotent.** botocore retrying a timed-out `CreateStack` / `CreateTable` /
  `CreateFunction` against moto comes back as "already exists". Retries are off in the shared
  `config` fixture, with timeouts at 60s so a genuinely slow response fails outright.
- **`AIOServer` leaked a process per use** and took >30s to bind on 3.12 specifically. It is an
  event-driven thread now, and `always_spawn` goes with it.
- **`--reruns` is gone**, along with `pytest-rerunfailures`: retrying hides flakes instead of
  fixing them, doubles the cost of every failure, and nothing was marked flaky.
- **Leaks are failures.** `ResourceWarning` and `PytestUnraisableExceptionWarning` are errors,
  which took closing three anyio listeners in the proxy/TLS helpers, botocore's CSM monitor
  socket, and two unread S3 response bodies. It has already paid for itself: it caught a
  never-awaited `AsyncMock` coroutine in the CRT credential tests, latent since they landed,
  because `client.meta` is sync but `AsyncMock` makes every child call a coroutine.
- **`test_endpoint_client_cert_is_not_offered_to_proxy` was failing ~60% of local runs** —
  anyio force-closes on handshake failure, so the client races the TLS alert against the RST.
- **`test_publish_to_http` called `httpbin.org`** on every run; it uses a local endpoint now.

## Cheaper, more honest CI

- **coverage was pinned to `==7.2.7`**, which ships no C tracer for 3.12+ — so coverage silently
  fell back to the pure-Python tracer at **+158%** instead of +20%. Unpinned to `>=7.10,<8`.
- **Coverage ran on all 13 matrix jobs and uploaded 26 times** (the eval step used `--cov-append`,
  re-sending the whole suite's data under a second flag — which is why `evals-common` read 92% on
  `main` without measuring evals). It now runs on the four 3.14 x86 jobs, one per backend
  variant: interpreter version changes almost nothing in the report, but the backend does —
  `aiobotocore/_httpx.py` is only reached by the legacy-httpx job.
- **Two coverage-reporting bugs.** Measuring tests via `source` made `tests` an XML `<source>`
  root, so files were written bare (`test_utils.py`) and Codecov could not tell them from their
  `tests/botocore_tests/unit/` namesakes — it silently dropped all five ambiguous ones.
  Separately, `source_pkgs` applies on top of the evals run's `--cov`, so that upload reported
  all of aiobotocore at ~0%.
- **Superseded PR runs cancel**, with the group key keeping merge-queue and tag/dispatch publish
  runs out of it.
- **`--dist worksteal`** instead of the default `load`: the server-heavy tests sort late, so
  `load` handed them out together and the last 10% of the suite took ~44% of the wall clock.
- **The moto Lambda image is pulled in the background** while dependencies install.

## Known wrinkle

Flat layout leaves one rough edge. `--import-mode=importlib` keeps the rootdir off `sys.path`
so the checkout cannot shadow the installed wheel — but ten modules do
`from tests.mock_server import ...`, which needs `tests` importable. `conftest.py` therefore
**appends** (never inserts) the rootdir: the imports resolve, and `site-packages` still precedes
the source tree. #1685 tracks the `src/` layout move that makes this structural instead of an
ordering convention — deliberately not bundled here, since renaming the package directory would
conflict with everything in flight.

## Test plan

- [x] `uv build` + `twine check --strict` pass on both dists.
- [x] sdist test modules 30 → 57, matching the checkout exactly.
- [x] `check-sdist` negative-tested both ways on throwaway commits: dropping `/tests` from
      `include` fails it (rc=2, naming every missing module), while adding a path to `exclude`
      does not — its hatchling backend forgives the `exclude` list by design. Documented in
      `CONTRIBUTING.rst` and `CLAUDE.md`.
- [x] Full local suite green with warnings-as-errors: 1657 passed, 0 failed, 0 warnings.
- [x] Same with `--group awscrt` installed: 1683 passed, 0 failed.
- [x] Leak enforcement verified against deliberately leaked coroutines, and isolated to confirm
      `PytestUnraisableExceptionWarning` is the filter that catches the GC-finalized ones.
- [x] All pre-commit hooks green.
- [x] CI green on the real artifact path.


pyproject.toml
```diff
@@ -90,31 +98,115 @@ docs = [
...
+# so what you run locally is what CI runs. Anything that is a property of the
+# suite rather than of one invocation belongs in the sections above, not here.
+[tool.poe.tasks]
+lint = "pre-commit run --all --show-diff-on-failure"
+test.cmd = "pytest -s -vv"
+test.deps = ["lint"]
+vtest.cmd = "pytest -s -vv"
+vtest.env = { PYTHONTRACEMALLOC = "5", PYTHONFAULTHANDLER = "1" }
+cov.cmd = "pytest -s -vv --cov --cov-report=term --cov-report=html"
+cov.deps = ["lint"]
```
thehesiod: **Medium** — `poe test` and `poe cov` cannot run from an unpacked sdist, but the sdist ships the doc that tells you to run them.

`.pre-commit-config.yaml` is deliberately git-only, and `pre-commit` also needs a git repo. So `deps = ["lint"]` aborts the task graph before pytest ever starts. Reproduced against the sdist built from this commit:

```
$ tar xzf aiobotocore-3.8.0.tar.gz && cd aiobotocore-3.8.0
$ uv run poe test
Poe => pre-commit run --all --show-diff-on-failure
An error has occurred: FatalError: git failed. Is it installed, and are you in a Git repository directory?
Error: Task graph aborted after failed task 'lint'
```

`CONTRIBUTING.rst` is in the sdist and documents both commands, and the sdist block one section up states the goal as "a downstream packager unpacks it, runs `uv sync --frozen` ... and gets a runnable suite": https://github.com/aio-libs/aiobotocore/blob/85fc0bed50a5fc45f3606d4d5602bd3ce4221243/pyproject.toml#L122-L128

Related: `uv sync --frozen` on a bare unpack resolves `default-groups = ["dev"]`, so it installs `pre-commit` and `anthropic` — the repo-only tooling the `test` group was split out to avoid. The documented incantation would need to be `uv sync --frozen --no-default-groups --group test`.

Options: drop `deps = ["lint"]` (CI runs lint in its own job anyway), or keep it and document `poe mototest` / bare `pytest` as the sdist-side entry point.

`mototest`, `vtest` and `docs` are unaffected; `clean` also needs a git repo.

CONTRIBUTING.rst
```diff
@@ -40,16 +40,48 @@ To run the non-moto tests, make sure you have your amazon key and secret accessi
...
 
 To run individual use following command::
 
     $ uv run pytest -sv tests/test_monitor.py -k test_name
 
+Everything lives in ``pyproject.toml``: the tasks above under
+``[tool.poe.tasks]`` (run ``uv run poe`` to list them), and the test contract
+itself -- import mode, marker strictness, warning filters, coverage paths --
+under ``[tool.pytest.ini_options]`` and ``[tool.coverage.*]``. CI runs the same
+tasks, and a bare ``pytest`` behaves the same as ``poe test``, so use whichever
```
thehesiod: **Low** — "a bare `pytest` behaves the same as `poe test`, so use whichever you prefer" isn't accurate: `test.deps = ["lint"]` makes `poe test` run the full `pre-commit run --all --show-diff-on-failure` first, which bare `pytest` skips.

```
$ uv run poe -d test
Poe => pre-commit run --all --show-diff-on-failure
Poe => pytest -s -vv
```

The pytest *contract* is identical (that part is true and is the useful claim); the task isn't. Suggest "a bare `pytest` runs the suite the same way `poe test` does, minus the pre-commit pass it depends on."

.github/workflows/reusable-lint.yml
```diff
@@ -48,4 +48,10 @@ jobs:
         restore-keys: pre-commit-${{ runner.os }}-
     - name: Run pre-commit hooks
       run: |
-        uv run make pre-commit
+        uv run poe lint
+    # tests/evals exercises plugins/, which the sdist excludes -- so the test
+    # matrix, which runs out of the sdist, never sees them. This is the only
+    # job with a full checkout.
```
thehesiod: **Low** — "This is the only job with a full checkout" isn't true: `build` also runs `actions/checkout` (it has to, to build the dists from source): https://github.com/aio-libs/aiobotocore/blob/85fc0bed50a5fc45f3606d4d5602bd3ce4221243/.github/workflows/reusable-build.yml#L18-L20

Only the `test` matrix swapped its checkout for `checkout-python-sdist`. Worth correcting since the comment is the stated justification for putting the evals run here rather than in `build` — and `build` is arguably the better home given it already has both the checkout and a collect-venv.

CLAUDE.md
```diff
@@ -32,14 +32,33 @@ Key constraints:
...
+
+# Packaging
+
+Built by `hatchling`; sdist contents are declared in `[tool.hatch.build.targets.sdist]`.
+
+CI tests the dist, not the checkout: `build` runs first, then the matrix unpacks the sdist
+as its working tree and installs the wheel over it. So:
+
+- New test-suite runtime dependency → the `test` dependency group (`dev` is repo tooling only).
+- New top-level file or directory → either the sdist `include` list or
```
thehesiod, line 76: **Low** — the make→poe migration is complete in `CLAUDE.md`, `CONTRIBUTING.rst`, `.readthedocs.yml` and `botocore-sync-prompt.md`, but three files the repo's own automation reads still point at the old world:

- `plugins/aiobotocore-bot/skills/draft-release/SKILL.md:140` classifies a PR into the "contrib" bucket from paths ``under `.github/`, `tests/`, `Makefile`, `pyproject.toml` `` — `Makefile` is deleted by this PR, so that clause can never match again.
- `.github/claude-review-prompt.md:265` — `Run pre-commit: uv run pre-commit run --all --show-diff-on-failure`
- `plugins/aiobotocore-bot/skills/analyze-pr-feedback/SKILL.md:134` — same invocation.

The last two still work (`pre-commit` is in the default `dev` group), so they're consistency-only; the `SKILL.md:140` one is a dead path reference. `.github/*-prompt.md` is explicitly in scope per this file's own AI-workflow conventions: https://github.com/aio-libs/aiobotocore/blob/85fc0bed50a5fc45f3606d4d5602bd3ce4221243/CLAUDE.md#L118-L121

pyproject.toml
```diff
@@ -90,31 +98,115 @@ docs = [
...
+  --log-cli-level=DEBUG --http-backend=${HTTP_BACKEND}
+"""
+mototest.env = { HTTP_BACKEND.default = "all", PYTHONTRACEMALLOC = "5", PYTHONFAULTHANDLER = "1" }
+
+clean.shell = "git clean -Xdf"
+
+[tool.coverage.run]
+relative_files = true
+source_pkgs = ["aiobotocore"]
+source = ["tests"]
```
thehesiod: **Low (coordination)** — open PR #1610 adds its own `[tool.coverage.run]` table to this same file with different contents (`source = ["aiobotocore", "tests"]`, plus `[tool.coverage.report]` exclusions). Two PRs creating the same TOML table will collide on whichever merges second, and TOML rejects a duplicate table outright rather than merging it. Worth deciding merge order or folding #1610's exclusions in here.

Same for #1667 (trio): it rewrites the dependency-group split and touches `reusable-test.yml`'s `uv sync --group awscrt` line that this PR replaces with the `--group=awscrt` env flag.

One intended-but-worth-stating consequence of this section: `tests/evals` moves to the `lint` job, which uploads no coverage, so those 3 files / ~265 statements leave the Codecov report.

thehesiod (commented):
### Code review

Found 5 issues, posted as inline comments below. The packaging/CI mechanism itself verified clean: built the sdist locally, confirmed `uv sync --frozen` + full collection works from a bare unpack, and confirmed `uv sync --no-install-project` prunes a stale `aiobotocore` from a restored `.venv` cache (so the wheel install is not a silent no-op). Tooling is already at the simplest currently-available option — `uv_build` still requires a static version and has no readme concatenation, uv still has no task runner ([astral-sh/uv#5903](https://github.com/astral-sh/uv/issues/5903) open), and `check-sdist` v1.5.0 / `checkout-python-sdist` v3.0.0 are the latest releases.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

pyproject.toml
thehesiod: Fixed in 21d46fc — dropped `deps = ["lint"]` from `test` and `cov`. CI runs lint as its own job, so nothing relied on the chaining, and the sdist stays runnable. Verified against a bare unpack of the sdist built from that commit:

```
$ tar xzf aiobotocore-3.8.0.tar.gz && cd aiobotocore-3.8.0
$ uv sync --frozen --no-default-groups --group test
$ uv run --no-sync poe test -k test_get_waiter_invalid_waiter_name
Poe => pytest -s -vv
1 passed, 1 skipped, 2363 deselected
```

Also documented that exact incantation in CONTRIBUTING.rst, including why `--no-default-groups --group test` is required, and noted that `lint` and `clean` are the two tasks that need a git repo.

CONTRIBUTING.rst
thehesiod: Fixed in 21d46fc. Dropping `deps = ["lint"]` makes the original claim true rather than needing the caveat, but I reworded it anyway to say what is actually being asserted: "a bare `pytest` runs the suite exactly as `poe test` does".

.github/workflows/reusable-lint.yml
thehesiod: Fixed in 21d46fc. Corrected to the real reason: this job syncs the default groups, so it already has the `anthropic` dependency `tests/evals` imports. `build` has a checkout but its collect-venv is `--no-default-groups --group test`, so putting the evals run there would mean an extra install.

CLAUDE.md
thehesiod: All three fixed in 21d46fc: dropped `Makefile` from the contrib path list in `draft-release/SKILL.md`, and switched `claude-review-prompt.md` and `analyze-pr-feedback/SKILL.md` to `uv run poe lint`.

pyproject.toml
thehesiod: Acknowledged — not fixable in this PR, flagging for merge order.

#1610: whichever merges second has to fold its `[tool.coverage.run]` into the existing table rather than add a second one. This PR needs `relative_files` and the `[tool.coverage.paths]` remap because coverage now measures the wheel in site-packages; #1610 wants `source = ["aiobotocore", "tests"]` and report exclusions. Those compose fine in one table, it just has to be done by hand.

#1667: overlaps on the dependency-group split and on the `uv sync --group awscrt` line this PR replaces with the `AWSCRT_GROUP` env flag.

On the coverage consequence: correct, `tests/evals` no longer reports. Happy to add `--cov` to the evals step in the lint job and upload it as a separate flag if the drop matters, but it looked like noise for 3 files of repo tooling.

tests/mock_server.py
```diff
@@ -17,59 +16,64 @@
...
+        self._stop = threading.Event()
+        self._error = None
+        self._thread = None
 
     def _run(self):
-        asyncio.set_event_loop(asyncio.new_event_loop())
+        loop = asyncio.new_event_loop()
+        try:
+            loop.run_until_complete(self._serve())
+        except BaseException as exc:
```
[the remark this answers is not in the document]
thehesiod: Fixed in f139832.

The broad catch was guarding a real case: an exception before `site.start()` would leave `_ready` unset, so `__aenter__` would wait out its full timeout and then report a misleading "never bound a port". But `finally` covers that better than catching `BaseException` does.

Now:

```python
except Exception as exc:
    self._error = exc
finally:
    # Always unblock __aenter__, even if the loop died before binding.
    self._ready.set()
    loop.close()
```

plus an explicit `endpoint_url is None` check in `__aenter__`, so a thread that exits before binding reports that instead of handing back a half-built server. `Exception` is sufficient for the error-reporting path now that unblocking is unconditional.

.github/workflows/ci-cd.yml
```diff
@@ -33,6 +33,17 @@ on:
...
         type: string
 
+# Superseded PR runs are wasted work -- a push should cancel the run it
+# replaces. Two things must NOT be cancelled, and the group key handles both:
+# ``event_name`` keeps pull_request, push, merge_group and workflow_dispatch in
+# separate groups, and on ``main`` the SHA is folded in so every commit gets its
+# own group. So merge-queue runs and the tag/dispatch publish path always run to
+# completion; only a superseded PR head is cancelled.
+concurrency:
+  # yamllint disable-line rule:line-length
```
webknjaz: Using `>-` would give you better readability instead of a long line and a rule to suppress.

.github/workflows/ci-cd.yml
```diff
@@ -111,7 +126,11 @@ jobs:
       enable-cache: ${{ github.ref_type == 'tag' && 'false' || 'auto' }}
       with-awscrt: ${{ matrix.with-awscrt }}
       httpx-extra: ${{ matrix.httpx-extra }}
-      timeout-minutes: ${{ matrix.timeout-minutes || 5 }}
+      # One job measures coverage for the whole matrix; it costs ~20% and the
+      # report is near-identical across interpreters.
+      # yamllint disable-line rule:line-length
+      coverage: ${{ matrix.python-version == '3.14' && matrix.os == 'ubuntu-24.04' && matrix.httpx-extra == 'httpx2' && !matrix.with-awscrt }}
```
webknjaz: Same w/ multi-line

thehesiod:
@webknjaz morning! :)

.github/workflows/reusable-build.yml
```diff
@@ -37,3 +37,38 @@ jobs:
...
+        uv venv "${RUNNER_TEMP}/collect-venv"
+        uv export --frozen --no-emit-project --no-default-groups --group test \
+          --extra httpx2 --output-file "${RUNNER_TEMP}/test-reqs.txt"
+        uv pip install --python "${RUNNER_TEMP}/collect-venv" \
+          -r "${RUNNER_TEMP}/test-reqs.txt"
+        uv pip install --python "${RUNNER_TEMP}/collect-venv" --no-deps dist/*.whl
+    - name: Unpack the sdist
+      run: |
+        mkdir -p "${RUNNER_TEMP}/sdist-src"
+        tar xzf dist/*.tar.gz -C "${RUNNER_TEMP}/sdist-src" --strip-components=1
```
webknjaz: Why can't you just use my action instead of reinventing the wheel, though?

.github/workflows/ci-cd.yml
```diff
@@ -111,7 +126,11 @@ jobs:
       enable-cache: ${{ github.ref_type == 'tag' && 'false' || 'auto' }}
       with-awscrt: ${{ matrix.with-awscrt }}
       httpx-extra: ${{ matrix.httpx-extra }}
-      timeout-minutes: ${{ matrix.timeout-minutes || 5 }}
+      # One job measures coverage for the whole matrix; it costs ~20% and the
+      # report is near-identical across interpreters.
```
webknjaz: What's lost then? Include it explicitly, perhaps, and document?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
-requires = ["setuptools>=77.0.0"]
-build-backend = "setuptools.build_meta"
+requires = ["hatchling>=1.27.0", "hatch-fancy-pypi-readme>=24.1.0"]
```
webknjaz, line 2: Move packaging refactoring into a standalone effort.

.github/workflows/ci-cd.yml
thehesiod: Done — folded both, plus the two long expressions I'd added in `reusable-test.yml` (the `.venv` cache key and `COV_ARGS`). Verified with `yaml.safe_load` that each folds back to the byte-identical string, since a break outside a `${{ }}` would have injected a space into the value:

```yaml
concurrency:
  group: >-
    ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}${{
    github.ref == format('refs/heads/{0}', github.event.repository.default_branch)
    && format('-{0}', github.sha)
    || '' }}
```

The remaining suppressions in these files are all on `uses:` lines, where the trailing `# vX.Y.Z` pin comment would be folded into the action ref.

.github/workflows/ci-cd.yml
thehesiod: Same treatment — this one is a single expression, so it folds cleanly:

```yaml
      coverage: >-
        ${{ matrix.python-version == '3.14'
        && matrix.os == 'ubuntu-24.04' }}
```

.github/workflows/reusable-build.yml
thehesiod: I do use it — `reusable-test.yml` runs the whole matrix out of it, which is the change this PR is really about.

It doesn't work for *this* check though, because the action unpacks with `tar --directory=$GITHUB_WORKSPACE --strip-components=1` over the existing tree. That's exactly right when the sdist *replaces* the checkout, but here it would overlay it: a test module missing from the sdist would still be sitting there from `actions/checkout`, so the collection would match and the gate would pass. The check needs the two trees side by side — the checkout's collected node IDs are the reference the sdist's are diffed against — and there's no target-directory input to put them anywhere else.

Happy to swap it in if you'd rather I wipe the workspace first and use the action, but the manual `tar` into `$RUNNER_TEMP` seemed like less machinery than `git clean -xdff` + removing tracked files.

.github/workflows/ci-cd.yml
thehesiod: Good push — I had it too narrow. It now measures every 3.14 x86 job, i.e. one per backend variant (httpx2, legacy httpx, no-httpx, awscrt), and the reasoning is in the comment:

```yaml
      # Coverage costs ~20%, and interpreter version changes almost nothing in
      # the report -- but the backend does: aiobotocore/_httpx.py is only
      # reached by the legacy-httpx job, and awscrt swaps whole code paths. So
      # measure every 3.14 x86 job (one per backend variant) and no others.
```

What the one-job version actually lost, from the Codecov diff: `aiobotocore/_httpx.py` 100% → 70% and `tests/test_httpx_compat.py` 100% → 87.3%, both the legacy-httpx job's contribution, plus drops in `credentials.py` and `retryhandler.py`. All of that is back.

What is still not measured is interpreter-specific code, and in this package that is exactly one line — the `typing_extensions` fallback at `aiobotocore/config.py:20`. arm64 and 3.15 execute no distinct source.

Worth noting the previous behaviour was 26 uploads per run: every one of the 13 matrix jobs uploaded twice, since the eval step used `--cov-append` and so re-sent the whole suite's data under a second flag. That second flag is why `evals-common` read 92% on `main` and is also what made it look like a cliff here once the append was dropped.

.github/workflows/reusable-test.yml
```diff
@@ -25,11 +25,16 @@ on:
         # 'httpx' (deprecated legacy fallback), or '' (no httpx backend).
         type: string
         required: true
+      coverage:
+        # Coverage costs ~20% and the report is near-identical across
+        # interpreters, so only one job in the matrix measures it.
+        type: boolean
+        default: false
```
thehesiod, line 33: **Medium** — "only one job in the matrix measures it" is off by 4x. The gate is `matrix.python-version == '3.14' && matrix.os == 'ubuntu-24.04'`, which matches four jobs in the current matrix:

- `Python 3.14 on ubuntu-24.04 httpx2`
- `Python 3.14 on ubuntu-24.04 httpx`
- `Python 3.14 on ubuntu-24.04 no-httpx`
- `Python 3.14 on ubuntu-24.04 with-awscrt httpx2`

ci-cd.yml's own comment on the same input gets it right ("measure every 3.14 x86 job (one per backend variant)"): https://github.com/aio-libs/aiobotocore/blob/dd1a0bb882ec92a77e184ae0d3fbbe1fded3849e/.github/workflows/ci-cd.yml#L131-L137

Worth fixing the wording here, since this is the comment a reader hits first when asking why their job produced no coverage — and "one job" invites someone to later narrow the expression and silently drop the legacy-httpx and awscrt code paths, which are the only reason four jobs are measured.

tests/conftest.py
```diff
@@ -196,16 +226,20 @@ def http_session_cls(request) -> type:
...
-        read_timeout=read_timout,
-        connect_timeout=connect_timeout,
-        **config_kwargs,
+        **{
+            'region_name': region,
+            'signature_version': signature_version,
+            'read_timeout': read_timout,
+            'connect_timeout': connect_timeout,
+            # Creates aren't idempotent: retrying one that timed out against local moto reports "already exists".
+            'retries': {'max_attempts': 0},
```
thehesiod, line 233: **Medium** — `alternative_s3_client` is exempt from this fix. It builds its own `AioConfig` instead of using the `config` fixture, so it still has the pre-PR `read_timeout=5` / `connect_timeout=5` and no `retries` override:

https://github.com/aio-libs/aiobotocore/blob/dd1a0bb882ec92a77e184ae0d3fbbe1fded3849e/tests/conftest.py#L335-L343

That is exactly the combination the comment on this line calls out — a tight timeout plus botocore's default retry behaviour, against a moto backend, on `CreateBucket`, which is not idempotent. With `--reruns` also gone there is nothing left to absorb it.

Suggest having it depend on `config` and override only `region_name`, so the two client fixtures cannot drift again.

thehesiod (commented):
### Code review (round 2, after the trio/anyio merge)

All five findings from the previous round are resolved — re-verified by building the sdist and running `poe test` from a bare unpack with no `.git`, which now works.

Found 4 issues, posted as inline comments below. Test failures excluded at the author's request.

Mechanisms I verified rather than assumed, all correct: the `COV_ARGS` gating (poe treats empty-but-set as set, so `.default` does *not* leak coverage flags onto the non-coverage jobs); the `concurrency` group key (the publish path dispatches with `--ref <tag>`, so `github.ref` is `refs/tags/X.Y.Z` and can never share a group with another run — no publish can be cancelled); and `retries={'max_attempts': 0}` against botocore's `_compute_retry_max_attempts`, where it means one total attempt rather than "unset".

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

CLAUDE.md
```diff
@@ -32,14 +32,43 @@ Key constraints:
...
 
+## Naming moto resources
+
+**Never hardcode a moto resource name — use `random_name()` from `tests/conftest.py`.**
+The `moto_server` fixture is function-scoped, but moto's backends are process-global
+singletons: a fresh `ThreadedMotoServer` on a new port still sees resources created by
+earlier tests in the same xdist worker. Cleaning up at the end of the test is not enough,
+because a test that fails partway (a read timeout, say) leaks its resources, and every
+later test using the same hardcoded name then fails with `AlreadyExists` — including its
+own `--reruns` retry and the other HTTP-backend parametrization.
```
thehesiod: **Low** — this new guidance cites a mechanism the same PR removes. `--reruns` is gone from `mototest`, and the task comment now says so explicitly: https://github.com/aio-libs/aiobotocore/blob/dd1a0bb882ec92a77e184ae0d3fbbe1fded3849e/pyproject.toml#L218-L221

So "including its own `--reruns` retry" describes behaviour that no longer exists. Worth correcting because `CLAUDE.md` is autoloaded into every session, so a stale claim here is re-read on every future task.

While editing: "the other HTTP-backend parametrization" is now also narrower than reality — after the trio merge the axes are async backend x HTTP backend, so a hardcoded name collides across up to four combinations, not two.

pyproject.toml
```diff
@@ -72,20 +77,27 @@ botocore-dev = [
...
     "anyio >= 4.11.0, < 5",
     "dill >= 0.3.3, < 0.5", # Used in test_patches.py
     "docker >= 7.1, < 8",
     "moto[server,s3,sqs,awslambda,dynamodb,cloudformation,sns,batch,ec2,rds] >= 5.1.8, < 6",
     "packaging >= 24.1, < 27", # Used in test_version.py
-    "pre-commit >= 3.5.0, < 5",
+    "poethepoet >= 0.29, < 1", # Runs [tool.poe.tasks]; CI drives the suite through it
     "pytest-mock >= 3.14.1, < 4", # Used in test_session.py
-    "pytest-rerunfailures >= 16.0.1, < 17", # Used in test_lambda.py
+    "pytest-rerunfailures >= 16.0.1, < 17", # Available for opt-in @pytest.mark.flaky
```
thehesiod: **Low** — `pytest-rerunfailures` is now unused. `--reruns` is gone from every task, and `grep -rn 'mark.flaky' tests/` returns zero hits, so the "available for opt-in" comment describes an option nobody exercises. It is in the `test` group, so it installs on every matrix job and ships in the resolved dependency set a downstream packager gets.

Either drop it, or keep it and say why the option is worth holding open — as written the comment reads as justification for a dependency that no longer has a caller.

.github/workflows/reusable-build.yml
thehesiod: You were right, just about a different wheel than I was defending — I've deleted the whole thing.

`check-sdist` is already wired into the lint job with `[tool.check-sdist] git-only` listing the deliberate exclusions, so "a git-tracked file didn't make it into the sdist" is caught there. The collect-and-diff steps were re-deriving that, badly. And since `reusable-test.yml` uses your action with *no* `actions/checkout` underneath it, the matrix workspace is the sdist and nothing else — a file that failed to ship has nowhere to hide there either.

(For what it's worth, looking at how yarl/multidict/frozenlist use it: they run `actions/checkout` first and let the action overlay it, which is what led me to think the overlay was the obstacle rather than the reason the check was unnecessary in the first place.)

Build job is now just build → twine check → upload.

.github/workflows/reusable-test.yml
thehesiod: Fixed — the input description now matches the gate:

```yaml
      coverage:
        # Coverage costs ~20%. Interpreter version changes almost nothing in
        # the report but the backend does, so ci-cd.yml turns this on for one
        # job per backend variant (currently the four 3.14 x86 jobs).
```

tests/conftest.py
thehesiod: Good catch, and the drift was live: those two timeouts were the pre-PR values with botocore's default retries still on. Now derived rather than rebuilt, as suggested:

```python
    config = config.merge(AioConfig(region_name=alternative_region))
```

`signature_version` and `request` drop out of the signature since `config` already resolves both (including `config_kwargs`). `test_can_follow_signed_url_redirect` and `test_bucket_redirect` still pass across all backends.

CLAUDE.md
thehesiod: Fixed. Rewritten to describe what the suite actually does now:

> including the test's own sibling parametrizations, of which there are up to four (async backend x HTTP backend). Nothing retries, so the first leak is permanent for that worker.

pyproject.toml
thehesiod: Dropped, and `uv.lock` regenerated — it was installing on every matrix job and shipping in the resolved set for no caller. If we ever want an opt-in flaky marker back it is one line to restore, and by then it would come with an actual use.

CLAUDE.md
```diff
@@ -32,14 +32,44 @@ Key constraints:
...
+
+Built by `hatchling`; sdist contents are declared in `[tool.hatch.build.targets.sdist]`.
+
+CI tests the dist, not the checkout: `build` runs first, then the matrix unpacks the sdist
+as its working tree and installs the wheel over it. So:
+
+- New test-suite runtime dependency → the `test` dependency group (`dev` is repo tooling only).
+- New top-level file or directory → either the sdist `include` list or
+  `[tool.check-sdist] git-only`. The `check-sdist` hook fails until it is in one of them.
+- New test file → it must reach the sdist. The build job diffs collected node IDs between
```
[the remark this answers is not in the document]
thehesiod: Correct on both counts — fixed in 3ba6813.

I went and checked what `check-sdist` actually guarantees before rewriting the claim, because "the file lists happen to line up" turned out not to be the real gap. Its hatchling backend does this:

```python
def git_only_excludes(self, pyproject, files, source_dir):
    exclude = pyproject["tool"]["hatch"]["build"]["targets"]["sdist"]["exclude"]
    return pathspec_filter(exclude, files)
```

So every path in the sdist `exclude` list is treated as legitimately git-only. Verified both directions against the real hook on a throwaway commit:

- add `/tests/test_sqs.py` to `exclude` → the module really is absent from the tarball, and check-sdist still reports "SDist matches git", rc=0.
- drop `/tests` from `include` → rc=2, and it lists every missing test module.

So the collect-diff was not covering "ships but isn't discoverable" (an unimportable module is a collection *error*, and the matrix runs out of the sdist, so that fails there). The one thing it covered that nothing else does is someone adding a path to `exclude`. That is a visible one-line pyproject edit rather than a silent packaging accident, so both docs now say exactly that and tell readers to keep the list minimal.

pyproject.toml
```diff
@@ -101,31 +112,139 @@ evals = [
...
-[tool.setuptools.dynamic]
-version = { attr = "aiobotocore.__version__" }
-readme = { file = ["README.rst", "CHANGES.rst"] }
+[tool.hatch.version]
+path = "aiobotocore/__init__.py"
 
-[tool.setuptools.packages.find]
-include = ["aiobotocore*"]
+[tool.hatch.metadata.hooks.fancy-pypi-readme]
+content-type = "text/x-rst"
```
webknjaz, line 122: It's easier to move to MD across the board. MyST-parser is your friend.

pyproject.toml
```diff
@@ -101,31 +112,139 @@ evals = [
...
+[tool.check-sdist]
+git-only = [
+    ".*",
+    "AGENTS.md",
+    "CLAUDE.md",
+    "plugins",
+    "tests/evals",
+]
 
 [tool.pytest.ini_options]
```
webknjaz, line 166: I still think that having this file as a dumping ground for configs of different tools is a mass-maintenance disaster. It may work when you are fine with special-casing each individual project scaffold but is not a great engineering decision.

pyproject.toml
```diff
@@ -101,31 +112,139 @@ evals = [
...
+
+clean.shell = "git clean -Xdf"
+
+# ``tests`` belongs in source_pkgs, not source: as a source *path* it becomes an
+# XML <source> root, so files are written bare (``test_utils.py``) and Codecov
+# can't tell them from their tests/botocore_tests/unit/ namesakes -- it drops the
+# ambiguous ones. As a package the rootdir stays the root and paths stay prefixed.
+[tool.coverage.run]
+relative_files = true
+source_pkgs = ["aiobotocore", "tests"]
```
webknjaz, line 241: Not really the best config. You actually want source=. with source_pkgs=aiobotocore

webknjaz: Also mv this to a `.coveragerc.toml`

PR #1696, opened 2026-08-06
fix: ensure concurrency-safe session management in `HTTPSession`

### LINKED ISSUE #1695: Concurrent requests each create their own ClientSession (session cache race in `_get_session`)
**Describe the bug**

`AIOHTTPSession._get_session()` is supposed to cache one `aiohttp.ClientSession` per
`proxy_url`. Since 3.8.0 the cache fill is no longer atomic: concurrent first-use on a single
client each creates its own `ClientSession` / `TCPConnector`. Only the last one is stored in
`self._sessions`; the rest stay alive in `self._exit_stack` but are unreachable.

Consequence: one client can hold many connection pools instead of one. `max_pool_connections`
is applied per connector, so the effective limit is multiplied by the number of duplicates, and
the extra sockets live until the client is closed.

Introduced by [#1587](https://github.com/aio-libs/aiobotocore/pull/1587), which made
`_create_connector` a coroutine that awaits `asyncio.to_thread(...)`. That added a suspension
point between the cache check and the cache store:

```python
async def _get_session(self, proxy_url):
    if not (session := self._sessions.get(proxy_url)):
        connector = await self._create_connector(proxy_url)   # suspends here
        self._sessions[proxy_url] = session = await self._exit_stack.enter_async_context(
            aiohttp.ClientSession(connector=connector, ...)
        )
    return session
```

While the first caller is suspended in `to_thread`, every other caller also sees an empty cache
and takes the same branch. In 3.7.0 `_create_connector` was synchronous and
`ClientSession.__aenter__` does not yield, so the fill was effectively atomic.

Moving the SSL build off the loop is the right change; the cache fill just needs to stay
single-flight (e.g. a per-`proxy_url` `asyncio.Lock`, or a placeholder/future installed in
`self._sessions` before the `await`).

Still present on `main` at the time of writing.

**Checklist**

- [x] I have tried the same code in botocore to ensure this is an aiobotocore specific issue
- [x] I have tried similar code in aiohttp to ensure this is is an aiobotocore specific issue

**pip freeze results**

Reproducing env (`aiobotocore==3.8.0`). `pip check` reports no broken requirements.

```
aiobotocore==3.8.0
aiohappyeyeballs==2.7.1
aiohttp==3.14.3
aioitertools==0.13.0
aiosignal==1.4.0
attrs==26.1.0
botocore==1.43.46
frozenlist==1.8.0
idna==3.18
jmespath==1.1.0
multidict==6.7.1
propcache==0.5.2
python-dateutil==2.9.0.post0
six==1.17.0
typing_extensions==4.16.0
urllib3==2.7.0
wrapt==2.3.0
yarl==1.24.5
```

**Environment:**

- Python Version: 3.12.3
- OS name and version: Linux 5.10.209-198.858.amzn2.x86_64 (Amazon Linux 2), x86_64, glibc 2.39

**Additional context**

### Reproduce

Ten concurrent requests on one client. No network or credentials needed — the connection
failures are irrelevant; the session count is the point.

```python
import asyncio

async def main():
    session = get_session()
    async with session.create_client(
        "s3",
        region_name="us-east-1",
        aws_access_key_id="x",
        aws_secret_access_key="y",
        endpoint_url="http://127.0.0.1:1",
    ) as client:
        await asyncio.gather(
            *(client.list_buckets() for _ in range(10)), return_exceptions=True
        )
        http = client._endpoint.http_session
        print("created:", len(http._exit_stack._exit_callbacks))
        print("cached: ", len(http._sessions))

asyncio.run(main())
```

Same script, two throwaway envs (identical Python / aiohttp; `botocore` cannot be held
constant because the two aiobotocore releases pin non-overlapping ranges):

```bash
python -m venv /tmp/aio-370 && /tmp/aio-370/bin/pip install 'aiobotocore==3.7.0'
python -m venv /tmp/aio-380 && /tmp/aio-380/bin/pip install 'aiobotocore==3.8.0'
/tmp/aio-370/bin/python repro.py
/tmp/aio-380/bin/python repro.py
```

| Version | created | cached |
|---|---|---|
| 3.7.0 | 1 | 1 |
| 3.8.0 | **10** | 1 |

Or more directly, without going through a client — this never enters botocore, so holds
everything except aiobotocore constant:

```python
import asyncio
from aiobotocore.httpsession import AIOHTTPSession

async def main():
    async with AIOHTTPSession() as http:
        sessions = await asyncio.gather(*(http._get_session(None) for _ in range(10)))
    print("distinct ClientSession objects:", len({id(s) for s in sessions}))

asyncio.run(main())
```

| Version | distinct |
|---|---|
| 3.7.0 | 1 |
| 3.8.0 | **10** |

### Checklist notes

- **botocore:** `botocore.httpsession.URLLib3Session` is synchronous and has no
  equivalent of this async `_get_session` cache. There is no concurrent-await race to
  reproduce there. The second snippet above isolates the bug to
  `aiobotocore.httpsession.AIOHTTPSession` without involving botocore at all.
- **aiohttp:** `aiohttp.ClientSession` / `TCPConnector` do not implement this
  `proxy_url`-keyed session cache or the `asyncio.to_thread` SSL-build path. The race is in
  aiobotocore's caching layer on top of aiohttp. `aiohttp` was held at 3.14.3 across both
  aiobotocore versions above.
- **regression:** 3.7.0 → 1 session; 3.8.0 → 10 sessions. Same Python 3.12.3, same
  `aiohttp==3.14.3`. Still present on `main`.

jakob-keller:
### Description of Change

Ensure concurrency-safe session management in `HTTPSession`.

### Assumptions

None

* [x] If this is resolving an issue (needed so future developers can
  determine if change is still necessary and under what conditions)
  (can be provided via link to issue with these details): closes #1695
* [ ] If this is providing a new feature (can be provided via link to issue with these details):
  * [ ] Why needed

* [ ] I have read and followed [CONTRIBUTING.rst](https://github.com/aio-libs/aiobotocore/blob/main/CONTRIBUTING.rst#how-to-upgrade-botocore)
* [ ] I have added URL to diff: <https://github.com/boto/botocore/compare/>[CURRENT_BOTO_VERSION_TAG]...[NEW_BOTO_VERSION_TAG]

### DIFF
--- a/aiobotocore/httpsession.py
+++ b/aiobotocore/httpsession.py
@@ -83,6 +83,7 @@ class AIOHTTPSession:
 
         # TODO: handle socket_options
         # keep track of sessions by proxy url (if any)
+        self._sessions_lock = asyncio.Lock()
         self._sessions: dict[str | None, aiohttp.ClientSession] | None = None
         self._verify = verify
         self._proxy_config = ProxyConfiguration(
@@ -128,10 +129,11 @@ class AIOHTTPSession:
 
     async def __aexit__(self, exc_type, exc_val, exc_tb):
         assert self._sessions is not None, 'Session was never entered'
-        self._sessions.clear()
-        await self._exit_stack.aclose()
-        # Make _sessions unusable once context is exited
-        self._sessions = None
+        async with self._sessions_lock:
+            self._sessions.clear()
+            await self._exit_stack.aclose()
+            # Make _sessions unusable once context is exited
+            self._sessions = None
 
     def _get_ssl_context(self):
         return create_urllib3_context()
@@ -222,17 +224,19 @@ class AIOHTTPSession:
 
     async def _get_session(self, proxy_url):
         if not (session := self._sessions.get(proxy_url)):
-            connector = await self._create_connector(proxy_url)
-            self._sessions[proxy_url] = (
-                session
-            ) = await self._exit_stack.enter_async_context(
-                aiohttp.ClientSession(
-                    connector=connector,
-                    timeout=self._timeout,
-                    skip_auto_headers={'CONTENT-TYPE'},
-                    auto_decompress=False,
-                ),
-            )
+            async with self._sessions_lock:
+                if not (session := self._sessions.get(proxy_url)):
+                    connector = await self._create_connector(proxy_url)
+                    self._sessions[proxy_url] = (
+                        session
+                    ) = await self._exit_stack.enter_async_context(
+                        aiohttp.ClientSession(
+                            connector=connector,
+                            timeout=self._timeout,
+                            skip_auto_headers={'CONTENT-TYPE'},
+                            auto_decompress=False,
+                        ),
+                    )
 
         return session
 


jakob-keller:
@claude please review

thehesiod (approved):

PR #1704, opened 2026-08-09
fix: use generated SignIn OAuth method for login refresh

### LINKED ISSUE #1697: `AioLoginCredentialFetcher` calls `create_o_auth2_token`, a ClientMethodAlias `AioBaseClient.__getattr__` cannot resolve — `login_session` credential refresh always fails
### Summary

`aiobotocore/credentials.py` mirrors botocore's `login_session` refresh path line-for-line, including its call to **`client.create_o_auth2_token(...)`**. That name is not a generated client method — it is a backwards-compatibility **`ClientMethodAlias`** that botocore registers and resolves in `BaseClient.__getattr__`. `AioBaseClient.__getattr__` raises `AttributeError` unconditionally (deliberately, per #789), so the aliased name can never resolve on an async client and every refresh raises:

```
AttributeError: 'SignInService' object has no attribute 'create_o_auth2_token'
```

The real generated method, **`create_oauth2_token`** (no underscore after `o`), exists on both clients. Calling it fixes the path with no behaviour change.

### Reproduction

```python
import asyncio
from aiobotocore.session import AioSession

KW = dict(region_name="us-east-1", aws_access_key_id="x", aws_secret_access_key="y")

async def main():
    async with AioSession().create_client("signin", **KW) as c:
        return hasattr(c, "create_o_auth2_token"), hasattr(c, "create_oauth2_token")

print(asyncio.run(main()))   # -> (False, True)
```

Measured side by side against a sync botocore client, same service, same versions:

| client | `create_o_auth2_token` (alias) | `create_oauth2_token` (real) |
|---|---|---|
| `botocore` (sync) | **True** | True |
| `aiobotocore` (async) | **False** ← the bug | True |

### Why it happens

`botocore/handlers.py` registers exactly **six** `ClientMethodAlias` entries, enumerated live:

```
getattr.mturk.list_hi_ts_for_qualification_type       -> list_hits_for_qualification_type
getattr.socialmessaging.delete_whatsapp_media_message -> delete_whatsapp_message_media
getattr.cloudwatch.get_o_tel_enrichment               -> get_otel_enrichment
getattr.cloudwatch.start_o_tel_enrichment             -> start_otel_enrichment
getattr.cloudwatch.stop_o_tel_enrichment              -> stop_otel_enrichment
getattr.signin.create_o_auth2_token                   -> create_oauth2_token   # <-- this one
```

These exist because `xform_name` once produced the underscored spelling; the aliases keep old code working. botocore resolves them by emitting `getattr.<service>.<name>` from `BaseClient.__getattr__`.

`AioBaseClient.__getattr__` cannot: `emit_until_response` is `async def`, so a synchronous `__getattr__` has no way to await it. Per its own comment, it raises rather than returning a deferred attrgetter (which would make `hasattr` always true and break ddtrace). **That decision is correct** — the defect is only that `credentials.py` was copied from botocore *including* a call that depends on the resolution aiobotocore deliberately doesn't provide.

Confirming which spelling is canonical:

```python
>>> from botocore import xform_name
>>> xform_name('CreateOAuth2Token')
'create_oauth2_token'
```

So `create_oauth2_token` is the **generated** method and `create_o_auth2_token` is the **legacy alias** — the underscored one is the compatibility shim, not the real name. (botocore's own `credentials.py:2633` also calls the alias; identical line, different outcome, because its `__getattr__` resolves it.)

### Suggested fix

One line, in `AioLoginCredentialFetcher.refresh_credentials`
(`aiobotocore/credentials.py`, master — line 1409 in the 3.9.0 sdist):

```diff
-                response = await client.create_o_auth2_token(
+                response = await client.create_oauth2_token(
```

The neighbouring event registration `'before-call.signin.CreateOAuth2Token'` is the PascalCase **operation** name and must stay exactly as it is — only the snake_case method call changes. Verified on a live async `signin` client: `create_oauth2_token` resolves and is callable.

A more general alternative, if maintainers prefer to close the whole class: have `AioBaseClient.__getattr__` consult botocore's registered `ClientMethodAlias` map directly (a pure dict lookup, no emit, so no await is needed) and re-dispatch to `_actual`. That would fix the remaining five aliases too, should an async caller ever reach them. The one-line change is sufficient for this bug and carries less risk.

### Impact

Any async S3/AWS workload authenticating via an AWS `login_session` (notably **`dvc push`** to an S3 remote, which uses aiobotocore) fails as soon as the cached token needs refreshing. It reads as flakiness and isn't: the refresh returns early while the token has life left, so the broken line is reached only at expiry — the clock triggers it, nothing the user changed. The classic `sso_session` flow is unaffected (it refreshes via `create_token`, a real generated name needing no alias).

### Versions

* aiobotocore 3.9.0 (latest on PyPI at time of writing)
* botocore 1.43.56
* Python 3.14, Linux
* Present in aiobotocore `master` and botocore `develop` as of 2026-08-06 — **no release avoids it**, so upgrading or reinstalling is not a workaround.

alloutflo:
### Description of Change

Login credential refreshes currently call botocore’s compatibility alias `create_o_auth2_token`. Synchronous botocore resolves that alias through `BaseClient.__getattr__`, but aiobotocore cannot provide an asynchronous equivalent; this change calls the generated `create_oauth2_token` method directly and updates the mocks so the tests exercise the runtime method. Closes #1697.

### Assumptions

The generated SignIn client exposes `create_oauth2_token` across the supported botocore range. This is the canonical generated method in the pinned botocore version and in aiobotocore’s async client API.

* [x] If this is resolving an issue (needed so future developers can determine if change is still necessary and under what conditions) (see #1697):
  * [x] Alternative methods considered: adding general alias resolution was broader and riskier than calling the generated method directly
* [ ] If this is providing a new feature (not applicable)
  * [ ] Why needed

### Validation

* `uv run pytest -q tests/botocore_tests/unit/test_credentials.py` (85 passed, 68 skipped, 81 deselected)
* `uv run pytest -q tests/botocore_tests/unit/test_credentials.py -k login` (13 passed, 17 skipped)
* `uv run pytest -q tests/test_patches.py` (242 passed)
* `uv run poe lint`

Not applicable; this PR changes neither dependency version.


.gitignore
jakob-keller, line 1: Could you please undo this change?

jakob-keller (changes requested):
Thanks for the contribution!

jakob-keller:
@claude please review

.gitignore
alloutflo: Done in 92cc453 — the .gitignore change is removed from the PR.

jakob-keller (approved):

